Issue #27, opened 2021-06-10
Create skeleton system architecture and diagram


carjug:
As a team, we need a rough outline for how our application will work so that we have a shared understanding and can divide up our work and plan the build of Unicorn accordingly.

At a level of detail suitable for team's own common understanding.

Pref to do this work collaboratively rather than async.

**Tasks**

- [x] Determine what diagramming software we want to use (OCIO may have a preference here)
- [x] Determine skeleton system architecture and document with a lightweight decision record and architecture diagram

**Resources**
System diagrams for other PB Portfolio projects that have gotten ATOs:
* https://github.com/HHS/TANF-app/blob/main/docs/Security-Compliance/boundary-diagram.md
* https://github.com/HHS/Head-Start-TTADP/blob/main/docs/boundary_diagram.md

amymok:
Potential diagram tools:

- TANF: [Draw.io](https://app.diagrams.net/)
  - Pros: feature-rich diagram tool; can use different icons (i.e. AWS icons) to create easy-to-read diagram; it can 
  - Cons: It is not diagram-as-code; exporting to XML does not show the structure of the diagram
- HeadStart TTADP: [C4 variant of PlantUML](https://github.com/RicardoNiepel/C4-PlantUML)
  - Pros: Diagram as code, can version control the code
  - Cons: The syntax and style is somewhat limited, and although possible, when altering from the existing syntax and style, it is somewhat hard to do.

carjug:
The completion of this task won't be "Done" until next week when all the engineers can get together and create the diagram collaboratively, but we will get started on initial investigation and leaving thoughts/comments this week.

carjug:
My initial attempts to capture what we currently know about our system's architecture. Let's definitely iterate on this together and do some live pairing! I think that will yield more fruitful results, because I definitely had big questions in my mind about how we should document our database/migration, etc.

![Start of a Diagram](https://user-images.githubusercontent.com/10871517/124330805-65177b80-db4b-11eb-9e27-7be0ccd2cca7.png)

ninamak:
Next steps:

- [ ] Eng co-work to collaborate on this diagram

Issue #55, opened 2021-06-22
Decide on tech services needed for continuous integration


alexsoble:
As a team, we need to agree upon which technologies we will be using for CI/CD.

Table for reference:

  | Carly | Amy | Alex
-- | -- | -- | --
Technology for CI/CD | CircleCI | CircleCI, What about GitHub Action? | CircleCI

**Tasks**

- [ ] Determine technology for CI/CD and document with a lightweight decision record
- [ ] Create "procurement" and/or any implementation follow on tasks

carjug:
This is blocked by needing to hear from the HHS OCIO what version of GitHub they use.

ninamak:
HHS OCIO does use a version of GH that allows for Actions, but now we're blocked by not having access to HHS GH (they don't have enough licenses).

Options:
* Set up CircleCI in our 18F repo for now
* Alternatively, move to a public HHS repo (until/unless the need for a private repo re-emerges)

Next steps:
- [ ] Look into compliance status of GH Actions vs CircleCI
- [ ] Discuss above options at Eng Co-Work Weds 7/14
- [ ] Rename/name repo RPM (aka Research Portfolio Management)

alexsoble:
☝️ All 4 of us have already started adding ideas here async, so I'm going to move this from TODO to In Progress!

alexsoble:
We know that at least 2 other projects in ACF are using the HHS CircleCI org. Adopting CircleCI for this project would let us move this task to "Done" the quickest, because we could use the tracks that those other projects have laid down. 

The main benefit I see to using GitHub Actions is that it would reduce the number of third-party tech services we need to manage, since we already plan to use GitHub for source control. It would also shrink our "surface", meaning the number of third-party services whose security & availability we rely on to keep our product secure & available. 

However, we may need to have a fair amount of discussion and back-and-forth with HHS GitHub administrators about GitHub Actions. For the enterprise version of GitHub, access to some Actions e.g. those on GitHub Marketplace is not enabled by default: 

https://docs.github.com/en/enterprise-server@2.22/admin/github-actions/managing-access-to-actions-from-githubcom/about-using-actions-in-your-enterprise

carjug:
Hey @alexsoble when we originally reached out to Domenic at HHS, he confirmed that the version of GH they use at HHS does have access to Actions. Just as an FYI 💟

alexsoble:
[...]

PR #65, opened 2021-07-07
Adds final state draw.io skeleton architecture diagram

carjug:
### What does this change?
Adds png for our desired final state architecture diagram to our documentation

### What does this change look like?
Diagram

### GitHub cards
Completes part of #27


carjug:
Question for the group: do we also need to add an ADR for the architecture we've come up with? That feels like it'd be covered by the tech stack ADR mostly, but....I wasn't quite sure.

amymok (commented):
Looks good!  I would suggest to rename the file to something like `final_state_architecture_diagram.png`, `architecture_diagram.png`, etc. because as-is reads like `state_diagram`, and it may be confusing that this is an actual "state diagram" and not an architecture diagram.

amymok:
> Question for the group: do we also need to add an ADR for the architecture we've come up with? That feels like it'd be covered by the tech stack ADR mostly, but....I wasn't quite sure.
> CC: @amymok @alexsoble

I think we need to explain and record some of the choices as ADRs if we haven't yet.  For example, CircleCI, S3.

carjug:
> Looks good! I would suggest to rename the file to something like `final_state_architecture_diagram.png`, `architecture_diagram.png`, etc. because as-is reads like `state_diagram`, and it may be confusing that this is an actual "state diagram" and not an architecture diagram.

Got it! Fixed

alexsoble:
Looks good to me as a starting point! A good visual we can use to talk with our PM and PO! 

I approve this PR, but I think we'll also want PM and maybe PO review for this one.

alexsoble (approved):

alexsoble:
I can also easily import the PNG into draw.io and edit it, which is great!

amymok (approved):
Looks great!

carjug:
@amymok to respond to your second comment, I think the CircleCI choice is covered by [#55](https://github.com/18F/OPRE-Unicorn/issues/55). The choice for S3 is not covered anywhere but I'm not sure if we should add it as a part of this issue or create a new one? either is fine for me.

amymok:
Oh I do not mean adding to this issue, I am just saying we should probably have an ADR on those.

ninamak:
Would it work to walk me through this diagram during our Demo block on Mon (7/12)?

Who is going to create cards for the ADRs that need to be written up?

carjug:
@ninamak Sure, I can walk you through it. 
There is one card that needs to be made for one ADR, I can do that.

Issue #25, opened 2021-06-10
Decide on a tech stack


carjug:
As engineers, before we can start building, we need to know what tech stack we'll be working in and why.

|    | Alex initial thoughts | Carly | Amy initial thoughts |
|---|------|------|------|
| Frontend | Use Django Template+HTML+USWDS <br/> Use JS as needed, start out with SSR <br/> (server-side rendering) | Depends on what the product needs to do. | Server-rendered HTML, like Django Template, USWDS, Javascript as needed <br/>or<br/>Complete decoupled frontend/backend: HTML/USWDS/React.js/Redux |
| Backend | Django | Django or Flask | Django |
| Database | Postgres | Postgres | Open Source Relational Database: Postgres or MySQL |

**Tasks**

- [x] Determine tech stack for backend and document with a lightweight decision record
- [x] Determine tech stack for frontend and document with a lightweight decision record
- [x] Determine whether or not we need our frontend and our backend to be completely decoupled and document with a lightweight decision record 
- [x] Determine database and document with a lightweight decision record

Related issue #25

carjug:
@amymok @alexsoble I'm pulling this into "In Progress" since Amy mentioned she's started looking into some of the options here already 😄

amymok:
I am adding a few links here, mostly for backend, comparing Django, Flask, and FastAPI:

https://www.section.io/engineering-education/choosing-between-django-flask-and-fastapi/
https://ahmed-nafies.medium.com/why-did-we-choose-fast-api-over-flask-and-django-for-our-restful-micro-services-77589534c036
https://www.reddit.com/r/django/comments/jfuk6d/django_rest_framework_vs_fastapistarlette/
https://codeahoy.com/compare/django-vs-flask.html
https://codeahoy.com/compare/flask-vs-fastapi.html
https://codeahoy.com/compare/django-vs-fastapi.html

carjug:
@amymok @alexsoble Anyone feel particularly strongly about being the "assignee" here? It could be any one of us, IMO.

amymok:
Have your pick!

carjug:
@amymok I'll assign you since you're not ~technically ~ on the board yet :)

amymok:
Add a few more notes, this one is for Django/frontend framework like React.js
https://thenewstack.io/djangos-place-in-a-web-development-world-ruled-by-react/
https://www.quora.com/What-are-the-benefits-of-using-react-js-with-Django

[...]

PR #79, opened 2021-07-14
Front-end tech stack: Start off with server-side rendering

alexsoble:
## What does this change? 

Adds ADR for front-end tech stack.

## What does this change look like?

[Preview link.](https://github.com/18F/OPRE-Unicorn/blob/ars/25/start-off-with-server-side-rendering.md/docs/adr/006-start-off-with-server-side-rendering.md) 👁️ 

## GitHub cards

This should complete #25!

### DIFF
--- a/docs/adr/006-start-off-with-server-side-rendering.md
+++ b/docs/adr/006-start-off-with-server-side-rendering.md
@@ -32,3 +32,5 @@ We will start off with server-side HTML rendering, and add more front-end toolin
 ## Consequences / premortem
 
 See tradeoffs above.
+
+If user research and initial prototyping suggest that we will need a large amount of JavaScript to implement the project, we may need to request more funding to bring on additional developers to write JavaScript. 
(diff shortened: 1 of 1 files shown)


carjug (approved):
Thanks for this, Alex! LGTM :shipit:

ninamak (approved):
LGTM!

alexsoble:
@amymok If you approve, could you also merge on approving? Thanks!

docs/adr/006-start-off-with-server-side-rendering.md
```diff
@@ -0,0 +1,36 @@
+# 6. Front-end tech stack: Start off with server-side rendering
+
+Date: 2021-07-14
+
+## Status
+
+Proposed
```
amymok, line 7: This should be Accepted before we merge.

amymok (changes requested):
LGTM, we should "Accept" this ADR.

docs/adr/006-start-off-with-server-side-rendering.md
alexsoble: Good catch!

Issue #56, opened 2021-06-23
Establish eng team roles


alexsoble:
As an engineering team, we would like to have a shared understanding for who fulfills/shares what roles.

**Notes**

Broke out this card from the larger eng team processes card, #23. 

**Tasks**

+ [x] Evaluate & discuss implementing a tech lead role
+ [x] Evaluate & discuss other eng team roles needed for the project
+ [x] Document our decisions

ninamak:
@alexsoble What are you seeing as the blockers here?  Outages?  Waiting on the work the Eng chapter is doing on this topic?  Something else?

alexsoble:
@ninamak Outages — just channeling what you wrote here! https://github.com/18F/OPRE-MAPS/issues/31#issuecomment-865370725

> This is unlikely to happen this week before people start heading out (and there will be fewer people to coordinate while people are out), so we will likely pick this back up after July 4. Eng may have more guidance for us by that point, too.

We could pick the discussion up after July 4. 

And as you note — eng may have more guidance for us by that point, too!

ninamak:
Got it, thanks!  Was just wondering if anything else had been raised.  So I'm going to remove the tech lead convo from the scope of #31 since it's covered here.

ninamak:
Eng has said they have no more guidance for us.  Mtg to discuss what we want in the Tech Lead role scheduled for Tues 7/13

carjug:
@amy to add things here :) the rest of the engineering team will review/help where needed! yay for formalizing things.

amymok:
During our eng team lead role discussion, we have also identified the engineering work we need to do for this project.  We created a table with the list of engineering work, and the primary person responsible for the work.  A backup person was also identified for each in the case the primary person is unavailable.

We also wanted to point out the work identified with an asterisk (*) are big tasks, and will require significant effort.  We want to be mindful when we assign work for each engineer.

 Engineering Work | Primary | Backup
-- | -- | --
More tech-lead-specific things… <br><br>Keep tabs on the different eng streams and whether they’re flowing smoothly individually and together<br><br>Ensure engs have a good sense of their roles; eng team roles/shape matched to the work that needs doing | AM | ARS
Meet with PM to provide initial perspective on draft milestones, cycles, stories | AM | CJ
Meet with OCIO to represent project, fill out ATO docs* | ARS | AM, CJ
Guide eng team to technical decisions | ARS | AM, CJ
Implementation — back end | CJ | AM
[...]

PR #80, opened 2021-07-14
Establish Engineering Team roles

carjug:
### What changed
Added "Engineering Team roles" section to `how_we_work.md`

@amymok @alexsoble I wasn't sure exactly what to put under the "Tech Lead role" subheading. Very open to suggestions there! Especially, would love your feedback @amymok since you're the Tech Lead after all :)

### Addresses
Linked issue #56 

### Preview link:
https://github.com/18F/OPRE-Unicorn/blob/0061fa980d01d800f278cb360c14465c945d8d91/docs/how_we_work.md

### DIFF
--- a/docs/how_we_work.md
+++ b/docs/how_we_work.md
@@ -32,11 +32,16 @@
 
 ## Engineering team roles
 
+### Tech Lead role
+
+Amy Mok will act as our Tech Lead for this project, with backup support from Alex Soble and Carly Jugler.
+
 ### Table of responsibilities
 
 Task/Responsibility | Primary | Backup
 -- | -- | --
-More tech-lead-specific things… Keep tabs on the different eng streams and whether they’re flowing smoothly individually and togetherEnsure engs have a good sense of their roles; eng team roles/shape matched to the work that needs doing | AM | ARS
+Keep tabs on the different eng streams and whether they’re flowing smoothly individually and together<br>
+Ensure engs have a good sense of their roles; eng team roles/shape matched to the work that needs doing | AM | ARS
 Meet with PM to provide initial perspective on draft milestones, cycles, stories | AM | CJ
 Meet with OCIO to represent project, fill out ATO docs** | ARS | AM, CJ
 Guide eng team to technical decisions | ARS | AM, CJ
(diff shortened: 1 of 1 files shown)


docs/how_we_work.md
```diff
@@ -30,6 +30,29 @@
...
+
+### Tech Lead role
+
+Amy Mok (AM) will act as our Tech Lead for this project, with backup support from Alex Soble (ARS) and Carly Jugler (CJ).
+
+### Table of responsibilities
+
+Task/Responsibility | Primary | Backup
+-- | -- | --
+Keep tabs on the different eng streams and whether they’re flowing smoothly individually and together<br>
```
alexsoble, line 43:
```suggestion
Keep tabs on the different eng streams and whether they’re flowing smoothly individually and together | AM | ARS
```

In the preview it looks like the first row of the table is blank; this suggestion would fill in that role.

alexsoble:
@carjug In terms of what goes under the Tech Lead subheadding: I think traditionally a tech lead might own rows 1 through 5 of the table we sketched out; in our case we're distributing those responsibilities a little more broadly across the team. Does that help? 

Or put another way, I don't think we necessarily need to add things under the Tech Lead subhead that won't be in the table of tasks/responsibilities.

amymok (commented):
This looks great!  Agree with Alex, we don't need to add anything else since we have clearly laid out the responsibilities.

alexsoble (approved):

ninamak (approved):
I still have concerns about dividing up the Tech Lead responsibilities, but I'm happy to give it a try.  Can we build in a retro in the near future to check in on how things are going with the role, how people are feeling about the load, etc?

carjug:
@ninamak I think talking about the tech lead role in our regularly scheduled retro is totally reasonable!
It's worth mentioning that Alex, Amy and I feel that if one person were to take on all of the listed responsibilities, we would either need to be prepared to significantly cut scope for the product or ask for more money for another engineer.

amymok:
@ninamak the way the engineers are in this project are more defined by our expertises which is a bit different from some of the other projects which started with one tech lead who owns the entire piece with knowledge in all the area from the start, and then building out the engineering team based on additional needs to do the work.

The tech lead in this project will still need to check in with each engineer to understand their domains and needs for this project, and then the lead can serve to understand how each stream is going based on that to align it for the overall project/goals.

carjug:
@amymok if you approve can you merge, plz 🙏 💟

amymok (approved):
Sorry about the delay!  LGTM!

PR #82, opened 2021-07-15
Change all ADR status from Proposed to Accepted

amymok:
## Notes
There are a few ADRs that have already been merged that are still in the "Proposed" state.  Since they have all been reviewed and approved by the team, they should be in the "Accepted" state.

## Preview link
https://github.com/18F/OPRE-Unicorn/tree/aym/68/accept-adrs/docs/adr


alexsoble (approved):
Good catch!

carjug (approved):
Thank you! LGTM

PR #104, opened 2021-07-26
Create .gitattributes for better compatibility with Windows-based devs

### LINKED ISSUE #52: Add a .gitattributes file
In 18F projects, we sometimes run into issues when government partners download a repo and track changes using git on Windows machines. Different operating systems have different ways of handling line endings, which can create compatibility issues when team members use git to track changes using different operating systems.

Adding a `.gitattributes` file from the beginning can help prevent these compatibility issues before they crop up, and avoid the need to renormalize later on.

General resources on `.gitattributes`: 

* https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
* https://git-scm.com/docs/gitattributes

Example `.gitattributes` files on 18F projects:

* https://github.com/usdoj-crt/crt-portal/blob/develop/.gitattributes
* https://github.com/HHS/TANF-app/blob/main/.gitattributes

alexsoble:
## Sources

+ https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file
+ https://tekin.co.uk/2020/10/better-git-diff-output-for-ruby-python-elixir-and-more
+ https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
+ https://git-scm.com/docs/gitattributes

## Issue 

+ Closes #52.

## Reviewer notes

Hmmm, how best to review this one? A very rigorous review would involve messing with line endings on your local machine and verifying that the `.gitattributes` enforces the right line endings, I guess? But that would probably be overkill. 

The overall idea here is that having `.gitattributes` up front sets us up for success if/when a developer with a Windows machine [a partner tech lead??] joins us down the line. 

So in my mind, "this looks like a sensible default `.gitattributes`" should be sufficient for a 👍  review.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,16 @@
+# Explicitly declare text files we want to always be normalized and converted on checkout.
+*.py text diff=python
+*.md text
+*.js text
+*.html text
+
+# Denote all files that are truly binary and should not be modified.
+*.ttf binary
+*.woff binary
+*.woff2 binary
+*.svg binary
+*.png binary
+*.jpg binary
+*.mp3 binary
+*.jpg binary
+*.jpeg binary


.gitattributes
```diff
@@ -0,0 +1,16 @@
+# Explicitly declare text files we want to always be normalized and converted on checkout.
```
alexsoble, line 1: I followed this recommendation from a StackOverflow user not to lead with `* text=auto`:

> We had a repo with * text=auto, and Git guessed wrong for an image file that it was a text file, causing it to corrupt it as it replaced CR + LF bytes with LF bytes in the object database. That was not a fun one to debug.

https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file

.gitattributes
```diff
@@ -0,0 +1,16 @@
...
+*.py text diff=python
```
alexsoble, line 2: Interesting tidbit on what specifying the `diff` does: 

https://tekin.co.uk/2020/10/better-git-diff-output-for-ruby-python-elixir-and-more

amymok: This is interesting, I actually never really look at the enclosing context, only line number :). But this is good to know!

amymok (approved):
LGTM!

.gitattributes
alexsoble: Yeah, same! Interesting little tidbit though!

alexsoble: Kind of silly how you have to tell git that `*.rb diff=ruby`, `*.py diff=python`, but 🤷 ?

carjug (approved):
Whoa! Cool stuff, thanks, Alex!

alexsoble:
@ninamak Would you like to review this one?

ninamak:
> @ninamak Would you like to review this one?

LOL, I don't think I have anything to contribute here!  I can merge it since the other devs have given 👍

Issue #85, opened 2021-07-16
[8] Create a "Hello World" application in 18F cloud.gov sandbox


ninamak:
### Goals
* Verify that we can set up a minimal application using our chosen tech stack that can be pushed to cloud.gov

### Tasks
- [x] Choose a Django version to use and create a "Hello World" app
- [x] Manually push to 18F cloud.gov sandbox
- [x] Create a Postgres DB and link it to the app
- [x] Give access to the app to others on the team

### Conditions of Acceptance
* Any member of this team can access the app and see "Hello World"
* People who are not members of this team cannot access the app

### Additional Context
* If we get a decision about how OPRE wants to set up their cloud.gov org & space before this issue is ready to be taken on, we will set up this application in that space rather than in the 18F sandbox.

carjug:
@amymok @alexsoble I was planning on just going with the latest version of Django (3.2.5). Is there any reason I shouldn't do this by y'all's estimation?

amymok:
@carjug Looking at [the LTS releases](https://www.djangoproject.com/download/#supported-versions), 3.2.x is the best option.  So yes, 3.2.5 sounds good!

carjug:
Not totally sure the 4th task can happen without me just sharing my TTS sandbox credentials directly with other devs. @amymok do you have any ideas here?

carjug:
@amymok ready for review!

carjug:
@ninamak ready for your review, see https://opre-ops-test.app.cloud.gov/hello_world/

ninamak:
Can confirm I can access it! :tada: 
How would I confirm that someone who is not on the team cannot access it?

carjug:
Oh @ninamak the URL is public, the task above is meant to give access to the developers (I can also give you access if you have a cloud.gov account) to the cloud.gov dashboard for the application.

ninamak:
Ok, that was a misunderstanding then.  As we start building the app, let's keep an eye out for when we need to make the URL non-public and make sure we have a story for that.

PR #118, opened 2021-08-05
Hello World app

carjug:
## What was done
- Fresh Django project named `opre_ops`
- Fresh Django app created inside `opre_ops/opre_ops` called `hello_world`
- Route and view `hello_world` declared that returns an HTTPResponse of "Hello world, it's me Margaret."
- database setup

Visit https://opre-ops-test.app.cloud.gov/hello_world/ for some exciting stuff

## Sources
These are all the sources I can remember visiting in the making of this PR  that might be useful to a reviewer

https://docs.djangoproject.com/en/3.2/intro/tutorial01/
https://stackoverflow.com/questions/22841764/best-practice-for-django-project-working-directory-structure
https://docs.djangoproject.com/en/2.2/ref/settings/#allowed-hosts
https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#minimal-manifest
https://docs.cloudfoundry.org/concepts/roles.html#activeroles
https://github.com/18F/usda-fns-ingest/blob/master/usda_fns_ingestor/usda_fns_ingestor/settings/base.py
https://djangostars.com/blog/configuring-django-settings-best-practices/

## Still to do/work that needs to be done but is potentially outside the scope of this ticket
- Figure out how we want to manage dependencies on the backend 
- Figure out how we want to manage settings between different environments

## Issue
#85


opre_ops/opre_ops/settings.py
```diff
@@ -0,0 +1,126 @@
...
+]
+
+WSGI_APPLICATION = 'opre_ops.wsgi.application'
+
+
+# Database
+# https://docs.djangoproject.com/en/3.2/ref/settings/#databases
+
+DATABASES = {
+    'default': dj_database_url.config()
```
carjug, line 78: Still working on db config stuff for local, will come back to this tomorrow AM.

opre_ops/opre_ops/urls.py
```diff
@@ -0,0 +1,23 @@
...
+    2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
+"""
+from django.contrib import admin
+from django.urls import path
+from opre_ops.apps.hello_world import views
+
+urlpatterns = [
+    path('admin/', admin.site.urls),
+    path('hello_world/', views.hello_world, name='hello_world')
+]
```
carjug, line 23: There's a bunch of different ways we can do routes in projects with apps with Django, this was just an easy first feature way, not sure how we want to do this in the future though

requirements.txt
```diff
@@ -0,0 +1,4 @@
+asgiref==3.4.1
+Django==3.2.5
+pytz==2021.1
+sqlparse==0.4.1
```
carjug, line 4: I mentioned this in the PR comment, but I think we probably want to manage our Python dependencies differently than a requirements.txt file.

start-web.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# todo: django docs say don't use this `runserver` command in prod
+python ./opre_ops/manage.py runserver 0.0.0.0:8080
```
carjug, line 4: Amy had a good resource to use guinicorn instead of runserver here

amymok (approved):
LGTM!

opre_ops/opre_ops/apps/hello_world/views.py
```diff
@@ -0,0 +1,6 @@
+from django.shortcuts import render
+from django.http import HttpResponse
+
+
+def hello_world(request):
+    return HttpResponse("Hello world, it's me Margaret!")
```
alexsoble, line 6: 😄 😄

manifest.yml
```diff
@@ -0,0 +1,9 @@
+---
+applications:
+    - name: opre-ops-test
+      buildpacks:
+        - https://github.com/cloudfoundry/python-buildpack.git#v1.7.44
```
alexsoble, line 5: If we pin the precise version here, will we still get the benefit of cloud.gov sending us an email telling us to re-stage the application when a newer version of the buildpack is released? 

Looking at: https://cloud.gov/docs/getting-started/concepts/#buildpacks.

carjug: Ah yeah, we can just bring in the `python_buildpack` if you think that's best, @alexsoble. I didn't know this was an option!

alexsoble (approved):
LGTM! 🎉 `:hello-world:`

PR #143, opened 2021-08-24
Setup Django application to run with Buildpack inside Docker

### LINKED ISSUE #123: As a dev, I want to run the Django server locally
### Goals

As a dev, I want to be able to run the Django server locally.

### Definition of Done 

All three engineers can run the Django server on their local machines using the same documented commands.

### Tasks

- [x] Determine the right command to run the Django server locally and document
- [x] Debug any issues that crop up with the application locally
- [x] Potentially: add `settings.py` override for running the application server locally

carjug:
## What's changed
Allows devs to run django app in Docker container and builds app locally using buildpacks `pack` CLI. Uses gunicorn server now instead of python manage.py runserver which django explicitly recommends against.

Still to do:
- Run server on cloud.gov using docker?? This may require updating `start-web.sh` script or something else entirely.

## Issue
Closes #123

## Preview link
README: https://github.com/HHS/OPRE-OPS/blob/9ba7c39b01cc55fa39e18639bb528d27b8a103eb/README.md

## Resources
https://buildpacks.io/docs/app-journey/
https://buildpacks.io/docs/tools/pack/
https://django.readthedocs.io/en/1.5.x/howto/deployment/wsgi/gunicorn.html
https://github.com/GoogleCloudPlatform/buildpacks/blob/main/README.md#configuration
https://github.com/GoogleCloudPlatform/buildpacks/blob/main/README.md#default-entrypoint-behavior


alexsoble (changes requested):
👋 👋  I love seeing the README starting to get built out here!

When I follow the README instructions, the app builds locally and all seemed fine at first:

```
[2021-08-24 19:34:33 +0000] [1] [INFO] Starting gunicorn 20.1.0
[2021-08-24 19:34:33 +0000] [1] [INFO] Listening at: http://0.0.0.0:8080 (1)
```

However, when I tried to visit `/hello_world/` or `/admin/` locally, I got a 500 Internal Server Error error in response. 

Stack trace: 

```
  File "/layers/google.python.pip/pip/lib/python3.9/site-packages/django/core/handlers/base.py", line 129, in get_response
    set_urlconf(settings.ROOT_URLCONF)
  File "/layers/google.python.pip/pip/lib/python3.9/site-packages/django/conf/__init__.py", line 83, in __getattr__
    val = getattr(self._wrapped, name)
AttributeError: 'Settings' object has no attribute 'ROOT_URLCONF'
```

Looks like a settings issue! @carjug and @amymok, do you see something different?

alexsoble:
Also @carjug on the "run docker in cloud.gov" topic, some relevant docs! https://cloud.gov/docs/deployment/docker/

My default would be to use the supported buildpack in production because the "pro"s there are big ones:

> It “just works”.
> Automatic and constant security updates.
> All you need to do is write code.

And on the mirror image side, the cons of running custom docker images in production on cloud.gov: 

> Added responsibility for all security updates and bug fixes.
> More compliance responsibility means more work.

carjug:
@alexsoble oh whoops! I moved the settings file into its own directory last minute, i wonder if that's the issue. Let me see what I can do 🕵️‍♀️ !

amymok:
@carjug I think there are couple things I am seeing that may cause this error:
- Move of the `settings.py` file: You will now need to specific the `--settings` to the right path because it is no longer the default choice.  Some resources:
  - https://docs.djangoproject.com/en/3.2/topics/settings/#designating-the-settings
  - https://stackoverflow.com/questions/49235486/how-to-properly-runserver-on-different-settings-for-django/49236676
- May want to revisit the file structures, I am unsure if this is a cause or not:
  - Probably need to create another `opre_ops` level to put in `settings.py` folder, wsgi.py, and some other files.  There may be some upper level `opre_ops` folder we can remove so it doesn't need to go as deep.  Some examples:
    - The [first answer](https://stackoverflow.com/questions/49235486/how-to-properly-runserver-on-different-settings-for-django/49236676) has an example.
    - [Settings](https://github.com/18F/tock) are in the 3rd level of `tock`
    - [Settings](https://github.com/18F/usda-fns-ingest) at 2nd level at `usda-fns-ingest`

alexsoble:
Building on what @amymok mentioned about the file/folder nesting: `settings.py` and `.env.py` need to be on the same level in the hierarchy because `settings.py` calls `.env.py`:

```
from .env import env
```

alexsoble:
@carjug I did some debugging on this and made progress, happy to pair with you on it tomorrow if you're blocked.

carjug:
Ok @alexsoble @amymok I just moved the settings file back to where it was originally, so not inside its own `settings` dir. Given that there is a separate ticket to address separating settings out per environment I vote to just leave as is and address the settings issue in that ticket. Is that good by y'all? Please pull down again and test locally if you'd like!!

README.md
```diff
@@ -1,2 +1,17 @@
...
+Instructions to start app locally:
+
+1. Make sure you have Docker installed locally
+2. Make sure you have pack installed, see [this link](https://buildpacks.io/docs/tools/pack/) for instructions
+
+From the project root run:
+```
+pack build opre-ops --builder gcr.io/buildpacks/builder:v1
+docker run --rm -p 8080:8080 opre-ops
+```
```
amymok, line 17: Add instruction on how to run it:
Navigate to http://localhost:8080/helloworld in your browser

amymok (changes requested):
I am able to get to http://localhost:8080/hello_world, but I ran into an issue when navigating to http://localhost:8080/admin.  I got an `ImproperlyConfigured at /admin/: settings.DATABASES is improperly configured. Please supply the NAME value.`

I believe the admin interface is out of scope, so you may want to comment out the admin url for now.

Other than those, it looks good to me!

README.md
```diff
@@ -1,2 +1,17 @@
...
+# OPRE OPS
+
+## Getting Started
+### Running the Application locally
+
+Application built using [pack Buildpack CLI](https://github.com/buildpacks/pack) and runs within Docker.
+
+Instructions to start app locally:
+
+1. Make sure you have Docker installed locally
```
amymok, line 10: installed and *started* locally.  I ran into an issue because I didn't start Docker.

amymok (changes requested):

alexsoble:
@carjug 🎉, works for me locally with that settings file change!

alexsoble:
[suggestion] Remove the `start-web.sh` script as part of this PR, since it's deprecated now that we are using `Procfile`?

carjug:
@amymok I addressed your requested changes! lmk what you think :)

carjug:
> [question] Do we want to have some kind of manual-deploy-to-cloud-dot-gov-and-verify-that-everything-works step before final approval and merge?

@alexsoble given that this ticket is for running locally only, I would say "idk" just because nothing will have fundamentally changed in terms of how the app is currently running on dev/cloud.gov (still using the `start-web.sh` script) as that will be addressed in a later ticket (IIRC). We can talk more about this in co-working if you want!

alexsoble:
@carjug Ah, I didn't realize that the cloud.gov app is still using `start-web.sh`. I wonder if introducing a Procfile will change things on cloud.gov at all? Like will CloudFoundry see the Procfile and use it? https://docs.cloudfoundry.org/buildpacks/prod-server.html

Anyways, I digress ... keeping this scope to local development for now makes sense to me! Build out a good local dev experience first, other things can come later!

amymok (approved):
LGTM!

PR #156, opened 2021-08-27
Separate Django settings for local and cloud environments; connect app to DB in cloud.gov

### LINKED ISSUE #158: As a dev, I want separate Django settings for local/cloud environments
### User story / goals

As a dev, I want the application code to connect to our Postgres database in the cloud.gov environment.

alexsoble:
## Issue

Closes #158.

## Documentation added

👁️  [How to: set Django SECRET_KEY in the cloud](https://github.com/HHS/OPRE-OPS/blob/ars/django-settings/docs/recipes/set_django_secret_key.md) 👁️ 
## How to test

+ [x] Verify application code runs and connects to Postgres locally.
+ [x] Deploy app code to the cloud.gov sandbox environment. Verify application code runs and connects to Postgres in a cloud.gov sandbox environment.

To verify that the app code connects to Postgres:

Visit the app at the `/admin/` URL, and type in an arbitrary username and password. If you not allowed in, but do not get a 500 error, that means the app code is successfully talking to the Postgres database when it tries to authenticate you!


opre_ops/opre_ops/settings/cloudgov.py
```diff
@@ -0,0 +1,27 @@
...
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': os.environ.get('DATABASE_NAME'),
+        'USER': os.environ.get('DATABASE_USER'),
+        'PASSWORD': os.environ.get('DATABASE_PASSWORD'),
+        'HOST': os.environ.get('DATABASE_HOST'),
+        'PORT': '5432'
+    }
```
amymok, line 19: I remember we can't expect the user-defined environment variables to stay put after redeployment to cloud.gov (I ran into that issue about 1-2 years ago).  Instead we will need to use [user-provided service credentials](https://docs.cloudfoundry.org/devguide/services/user-provided.html).

But I don't think we need to do that, cloud.gov provides the variable called [`DATABASE-URL`](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#DATABASE-URL) that we can use for this purpose.

I think that's what [`dj-database-url`](https://github.com/jacobian/dj-database-url/blob/master/test_dj_database_url.py#L149-L158) is for.  It takes `DATABASE_URL` and pulls out the necessary information, which is in cloud.gov environment.  This will work because it takes that variable and populates everything correctly.

After taking a look at [my other example](https://github.com/18F/usda-fns-ingest/blob/master/usda_fns_ingestor/usda_fns_ingestor/settings/base.py#L78) in previous project, now that makes sense.  `default` within `config` is only used for local db, because there's no need for hostname/username/password, only the name of the database (that's why there's three `/`s.  But in cloud.gov settings, the `default` is not used, it will take cloud.gov's `DATABASE_URL`.

I would either consider giving `dj-database-url` another try, or we can pull out information from `DATABASE_URL` from cloud.gov and do essentially the same thing to fill out this information.

opre_ops/opre_ops/settings/cloudgov.py
```diff
@@ -0,0 +1,27 @@
...
+}
+
+# SECURITY: Keep the secret keys used in production secret!
+SECRET_KEY = os.environ.get('SECRET_KEY'),
```
amymok, line 23: Same comment about using [user-provided service credential](https://docs.cloudfoundry.org/devguide/services/user-provided.html)

amymok (changes requested):

opre_ops/opre_ops/settings/cloudgov.py
alexsoble: Very helpful, thank you @amymok!

Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM python:3.9.6
 ENV PYTHONUNBUFFERED=1
+ENV DJANGO_SETTINGS_MODULE=opre_ops.settings.local
```
alexsoble, line 3: Local settings need to be opted into explicitly, I want settings to default to cloud rather than local as a fail-safe.

manifest.yml
```diff
@@ -1,9 +1,11 @@
...
 applications:
-    - name: opre-ops-test
-      buildpacks:
-        - python_buildpack
-
+  - name: opre-ops-test
+    buildpack: python_buildpack
+    # TODO: Django docs say not to use the `runserver` command in prod;
+    # switch over to a production-ready webserver like gunicorn
+    command: cd opre_ops && python manage.py migrate && python manage.py runserver 0.0.0.0:8080
```
alexsoble, line 7: Updated this infrastructure as code configuration to run our Django migrations before running the application server.

opre_ops/opre_ops/settings/cloudgov.py
alexsoble: @amymok I updated this PR so that it now reads in the cloud.gov configuration from `VCAP_SERVICES`. Appreciate the pointers here, I did not realize that setting ENV variables through the cloud.gov UI/CLI wasn't the way to go! I didn't implement `dj-database-url` because my rule of thumb is to avoid third-party libraries when it's easy/straightforward to do so. Please take another look when you have time!

opre_ops/opre_ops/settings/cloudgov.py
alexsoble: Changed to use a user-provided service credential here!

manifest.yml
```diff
@@ -1,9 +1,11 @@
...
 services:
  - opre-ops-psql-db
- 
\ No newline at end of file
+
```
amymok, line 10: `OPRE_OPS_ENV_SERVICE` will now be part of the `services`

alexsoble: Oooooh! And that will do the "bind" step automatically? Looking at:

https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#services-block

alexsoble: Also should this be lowercase/kebab-case to match `opre-ops-psql-db`?

amymok: Yes.  It looks like that's what they use, I am unsure if there's an actual style guide, but looking at [examples](https://docs.cloudfoundry.org/devguide/services/user-provided.html), it uses lowercase with hyphens

alexsoble: Gotcha, thank you!  Will make those changes!

amymok (approved):
I tested out locally and on cloud.gov.  They both look good.

There seems to be a problem on loading asset that needs to resolve in another issue.

alexsoble:
@amymok Broke out that static assets issue here: #162.

docs/recipes/set_django_secret_key.md
```diff
@@ -0,0 +1,15 @@
...
+
+```
+# first, log in:
+cf login --sso
+
+# then, create a service named "opre-ops-env-service" and pass in this key/value pair:
+cf cups opre-ops-env-service -p '{"DJANGO_SECRET_KEY":"your_secure_secret_key_here"}'
+
+# the app manifest (manifest.yml) specifies an application bind to opre-ops-env-service; you may need to re-start the app to apply the bind
+```
```
carjug, line 15: Yay for docs!!

manifest.yml
carjug: Thanks for maintaining the note about not using `runserver` in prod!

carjug (approved):
LGTM, Alex! 🚢

alexsoble:
Woohoo! 🚢

Issue #133, opened 2021-08-19
[4] View agencies


amymok:
### User Story
As a user, I will want to know the agency that provided CAN money and whether they need to authorize requisitions that use that CAN.  

### Conditions of Acceptance
- [x] I can use Django Admin in the dev environment to see a list of agencies (based on dummy data).
- [ ] It would be ok if the PM can only view the interface on a dev's local.

### Tasks
- [x] Implement Agency model based on [Agency tab](https://docs.google.com/spreadsheets/d/190drEOaOFH7cwCBM7eNB6vmnjICDiiShQiut7B55HCY/edit#gid=0)
- [x] Populate with dummy data from AirTable [Agencies tab](https://airtable.com/tblGGRgffwAm6oV3h/viw42XXeFHzgoXwuB?blocks=hide)
- [x] Enable Django Admin interface to display the Agency table

### Additional Context

- [Database Models Spreadsheet](https://docs.google.com/spreadsheets/d/190drEOaOFH7cwCBM7eNB6vmnjICDiiShQiut7B55HCY/edit#)
- [ER Diagram](https://docs.google.com/drawings/d/106BT0UIW_-zmocmNWMoGbe4nVmHGU1fYfsQWnTccsiI/edit)

### Resources
* #91

alexsoble:
@carjug is picking this one up, we're excited to move into implementation world! 🎉

ninamak:
@carjug I tried to update this card to include a user story and CoAs, though it's a little awkward since this is more of a technical task in support of a user story captured elsewhere.  It's not intended to change anything about what work you were planning to do, but let me know if it does!

carjug:
@ninamak It's all clear to me!

carjug:
Ready for review @alexsoble @amymok

alexsoble:
Reviewed and approved with @ninamak synchronously! 👍 👍 🚀

PR #167, opened 2021-09-02
Add model to view agencies in dashboard

carjug:
## Issue
#133 

## What changed

- Added Agency model
- Seeded with test data
- Added Agency model view to Django Admin
- Changed display of Agency list in Django Admin

## Screenshots

Screen Shot 2021-09-01 at 10 19 26 PM
Screen Shot 2021-09-03 at 11 01 50 AM


alexsoble:
Thank you for demo'ing work-in-progress earlier today @carjug! 

A couple of quick notes on topics that came up during the demo:

### Pluralizing agencies:

Small thing, but looks like a straightforward fix:

https://stackoverflow.com/questions/2587707/django-fix-admin-plural
https://docs.djangoproject.com/en/dev/ref/models/options/#verbose-name-plural

### Showing names instead of Agency Object (n):

https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display

Both of these are small things but tweaking them could help make the presentation and intention of our Django Admin a little clearer! Leaving links to the docs in case they're helpful.

carjug:
Thanks @alexsoble!

alexsoble:
@carjug :highfive: :oneteamonedream:

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 3.2.6 on 2021-09-01 23:27
+
+import os
+
+from opre_ops.settings.local import BASE_DIR
```
alexsoble, line 5: _[caveat: I know this is a draft PR, you may have been planning to change this anyway!]_

The migrations will run in both local and cloud environments, so I don't think we'll want to pull from local settings. Looks like we could pull this from `settings.common`?

However, I don't see this imported variable used anywhere, so maybe we don't need it?

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
```diff
@@ -0,0 +1,27 @@
...
+from django.db import migrations
+from csv import DictReader
+from ops_site.models import Agency
+from pathlib import Path
+
+def seed_agencies(apps, schema_editor):
+    filename = os.path.join('ops_site', 'migrations', 'seed_data', 'agencies.csv')
```
alexsoble, line 12: _[caveat: I know this is a draft PR, you may have been planning to change this anyway!]_

Not such a big deal for this PR, but for future PRs in particular, I'd suggest using file names like `dummy-agency-data.csv`, `fake-invoice-amounts.csv`, etc just to make it super clear that this is dummy/fake data.

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
carjug: Oh yes! I was using this setting in a test of something and forgot to remove it. If we needed it you'd be correct though :)

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
carjug: Changed 😄

alexsoble: Woohoo, thanks! I think this establishes a nice pattern!

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
alexsoble: 👍 👍  Thanks!

opre_ops/ops_site/admin.py
```diff
@@ -0,0 +1,8 @@
+
+from django.contrib import admin
+from ops_site.models import Agency
+
+
+@admin.register(Agency)
+class AgencyAdmin(admin.ModelAdmin):
+    list_display = ("name",)
```
amymok, line 8: [Suggestion] Since you already seeded the `nickname` data, you may want to add them in the display for viewing purposes

opre_ops/opre_ops/settings/common.py
```diff
@@ -25,6 +25,7 @@
 # Application definition
 
 INSTALLED_APPS = [
+    'opre_ops.apps.OpsSiteConfig',
```
amymok, line 28: [Question] [the order of `INSTALLED_APP` matters](https://docs.djangoproject.com/en/3.2/ref/templates/api/#django.template.loaders.app_directories.Loader), should this addition be at the top?

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ops_site', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunPython(seed_agencies)
+    ]
```
alexsoble, line 25: [Thought, Non-Blocking] All of our migrations will run when we deploy the application code in a new environment. Seeding this fake data with a data migration will work for development, but in production we won't want to seed our database with fake data. 

Options we could consider:

### 1. Don't deal with this now

Keep seeding fake/development data using database migrations. When we get to production, we may need to change or delete these development data migration files and adopt one of the strategies below. 

### 2. Move data to a Django fixture

Relevant docs: https://docs.djangoproject.com/en/3.2/howto/initial-data/

We could move our seed data to JSON or YAML and load using `python manage.py loaddata <fixturename>`. We could add the `loaddata` task to our `Dockerfile` or `docker-compose` setup for local development, and use a settings flag to have it run in our cloud.gov Development/Demo environments, but not in Staging or Production. 

### 3. Add `if` logic to database migrations for seed data

Add custom logic or flags to our database migrations that seed data, so that they only run in certain environments.

--- 

@amymok @carjug Let me know what you think! How have you approached this in other projects? This comment is non-blocking because I would be totally fine with option (1). Looks like option (2) might be more Django-idiomatic over the long run?

amymok: I prefer option (2).  This way we can keep the migration clear from dev data.  But I agree that we can add a new ticket to move that option.  The only thing I am unsure about is, if we do that later, how does that work to "remove" a migration file?  Will that give us more headaches later when we have other new migrations?  Or is it easier to do that now and remove this migration before we add new one?

opre_ops/opre_ops/settings/common.py
carjug: This is the order the documentation had the installed apps in, but I don't think it matters.

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
alexsoble: @amymok Yep, I agree that removing migration files might get slightly messy because databases (local, deployed, etc) might be in different states. So that would suggest making the switch over to using Django fixtures in this PR rather than in a follow-on issue or PR. 

@carjug What do you think?

carjug: I can do that. There's a bit of manual labor involved because the file structure for Fixtures is not easily exportable from the AirTable but I can definitely do it.

opre_ops/ops_site/admin.py
carjug: I can do that!

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
alexsoble: @carjug Ooooh, maybe `python manage.py dumpdata` could give you the data in a fixture-friendly way without manual labor? 

https://docs.djangoproject.com/en/3.2/ref/django-admin/#django-admin-dumpdata

alexsoble: @carjug Since future issues will follow a similar pattern, maybe a lightweight CSV-to-JSON or CSV-to-YAML script will be helpful to us in the future? Like to get us a quick "export Airtable CSV" > "transform to JSON or YAML" pipeline.

carjug:
Ok @alexsoble so since I already had the agency data loaded into the db I could use the `dumpdata` command pretty easily here, therefore I didn't create the CSV -> JSON script, but I will do that with the next model ticket. LMK if this works locally on your machine. I added the `loaddata` step to the `docker-compose` and it worked on my local, but my local is weird due to mystery psql db that never shuts down so I often have to run the steps from the docker compose manually after exec'ing into the container.

alexsoble:
@carjug Sounds good! Here is what I'm seeing locally on my machine:

### What isn't working 

When I run a fresh `docker-compose down && docker-compose-build && docker-compose up` against the latest on this branch, I see these errors.

From `migration_1`: 

```
migration_1  |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
migration_1  | django.db.utils.OperationalError: could not connect to server: Connection refused
migration_1  | 	Is the server running on host "db" (172.24.0.2) and accepting
migration_1  | 	TCP/IP connections on port 5432?
migration_1  | 
```

From `data_1`:

```
data_1       | psycopg2.errors.UndefinedTable: relation "ops_site_agency" does not exist
data_1       | LINE 1: UPDATE "ops_site_agency" SET "name" = 'Children''s Bureau', ...
```

This looks like a race conditions issue! The migrations are trying to run before the database is ready to accept connections, so the migration fails, and then the data loading fails because the migrations haven't run.

### What gets it to work

If I run each step in turn, like this, then each step runs successfully!

```
docker-compose up db
docker-compose up migration
docker-compose up data
docker-compose up web
```

When I create a local superuser and log in, I see the Agencies table as described in the Conditions of Acceptance! 🎉  

However, we don't want devs to have to run these four steps every single time ... `docker-compose up` should just do the trick! We want our migration and data and web steps to wait patiently for the database to be fully ready before attempting to connect. 

### The fix?

Did some Googling for things like "wait for database docker-compose". Looks like this is a known issue and there are some tools Docker recommends that we can use to fix it: https://docs.docker.com/compose/startup-order/.

Will follow up in Slack, maybe we can pair on this!

opre_ops/ops_site/migrations/0002_auto_20210901_2327.py
alexsoble: @carjug You've addressed this by moving to a fixture pattern, excellent! Thank you!

alexsoble:
@carjug Worked on a small PR to this PR that we can look at together during pairing! #173

alexsoble:
Woohoo! We solved the race conditions issue in pairing! 🎉

alexsoble (approved):
LGTM! 🎉

alexsoble:
Local screenshot: 

Screen Shot 2021-09-08 at 11 47 09 AM

alexsoble:
Reviewed and approved with @ninamak synchronously! 👍 👍

PR #173, opened 2021-09-08
Modify docker-compose depends_on steps to prevent race condition

alexsoble:
## What does this change? 

Modify our use of docker-compose `depends_on` to set up an order for our services' dependences: 

`db` > `migration` > `data` > `web` 

https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on

When I run `docker-compose up` with this config locally, I see the services start in the right order.


carjug:
LGTM! 🎉 yay for pairing.

Issue #134, opened 2021-08-19
[4] View roles


amymok:
### User Story
* As a user, I want to see a list of Roles as they relate to CAN Management (realistic-looking dummy data).
* As a PM, I want to use this view to facilitate conversations about OPRE roles vs user types.

### Conditions of Acceptance
- [x] I can use Django Admin in the dev environment to see a list of roles (based on dummy data).
- [ ] It would be ok if the PM can only view the interface on a dev's local.

### Tasks
- [x] Implement Role model based on [Role tab](https://docs.google.com/spreadsheets/d/190drEOaOFH7cwCBM7eNB6vmnjICDiiShQiut7B55HCY/edit#gid=347850494)
- [x] Populate with dummy data
   - Project Officer, Team Leader, Division Director, Budget Team
- [x] Enable Django Admin interface to display the Roles table

### Additional Context

- [Database Models Spreadsheet](https://docs.google.com/spreadsheets/d/190drEOaOFH7cwCBM7eNB6vmnjICDiiShQiut7B55HCY/edit#)
- [ER Diagram](https://docs.google.com/drawings/d/106BT0UIW_-zmocmNWMoGbe4nVmHGU1fYfsQWnTccsiI/edit)

### Resources
* #91

alexsoble:
LGTM! 👍

alexsoble:
FWIW my understanding from our meeting with @ninamak this morning was that this work is 👍 👍 and good to go from her POV!

PR #174, opened 2021-09-08
Add Role model to view roles in django admin

carjug:
## Issue
#134

**NOTE**
This branch was made off of `cj/133/view-agencies`. To avoid merge conflicts, once #133 is merged into `main`, merge `main` into this branch.

## What changed
- Added Role model
- Seeded with test role data
- Added Role model view to Django Admin
- Role Name is set as display name in Django Admin

## Screenshots
Screen Shot 2021-09-08 at 4 47 52 PM
Screen Shot 2021-09-08 at 4 47 37 PM


carjug:
@alexsoble because the fixture data for roles was really small, converting the data to csv and then to json would have been more work than the work it took to just add it to the file manually, so still no CSV-to-JSON helper function.

alexsoble:
@carjug Yep, that makes sense! We only need the helper when we need the helper!

alexsoble:
I believe once we merge in `main`, the only new commit in this PR is 831cd229e48ba2addb7a76e215947836410f190d, which looks good to me! 

We got @ninamak's 👍 on this synchronously as well. 

User story for #134 says, "As a PM, I want to use this view to facilitate conversations about OPRE roles vs user types" — looks like this view can do that very well since we had a good conversation about the topic just now!

alexsoble (approved):

Issue #163, opened 2021-08-31
[2] As a dev, I want to make sure that only libraries we are really using are included in our Pipfile


alexsoble:
### Goals

As a dev, I want to make sure that only libraries we are really using are included in our Pipfile

### Tasks
- [ ] _Key pieces of work that will need to be done to accomplish our goal(s)_

### Additional Context
* _Any additional information that would be useful to share about this work (e.g., dependencies, constraints, previous work that's been done, info already known, etc)_

### Resources
* _Links to useful related info (e.g., articles, help docs, etc)_

alexsoble:
Pulling this one in because Dependabot is opening a bunch of PRs to update libraries that I don't think we're using.

alexsoble:
We did this via #176!

PR #176, opened 2021-09-09
Remove libraries from Pipfile that we don't need at the top level

alexsoble:
# Issue 

+ #163

# How to test locally

1. Pull down this branch
2. `docker-compose down && docker-compose build && docker-compose up`
3. Verify that the local Django admin functionality works as expected

# Notes

Some of the libraries that were listed in our `Pipfile` may be sub-dependencies; in that case we don't need to list them explicitly at the `Pipfile` level. `Pipfile.lock` gathers and declares sub-dependencies: https://pipenv.pypa.io/en/latest/basics/#pipenv-lock

Some like `flake8` we intend to use in the future, but let's remove from the `Pipfile` until we've implemented and/or documented how we'll use them.

Some of the libraries listed may just be unused at this point.


Pipfile
```diff
@@ -4,37 +4,8 @@ verify_ssl = true
...
-psycopg2-binary = "==2.9.1"
-pycodestyle = "==2.7.0"
-pyflakes = "==2.3.1"
-pylint = "==2.10.2"
-pytz = "==2021.1"
-six = "==1.16.0"
-sqlparse = "==0.4.1"
-toml = "==0.10.2"
-virtualenv = "==20.7.2"
-wrapt = "==1.12.1"
```
carjug, line 35: Thank you!! Yay for line deletion :)

carjug (approved):

alexsoble:
@carjug Woohoo on the approval! Did you run through the "How to test locally" steps? If all looked good, please do the honor and hit merge when you have a sec! Thanks!

alexsoble:
(I don't think this one needs PM review since the goal is to have no feature- or user-facing changes whatsoever, just remove dependencies we aren't really using.)

carjug:
Tested locally! Looks good. Thanks, Alex 🎉

alexsoble:
@carjug Ty!

Issue #93, opened 2021-07-16
[8] View CANs


ninamak:
### User story
As a PM/PO, I want to see what it would look like to view and edit CAN data in the Django interface so that I can provide feedback on how well it meets our needs.  

As a PM, I want to show our PO our initial data model so that we can feedback and refine it.

### Conditions of Acceptance
- [x] I can use Django Admin in the dev environment to see how much money is available to OPRE and for what purposes (based on dummy data).
- [x] I can test out the different out-of-box viewing and editing functionalities (or lack thereof).
- [x] It would be ok if the PM can only view the interface on a dev's local.

### Tasks
- [x] Implement CAN Info model based on [CAN Info tab](https://docs.google.com/spreadsheets/d/190drEOaOFH7cwCBM7eNB6vmnjICDiiShQiut7B55HCY/edit#gid=1853066578)
- [x] Implement CAN Amount model based on [CAN Amount tab](https://docs.google.com/spreadsheets/d/190drEOaOFH7cwCBM7eNB6vmnjICDiiShQiut7B55HCY/edit#gid=979139056)
- [x] Populate the tables with the dummy data in [AirTable](https://airtable.com/tblC1gcLbcE5wou4Z/viwkZHevQYTEb1rRF?blocks=hide)
- [x] Display the dummy data as a single table using the admin interface ([see CAN (Display Only) tab](https://docs.google.com/spreadsheets/d/190drEOaOFH7cwCBM7eNB6vmnjICDiiShQiut7B55HCY/edit#gid=1040321356))

alexsoble:
Discussing this issue in Grooming with @ninamak. Nina is thinking that CANs are probably the most high value data for Sheila to start seeing and interacting with in development mode. For that reason, @ninamak is proposing a flow something like:

1. tackling CANs views first without all of the Agency, People, Roles relationships in place
2. next, implement Agency, People, Roles tables and views
3. associate CANs with Agency, People, Roles tables

We want your feedback on this idea @amymok, let's discuss either during coworking tomorrow or eng coworking Wednesday!

ninamak:
[...]

PR #181, opened 2021-09-15
View CANs

carjug:
## Issue
#93 

## What changed
- Add model for CANInfo
- Add model for CANAmount
- Add Django Admin view for combined CANs data (see screenshot)
- Adds migrations for the two models
- Adds fixture data for CANInfo and CANAmount models

## How to review (for devs)
Compare schema in Django migration against Google Sheet model
Compare fake data in fixture to Airtable prototype
Run project locally and review Persons page in Django Admin

## Screenshots
Screen Shot 2021-09-17 at 6 20 03 PM
Screen Shot 2021-09-17 at 6 20 19 PM
Screen Shot 2021-09-17 at 6 20 38 PM


opre_ops/ops_site/models.py
```diff
@@ -28,5 +31,43 @@ class Meta:
...
-    
     full_name = property(display_name)
+
+    def __str__(self):
+        return self.full_name
+
+
+ARRANGEMENT_TYPES = [("1", "OPRE Appropriation"), ("2", "Cost Share"), ("3", "IAA"), 
+                    ("4", "IDDA"), ("5", "MOU")]
+class CANInfo(models.Model):
```
amymok, line 42: (nitpick) a newline here.  Although we haven't set up circleci, I wonder if we want to run flake8 before we merge in a pull request manually.  I don't think there's any bad styling, just that it may snowball into a bigger task if we haven't done it since the beginning.

amymok (approved):
LGTM!

opre_ops/ops_site/models.py
alexsoble, line 42: [_suggestion_]: A docstring comment here explaining what a CAN is, plus any other important information that a dev working on this model should know about the CANInfo model.

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
```diff
@@ -0,0 +1,47 @@
...
+        migrations.CreateModel(
+            name='CANInfo',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('number', models.CharField(max_length=30)),
+                ('description', models.CharField(max_length=100)),
+                ('nickname', models.CharField(max_length=30)),
+                ('purpose', models.TextField(null=True)),
+                ('arrangement_type', models.CharField(choices=[('1', 'OPRE Appropriation'), ('2', 'Cost Share'), ('3', 'IAA'), ('4', 'IDDA'), ('5', 'MOU')], max_length=30)),
+                ('authorizer', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='authorizer', to='ops_site.agency')),
```
alexsoble, line 23: [_praise_] The Google Sheet describing this table says "On_delete(SET_NULL) or (PROTECT)". I think Protect was the right choice here! In general I see us using soft delete strategies, where we retain records in the database and store some state like "active"/"inactive" rather than hard-deleting database entries. I like the idea of an error being raised if an operation tries to delete an agency referenced by CAN data.

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
```diff
@@ -0,0 +1,47 @@
...
+            ],
+            options={
+                'verbose_name_plural': 'CANs',
+            },
+        ),
+        migrations.CreateModel(
+            name='CANAmount',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('fiscal_year', models.CharField(max_length=5)),
```
alexsoble, line 34: [_question_] The Google Sheet spec says DateField, not CharField. Which do we want to use here, and why? 

Are there advantages to CharField? If we go with CharField, should the max_length be 4 and not 5? 

My assumption is that using a DateField (or even an integer) would let us do math — e.g. `+= 1 year` operations — which we can't do on CharFields.

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
```diff
@@ -0,0 +1,47 @@
...
+                ('amount_available', models.DecimalField(decimal_places=2, max_digits=10)),
+                ('amount_budgeted', models.DecimalField(decimal_places=2, max_digits=10)),
+                ('additional_amount_anticipated', models.DecimalField(decimal_places=2, max_digits=10)),
```
alexsoble, line 37: [_thought, non-blocking_]: I notice that the Google Sheet spec doesn't specify a maximum amount for these values, but this implementation does. This would restrict these database fields to amounts below $100 million, because the decimal places count towards the maximum. 

$100 million or below seems reasonable. If I were the implementer I might have set the maximum higher? Non-blocking, easy enough to adjust if we find that the real data includes larger sums than we expected.

opre_ops/ops_site/models.py
```diff
@@ -28,5 +31,43 @@ class Meta:
...
+    purpose = models.TextField(null=True)
+    arrangement_type = models.CharField(max_length=30, choices=ARRANGEMENT_TYPES)
+    source = models.ManyToManyField(Agency)
+    authorizer = models.ForeignKey(Agency, on_delete=models.PROTECT, related_name="authorizer")
+
+    class Meta:
+        verbose_name_plural = "CANs"
+
+
+class CANAmount(models.Model):
```
alexsoble, line 55: [_suggestion_]: Parallel suggestion to the one above ~ a docstring comment explaining what a CAN is, plus any other important information that a dev working on this model should know about the CANAmount model.

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
```diff
@@ -0,0 +1,47 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='CANInfo',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('number', models.CharField(max_length=30)),
+                ('description', models.CharField(max_length=100)),
+                ('nickname', models.CharField(max_length=30)),
+                ('purpose', models.TextField(null=True)),
```
alexsoble, line 21: [_issue, blocking_]: Following up on our conversation in Slack, suggest that we use `blank=True` here instead. This should resolve the issue we saw where CANs fail to save in Django Admin.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#null
https://docs.djangoproject.com/en/3.2/ref/models/fields/#blank

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
```diff
@@ -0,0 +1,47 @@
...
+        ),
+        migrations.CreateModel(
+            name='CANAmount',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('fiscal_year', models.CharField(max_length=5)),
+                ('amount_available', models.DecimalField(decimal_places=2, max_digits=10)),
+                ('amount_budgeted', models.DecimalField(decimal_places=2, max_digits=10)),
+                ('additional_amount_anticipated', models.DecimalField(decimal_places=2, max_digits=10)),
+                ('notes', models.TextField()),
```
alexsoble, line 38: [_issue, blocking_]: Similar to the issue about the `purpose` field above. Suggest that we use `blank=True` here. This should resolve an issue where CAN amounts fail to save in Django Admin.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#null
https://docs.djangoproject.com/en/3.2/ref/models/fields/#blank

alexsoble (changes requested):

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
carjug: Ah yes, I meant to ask the group about this. I'm not positive what we should do here. We'll never use a full date for just the year. If we need to do math on the year, I can possibly be convinced, but the DateField requires a month and day to be specified, which to me did not seem like something we would want or need. But, yes, open to discussion!

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
carjug: Ah, wait, I see. The decimal places gets factored into the max digits (didn't realize this).

alexsoble: @carjug 👍 👍 on `max_digits=12`

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
alexsoble: Would be interested to hear @amymok's thoughts when she is back in the office because she authored the Google Sheet spec. 

I wonder if integer gives us the best of all worlds — you can get the next fiscal year given a current year if you need it (`next = current + 1`), no extraneous month or day data involved, some built-in validation:

```python
>>> int('garbage input')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: 'garbage input'
```

carjug: @amymok do you have thoughts here?

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
carjug: @alexsoble I think this is resolved now, can you double check

carjug:
Screen Shot 2021-09-22 at 6 53 10 PM
Screen Shot 2021-09-22 at 6 53 18 PM

@amymok @alexsoble _I think_ I addressed most of the review feedback, plus the requested changes from Nina here. Can you all run through the changes again and LMK? Still wondering on Datefield question for `fiscal_year` but everything else should be resolved, I believe.

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
alexsoble: Yep, this looks resolved when I try editing/saving CANs locally!

Pipfile
```diff
@@ -6,6 +6,7 @@ name = "pypi"
 [packages]
 Django = "==3.2.6"
 psycopg2-binary = "==2.9.1"
+django-model-utils = "*"
```
alexsoble, line 9: [_suggestions, blocking_]: I notice a new library introduced in the last round of changes. Two notes/suggestions here:

1. Importing the library with an asterisk in place a version means that no version is being pinned. This means that updating our dependencies can introduce breaking changes without us knowing. Let's pin to the latest major version when we introduce a new library. The Eng Practices Guide has a little bit about pinning dependencies here: https://engineering.18f.gov/python/.
2. It looks like the only import from this library is `Choices`. The [django-model-utils docs for Choices](https://django-model-utils.readthedocs.io/en/latest/utilities.html#choices) say: "Django 3.0 adds enumeration types. These provide most of the same features as Choices." Our project is using Django 3.2.6, so I would suggest we use Django-provided built-in `Choices` rather than a module from an external library. If there are specific features we need from django-model-utils that we can't get straight from Django, let's document those. Otherwise, the fewer external libraries we use the better off we are in terms of both security and complexity. 

Summary: I would suggest we remove the library if possible and use built-in Django features instead. If the library is needed, let's pin to a major version.

alexsoble (changes requested):

alexsoble:
@carjug All of my review feedback from the first round is addressed except for the discussion about the fiscal year data type.  I also left suggestions about the introduction of a new library since my first review. So, nearly there! Just two things to resolve from my point of view.

Pipfile
carjug: Ok sounds good. I didn't know about the enumeration types in Django 3.0. Will make those changes now.

carjug:
@alexsoble yep just waiting to hear from @amymok about the `fiscal_year`.

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
amymok: Oh oops, definitely missed this.  I chose DateField under the impression that it gives us more validation out of the box.  But since we are only using the year field, it may not be necessary because we probably don't need time delta or any of those.  So we can be flexible.  If we are giving the choices of CharField and IntegerField.  IntegerField would be better both for us to increment as well as to validate.

Here're some resources on how people have done it with choices:

- https://newbedev.com/year-field-in-django
- https://stackoverflow.com/questions/1517474/only-showing-year-in-django-admin-a-yearfield-instead-of-datefield/24656072#24656072

Pipfile
carjug: @alexsoble so I just added the list of choices tuples back in that was there in the original code for the PR and moved them inside the relevant class. I was using `Choices` from the imported library to try and simplify the code (AKA not need a tuple for a single choice, and instead just a single string) but I understand your concern about unnecessary libraries. If I were to use an Enumeration class such as TextChoices, IMO the benefits do not outweigh the overly complicatedness and, also, the TextChoices class gives me less control over how things are displayed in the UI, i.e. Django automatically uppercases the first letter and lower cases the rest so what should display as `OPRE Appropriation` displays as `Opre Appropriation` in the UI and it's more complicated to overwrite this than it's worth. So back to the simple way!

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
carjug: Ok done! Changed to integer field/

carjug: I actually ended up changing it to a DecimalField so that I could set a max_digits on it, max_length does not work on IntegerField.

opre_ops/ops_site/admin.py
```diff
@@ -50,7 +50,7 @@ class CANAmount(admin.ModelAdmin):
                     "team_leader", "can_division", "notes")
 
     def can_display_name(self, obj):
-        return obj.can.number + " (" + obj.can.nickname + ")" + " - " + obj.fiscal_year
+        return obj.can.number + " (" + obj.can.nickname + ")" + " - " + str(obj.fiscal_year)
```
amymok, line 53: [thought][non-blocking] There's something in Python 3 that called [f-string](https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/).  It makes the string looks cleaner.

opre_ops/ops_site/migrations/0007_auto_20210923_1822.py
```diff
@@ -36,7 +36,7 @@ class Migration(migrations.Migration):
             name='CANAmount',
             fields=[
                 ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
-                ('fiscal_year', models.IntegerField(max_length=4)),
+                ('fiscal_year', models.DecimalField(decimal_places=0, max_digits=4)),
```
amymok, line 39: [suggestion][blocking]: The [resource here](https://newbedev.com/year-field-in-django) suggests the use of validating and bounding the year.  We should consider using some form of that, so we can prevent numbers that are not year to be rejected before saving the record.

opre_ops/ops_site/models.py
```diff
@@ -73,7 +73,7 @@ class CANAmount(models.Model):
     information by fiscal year for a given CAN
     """
     can = models.ForeignKey(CANInfo, on_delete=models.PROTECT)
-    fiscal_year = models.IntegerField(max_length=4)
+    fiscal_year = models.DecimalField(max_digits=4, decimal_places=0)
```
amymok, line 76: [suggestion][blocking]: The resource here suggests the use of validating and bounding the year. We should consider using some form of that, so we can prevent numbers that are not year to be rejected before saving the record.

carjug: @amymok sounds good, what year do you think we should set as the upper and lower limits??

Pipfile
alexsoble: @carjug Gotcha! I haven't done a deep dive and might be wrong, but it looks like [TextChoices](https://docs.djangoproject.com/en/3.0/ref/models/fields/#enumeration-types) could give you more control over how choices are displayed, by surfacing a human readable `label` option? 

I'll marked this as resolved for now, since I am happy with any solution that uses either no new libraries, or a new library pinned to a specific version. 

Thank you for implementing this feedback @carjug!

carjug: Yeah I tried using the `label` option, but it was not working as one would expect.

opre_ops/ops_site/models.py
alexsoble: Already posted this in Slack, but I'm not sure this suggestion should be a blocking one. The data model we are implementing here doesn't say anything about a max and a min value for fiscal years. Seems to me that data validation for fiscal years would fall into a different ticket / scope of work.

opre_ops/ops_site/admin.py
alexsoble: Fwiw I like this non-blocking suggestion, f-strings make interpolated strings like this much easier to read.

opre_ops/ops_site/models.py
amymok: I am ok with this as well.

opre_ops/ops_site/migrations/0007_canamount_caninfo.py
alexsoble: My first initial reaction was that DecimalField with `decimal_places=0` seems a little less idiomatic than an IntegerField for storing years. I pulled down this latest code using DecimalField and noticed that Django Admin formats the fiscal year correctly. No weird decimal places anywhere. I was able to increment the fiscal year seamlessly using the Django Admin UI. And I see how it gets us some easy-to-implement `max_length` validation @carjug. So I don't have any blocking suggestions or thoughts here! This isn't exactly what's in the Google Sheets spec but it doesn't seem unreasonable. Open to your input/feedback as one of the co-designers of the data model @amymok.

carjug: The reason I used DecimaField instead of Integer is that IntegerField does not have a way to set a max_length. You can pass in `max_length` as a kwarg to `IntegerField` but it gets ignored. @alexsoble

amymok: I would suggest we use `IntegerField` for a few reasons, (but also do not have the exact proofs though):
- it is more idiomatic to use `IntegerField`
- there may be unnecessary performance hit when we do comparison and filtering when using `DecimalField`
- we can use the `MidValueValidator` and `MaxValueValidator` to bound the field once we figure out what validation we want to do, I would not worry about needing it to have a `max_length` right now because there are still a big gap on what really is the validate numbers within that

opre_ops/ops_site/admin.py
```diff
@@ -4,20 +4,74 @@
...
+    display_can_name.short_description = "CAN Name"
+
+@admin.register(CANAmount)
+class CANAmount(admin.ModelAdmin):
+    list_display = ("can_display_name", "can_description", "can_purpose", "amount_budgeted", "amount_available",
+                    "additional_amount_anticipated", "can_arrangement_type", "can_source", "can_authorizer", 
+                    "team_leader", "can_division", "notes")
+
+    def can_display_name(self, obj):
+        return f"{obj.can.number} ({obj.can.nickname}) - {str(obj.fiscal_year)}"
```
amymok, line 53: [suggestion][blocking] removing `str()` should work

carjug: Ok should be good @amymok . If it looks good can you 👍 approve :)

carjug:
@alexsoble you're still showing as changes requested, can you approve if everything looks good to you?

amymok:
It looks like Arrangement Type and OPRE Division are showing indices
Screen Shot 2021-09-24 at 10 14 09 AM

I am also getting some errors in migration when I bring up the system.
```
➜  OPRE-OPS git:(cj/93/view-cans) ✗ docker-compose up   
Starting opre-ops_db_1 ... done
Starting opre-ops_migration_1 ... done
Starting opre-ops_web_1       ... done
Attaching to opre-ops_db_1, opre-ops_migration_1, opre-ops_web_1
migration_1  | checking for database on db:5432
migration_1  | Connection to db (172.18.0.2) 5432 port [tcp/postgresql] succeeded!
migration_1  | found database on db:5432!
db_1         | 
db_1         | PostgreSQL Database directory appears to contain a database; Skipping initialization
db_1         | 
db_1         | 2021-09-24 17:16:14.280 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db_1         | 2021-09-24 17:16:14.280 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db_1         | 2021-09-24 17:16:14.285 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1         | 2021-09-24 17:16:14.303 UTC [26] LOG:  database system was interrupted; last known up at 2021-09-24 17:12:34 UTC
db_1         | 2021-09-24 17:16:14.596 UTC [26] LOG:  database system was not properly shut down; automatic recovery in progress
db_1         | 2021-09-24 17:16:14.599 UTC [26] LOG:  redo starts at 0/17891C0
db_1         | 2021-09-24 17:16:14.599 UTC [26] LOG:  invalid record length at 0/17892A0: wanted 24, got 0
db_1         | 2021-09-24 17:16:14.599 UTC [26] LOG:  redo done at 0/1789268
db_1         | 2021-09-24 17:16:14.614 UTC [1] LOG:  database system is ready to accept connections
db_1         | 2021-09-24 17:16:14.796 UTC [33] LOG:  incomplete startup packet
migration_1  | Operations to perform:
migration_1  |   Apply all migrations: admin, auth, contenttypes, ops_site, sessions
migration_1  | Running migrations:
db_1         | 2021-09-24 17:16:15.446 UTC [35] ERROR:  relation "ops_site_caninfo" already exists
db_1         | 2021-09-24 17:16:15.446 UTC [35] STATEMENT:  CREATE TABLE "ops_site_caninfo" ("id" bigserial NOT NULL PRIMARY KEY, "number" varchar(30) NOT NULL, "description" varchar(100) NOT NULL, "purpose" text NOT NULL, "nickname" varchar(30) NOT NULL, "arrangement_type" varchar(30) NOT NULL, "authorizer_id" bigint NOT NULL)
migration_1  |   Applying ops_site.0007_auto_20210924_1700...Traceback (most recent call last):
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 82, in _execute
migration_1  |     return self.cursor.execute(sql)
migration_1  | psycopg2.errors.DuplicateTable: relation "ops_site_caninfo" already exists
migration_1  | 
migration_1  | 
migration_1  | The above exception was the direct cause of the following exception:
migration_1  | 
migration_1  | Traceback (most recent call last):
migration_1  |   File "/opre_project/manage.py", line 23, in <module>
migration_1  |     main()
migration_1  |   File "/opre_project/manage.py", line 19, in main
migration_1  |     execute_from_command_line(sys.argv)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
migration_1  |     utility.execute()
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
migration_1  |     self.fetch_command(subcommand).run_from_argv(self.argv)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
migration_1  |     self.execute(*args, **cmd_options)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 398, in execute
migration_1  |     output = self.handle(*args, **options)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 89, in wrapped
migration_1  |     res = handle_func(*args, **kwargs)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 244, in handle
migration_1  |     post_migrate_state = executor.migrate(
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 117, in migrate
migration_1  |     state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 147, in _migrate_all_forwards
migration_1  |     state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 227, in apply_migration
migration_1  |     state = migration.apply(state, schema_editor)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/migration.py", line 126, in apply
migration_1  |     operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/operations/models.py", line 92, in database_forwards
migration_1  |     schema_editor.create_model(model)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/schema.py", line 345, in create_model
migration_1  |     self.execute(sql, params or None)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/schema.py", line 145, in execute
migration_1  |     cursor.execute(sql, params)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 98, in execute
migration_1  |     return super().execute(sql, params)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 66, in execute
migration_1  |     return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
migration_1  |     return executor(sql, params, many, context)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
migration_1  |     return self.cursor.execute(sql, params)
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/utils.py", line 90, in __exit__
migration_1  |     raise dj_exc_value.with_traceback(traceback) from exc_value
migration_1  |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 82, in _execute
migration_1  |     return self.cursor.execute(sql)
migration_1  | django.db.utils.ProgrammingError: relation "ops_site_caninfo" already exists
migration_1  | 
opre-ops_migration_1 exited with code 1
web_1        | Watching for file changes with StatReloader
web_1        | Performing system checks...
web_1        | 
web_1        | System check identified no issues (0 silenced).
web_1        | 
web_1        | You have 1 unapplied migration(s). Your project may not work properly until you apply the migrations for app(s): ops_site.
web_1        | Run 'python manage.py migrate' to apply them.
web_1        | September 24, 2021 - 17:16:16
web_1        | Django version 3.2.6, using settings 'opre_ops.settings.local'
web_1        | Starting development server at http://0.0.0.0:8080/
web_1        | Quit the server with CONTROL-C.
```

amymok:
Okay sorry, it looks like I didn't clean the docker container.  It is good after I do `docker-compose down` then re-build/up.

LGTM!

alexsoble (approved):
LGTM! 🚀  I will 🚢 since Amy and Nina have both approved!

PR #182, opened 2021-09-15
Add comma separated display of roles in django admin

carjug:
Adds comma separated display of roles for Person in Django Admin.

**note**: could not figure out an elegant/easy way to display the roles on newlines. I tried using `"\n"` in the string builder in the `show_roles()` function as well as injecting `html` using the django `linebreaks()` function and neither produced desired results. 

I landed on comma separation (see below) but open to other's thoughts about how best to proceed.

Screen Shot 2021-09-15 at 1 25 08 PM

### DIFF
--- a/opre_ops/ops_site/admin.py
+++ b/opre_ops/ops_site/admin.py
@@ -1,5 +1,6 @@
 
 from django.contrib import admin
+
 from ops_site.models import Agency
 from ops_site.models import Role
 from ops_site.models import Person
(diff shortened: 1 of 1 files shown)


amymok (approved):
LGTM!

alexsoble (approved):
LGTM! 

I tested out adding multiple roles to a person, and it worked great! 

I also tried removing all roles from a person and I got some nice data validation errors:

Screen Shot 2021-09-15 at 3 47 23 PM

carjug:
Thanks y'all! Merging now.

Issue #199, opened 2021-10-04
Refine CANs - Part 2


ninamak:
### Conditions of Acceptance
~~- [ ] When more than one CAN Lead is assigned, I see them displayed in the order in which they were entered.~~  
  ~~- [ ] When the people belong to different Divisions, I see the Divisions displayed in the same order as the CAN Leads~~
    ~~- Note: I'm not sure how much is worth investing in improving this here since this won't ultimately be the UI users see.  I'm inclined to say that if this is an easy tweak, then why not.  But if it's going to take research and/or effort, it can wait.~~
- [x] When the Division is DCFD, it displays as "DCFD" and not "DCFC"
- [x] "Source" is renamed "Funding Source"
- [x] "Agencies" is renamed to "Funding Partners"

carjug:
@ninamak I would say the first task is a bit more work than it's worth. I could spend today and tomorrow implementing it but I agree, given that the django admin is not going to be the UI, it's not worth investing the time/energy to do that on this particular page -- we can definitely do it down the line though.

ninamak:
@carjug Sounds good!  Thanks for that update.  I'll strikethrough that CoA.

carjug:
<img width="1430" alt="Screen Shot 2021-10-05 at 12 46 56 PM" src="https://user-images.githubusercontent.com/10871517/136270898-123d9c0e-23aa-4239-bc66-54a5c07fa7b5.png">
<img width="1438" alt="Screen Shot 2021-10-05 at 12 47 05 PM" src="https://user-images.githubusercontent.com/10871517/136270905-4d0202b5-3508-4862-aede-aee0295ccb4a.png">

Screenshots for posterity!

PR #201, opened 2021-10-07
Refine CANs pt 2

carjug:
## What changes

- Displays division as "DCFD", not "DCFC"
- "Source" is renamed "Funding Source"
- "Agencies" is renamed to "Funding Partners"

## Issue

#199 

## How to test

- Pull down branch
- Build locally
- Navigate to `CANs` page and see changes named above

## Screenshots
Screen Shot 2021-10-05 at 12 46 56 PM
Screen Shot 2021-10-05 at 12 47 05 PM


amymok (approved):
LGTM!

[question]: It looks like the migration files are being removed and regenerated by editing the first initial migration file with all the changes here.  I am curious why it was done this way?  This is probably okay for now since we don't have too much, but am thinking about down the line when we have more migrations, may be even with data migrations, this may be a potential issue to clear out everything rather than build on previous migrations?

carjug:
Merging!

PR #217, opened 2021-11-16
Setup automated deploys to development environment

carjug:
## What changes

Deploys app code to cloud.gov sandbox from CircleCI. 

## How to test

See details on this PR.

## Screenshots

Screen Shot 2021-11-23 at 1 20 07 PM

## Links

https://app.circleci.com/pipelines/github/HHS/OPRE-OPS/58/workflows/9418cd91-4672-423d-b4e4-831e74c207f2/jobs/63


.circleci/config.yml
```diff
@@ -8,3 +8,23 @@ jobs:
...
+          name: Install cf cli
+          command: |
+            mkdir -p $HOME/bin
+            export PATH=$HOME/bin:$PATH
+            curl -L "https://cli.run.pivotal.io/stable?release=linux64-binary" | tar xzv -C $HOME/bin
+            cf install-plugin autopilot -f -r CF-Community
+      - run:
+          name: Deploy to cloud.gov
+          command: |
+            if [[ -n "${CIRCLE_BRANCH}" && "${CIRCLE_BRANCH}" != "master" ]]; then
```
amymok, line 21: [issue] `"master"` should `"main"`

.circleci/config.yml
```diff
@@ -8,3 +8,23 @@ jobs:
...
+              export PATH=$HOME/bin:$PATH
+              cf login -a https://api.fr.cloud.gov -u $DEV_USER -p $DEV_PASSWORD -o $ORG_NAME -s $SPACE_NAME
+              echo "Login successful."
+              cf zero-downtime-push opre-ops-test -f manifest.yml
+              echo "Deploy successful."
+            else
+              echo "Skipped"
+            fi
```
amymok, line 29: [question]: So what we are doing here is that we are deploying to this cloud environment when something is being push to some branch.  This will be triggered?

Pipfile
```diff
@@ -6,6 +6,7 @@ name = "pypi"
 [packages]
 Django = "==3.2.6"
 psycopg2-binary = "==2.9.1"
+cfenv = "*"
```
amymok, line 9: [nitpick]: lock this version

opre_ops/opre_ops/settings/cloudgov.py
```diff
@@ -45,7 +43,7 @@ def get_json_env_var(variable_name):
 }
 
 # SECURITY: Keep the secret keys used in production secret!
-SECRET_KEY = user_provided_env['DJANGO_SECRET_KEY']
+SECRET_KEY = env.get_credential('APP_SECRET_KEY', generate_random_string(50))
```
amymok, line 46: [question] Are we setting `'APP_SECRET_KEY'` as user-provided environment variable on cloud.gov?  If not, it may not be kept between deployment, and is that okay if we always use a random string?

amymok (changes requested):

.circleci/config.yml
carjug: Ah good catch, I had it set up this way to test, but we should make it so on a merge to a new `development` branch this gets deployed. Does that sound good to you?

amymok: I am looking back on what we agreed on as our deployment process.  In [continuous deployment to dev](https://github.com/HHS/OPRE-OPS/blob/main/docs/how_we_work/deploy_flow.md#continuous-deployment-to-dev):
>Our plan is to allow any dev to trigger a deploy to the dev app via a git tag. (See "Executing workflows for a git tag" in the Circle docs.) This will give devs the flexibility to deploy to dev and test out new code at any time, without waiting for a review process or a git merge process.

So it seems like we are choosing any git-tag to deploy to.  If we want to alter from this, we probably need to update this piece of documentation.

carjug: 🤔 oh man, I don't remember this at all lol. Many things have left my brain over the last couple of months.

opre_ops/opre_ops/settings/cloudgov.py
carjug: yeah this is something that will need to be addressed in the future ticket (I mentioned on Slack I ripped out what Alex had previously implemented here but it was incomplete so it was complicating this work).

PR #218, opened 2021-11-26
Update docs for circle

carjug:
## What changes

Adds documentation for CircleCI and cloud.gov deployment setup

## Links

https://github.com/HHS/OPRE-OPS/blob/1c80119671b47e8d129cc3d6f7978d1c525659a8/docs/recipes/setup_circleci.md

### DIFF
--- a/docs/adr/008-use-circle-for-ci-cd.md
+++ b/docs/adr/008-use-circle-for-ci-cd.md
@@ -12,29 +12,29 @@ Accepted
 
 ## Context and explanation of the tech concepts
 
-We need a tool to power our CI/CD pipelines. 
+We need a tool to power our CI/CD pipelines.
 
-Continuous integration (CI) will let us propose code changes on a consistent basis. Automated tests will run on every pull request, keeping code quality high and checking for regressions. 
+Continuous integration (CI) will let us propose code changes on a consistent basis. Automated tests will run on every pull request, keeping code quality high and checking for regressions.
 
 Continuous deployment (CD) will let us create an automated deployment path, making the release of new code more predictable, easier for developers, and less error-prone than manual deployments.
 
-### Options considered 
+### Options considered
 
 #### CircleCI
 
 * The Department of Health and Human Services (HHS) owns and operates a CirlceCI organization. This makes CircleCI  a natural choice, because we know it is compliant with HHS standards.
 * Several other projects working within the Administration for Children and Families (ACF) are successfully using CircleCI for CI/CD.
 
-#### GitHub Actions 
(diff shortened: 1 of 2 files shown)


docs/adr/008-use-circle-for-ci-cd.md
```diff
@@ -12,29 +12,29 @@ Accepted
```
carjug, line 12: All the changes on this file were to just fix some whitespace issues.

docs/recipes/setup_circleci.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+The last command will return a username/password like this:
+```
+{
+ "password": "oYasdfliaweinasfdliecV",
+ "username": "deadbeed-aabb-1234-feha0987654321000"
+}
+```
+These are the credentials you will need to insert into the CircleCI OPRE-OPS project environment variables the names of which will need to correspond to the environment variables in the deploy scripts in `.circleci/config.yml`
```
amymok, line 30: [nitpick] Maybe adding something like:
>i.e. $DEV_USER and $DEV_PASSWORD for our case

makes it a bit more explicit what you are referring to.

docs/recipes/setup_circleci.md
```diff
@@ -0,0 +1,30 @@
+# How to: setup CircleCI to deploy to cloud.gov
+
+The cloud.gov docs recommend using a [space-deployer](https://cloud.gov/docs/services/cloud-gov-service-account/) to deploy applications from continuous integration services such as CircleCI.
```
amymok, line 3: [thought] I wonder if we should say something about how this is already completed, and in what case would you need to revisit.

carjug: Yes good call!

docs/recipes/setup_circleci.md
carjug: I went back and forth on this because they'll need to do this for not just $DEV_ but it might be better to be more explicit.

amymok: I was thinking it may not be apparent what we are referring to.  If we don't want to use `$DEV_USER`, may be specified that this is the credentials you need to supply when logging into cloud.gov

carjug: Between the two, I think I'll just use `$DEV_USER` for now, because these generated creds are not something they should use when logging into cloud.gov

carjug:
@amymok can you come back to this when you get a sec for a final pass? I addressed comments :)

amymok (approved):
This looks great to me!

PR #236, opened 2021-12-17
Development

carjug:
## What changes

Deploys app code to cloud.gov sandbox from CircleCI. 

## How to test

See details on this PR.

## Screenshots

Screen Shot 2021-11-23 at 1 20 07 PM

## Links

https://app.circleci.com/pipelines/github/HHS/OPRE-OPS/77/workflows/643c1ff4-d2c8-43b1-a045-634a19d4846a/jobs/81

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,6 @@ jobs:
             cf install-plugin autopilot -f -r CF-Community
       - run:
           name: Deploy to cloud.gov
-          filters:
           command: |
             if [[ -n "${CIRCLE_BRANCH}" && "${CIRCLE_BRANCH}" == "development" ]]; then
               export PATH=$HOME/bin:$PATH
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,3 +8,26 @@ jobs:
       - run:         
           name: Run Tests
           command:  docker-compose run web pytest --cov-config=.coveragerc --cov=ops_site --cov-report term-missing
+      - run:
+          name: Install cf cli
+          command: |
+            mkdir -p $HOME/bin
+            export PATH=$HOME/bin:$PATH
+            curl -L "https://cli.run.pivotal.io/stable?release=linux64-binary" | tar xzv -C $HOME/bin
+            cf install-plugin autopilot -f -r CF-Community
+      - run:
+          name: Deploy to cloud.gov
+          filters:
+            tags:
+              only: /^v*/
+          command: |
+            if [[ -n "${CIRCLE_BRANCH}" && "${CIRCLE_BRANCH}" == "development"}" ]]; then
+              export PATH=$HOME/bin:$PATH
+              cf login -a https://api.fr.cloud.gov -u $DEV_USER -p $DEV_PASSWORD -o $ORG_NAME -s $SPACE_NAME
+              echo "Login successful."
+              cf zero-downtime-push opre-ops-test -f manifest.yml
+              echo "Deploy successful."
+            else
+              echo "Skipped"
+            fi
+
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,8 +18,6 @@ jobs:
       - run:
           name: Deploy to cloud.gov
           filters:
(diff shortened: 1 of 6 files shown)


carjug:
@amymok this is everything from the previous pull request but just merging from `development` to `main`

carjug:
Still need to update the ADR, so I will let you know when that part is ready @amymok

amymok:
@carjug This looks good so far.  Do you plan to update the ADR in this PR or in another PR?

carjug:
@amymok I'll do it in this PR :) coming right up!

carjug:
@amymok doc updated :)

.circleci/config.yml
```diff
@@ -8,3 +8,26 @@ jobs:
...
+          command: |
+            mkdir -p $HOME/bin
+            export PATH=$HOME/bin:$PATH
+            curl -L "https://cli.run.pivotal.io/stable?release=linux64-binary" | tar xzv -C $HOME/bin
+            cf install-plugin autopilot -f -r CF-Community
+      - run:
+          name: Deploy to cloud.gov
+          filters:
+            tags:
+              only: /^v*/
```
amymok, line 22: [issue] I don't think this belongs here.  `filters` are for workflow/jobs, so it goes to a level up.  And I think this is not correct either to only look for tags that begins with `v`.
[suggestion] This should be removed.

docs/how_we_work/deploy_flow.md
```diff
@@ -51,25 +51,25 @@ Since development is a space for experimental changes, don't count on the develo
 
 For initial development and testing, we are currently using a sandbox on cloud.gov: https://cloud.gov/pricing/.
 
-### Continuous deployment to dev 
+### Continuous deployment to dev
 
 Unless we create many dev apps and write functionality similar to [Heroku Review Apps](https://devcenter.heroku.com/articles/github-integration-review-apps) (one app per in-progress PR), we will only have one dev app for an engineering team of three. How should devs trigger deploys to the dev app?
 
-Our plan is to allow any dev to trigger a deploy to the dev app via a [git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging). (See ["Executing workflows for a git tag" in the Circle docs](https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag).) This will give devs the flexibility to deploy to dev and test out new code at any time, without waiting for a review process or a git merge process. 
+Our plan is to allow any dev to trigger a deploy to the dev app via a merge to the `development` branch.
```
amymok, line 58: [question] I think it runs even before it gets merged, it will run in a PR if it attempts to merge right?  We may want to express that too.

amymok (changes requested):

docs/how_we_work/deploy_flow.md
carjug: No, I don't think so. We're only deploying to the development environment if it gets merged into development, otherwise we would overwrite development with each new PR to the development branch and I don't think we want that, right? @amymok

carjug: @amymok I made a junk change with a junk PR #237  just to verify the above is true, and you can see the deploy is skipped if the branch is not `development`. LMK if this is not what you expected.

Screen Shot 2021-12-22 at 10 08 36 AM

amymok: Oh good good.  Thanks for verifying.  Yes, we only want to deploy when it is merged.  But tests will run regardless.

.circleci/config.yml
```diff
@@ -18,8 +18,6 @@ jobs:
       - run:
           name: Deploy to cloud.gov
           filters:
```
amymok, line 20: [nitpick]: This should probably be removed too since there's nothing to filter.

amymok (changes requested):

.circleci/config.yml
carjug: oh whoops! meant to delete that. will do :)

amymok (approved):
LGTM!

PR #241, opened 2022-01-19
Add pipenv verify and install to README

jskinne3:
## What changes

An addition to the README, describing the process I needed to use pipenv

## How to test

Run the commands in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Then navigate to http://localhost:8080 in your browser
 
 ### Dependency management with pipenv
 
-We use pipenv locally, not in Docker. To verify pipenv is installed and working, run:
+To verify pipenv is installed and working locally, run:
 ```
 pipenv graph
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,17 @@ Then navigate to http://localhost:8080 in your browser
 
 ### Dependency management with pipenv
 
+We use pipenv locally, not in Docker. To verify pipenv is installed and working, run:
```
carjug, line 23: The first part of this isn't technically true, refer to the `Dockerfile` to see how we use Pipenv to manage dependencies inside Docker. If one wants to modify the Pipfile however, they will need pipenv installed. So I think just removing the first statement will make this a-ok.

carjug (changes requested):

README.md
jskinne3: Great point! I have removed it

carjug (approved):
LGTM!

PR #246, opened 2022-03-14
Development

jskinne3:
## What changes

Accumulated development changes into trunk, including setting 700M memory allocation in the mainfest.yml file to address the error `app requested more memory than available` during deployment from CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
               export PATH=$HOME/bin:$PATH
               cf login -a https://api.fr.cloud.gov -u $DEV_USER -p $DEV_PASSWORD -o $ORG_NAME -s $SPACE_NAME
               echo "Login successful."
-              cf push opre-ops-test -f manifest.yml
+              cf zero-downtime-push opre-ops-test -f manifest.yml
               echo "Deploy successful."
             else
               echo "Skipped"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
               export PATH=$HOME/bin:$PATH
               cf login -a https://api.fr.cloud.gov -u $DEV_USER -p $DEV_PASSWORD -o $ORG_NAME -s $SPACE_NAME
               echo "Login successful."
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ jobs:
               export PATH=$HOME/bin:$PATH
               cf login -a https://api.fr.cloud.gov -u $DEV_USER -p $DEV_PASSWORD -o $ORG_NAME -s $SPACE_NAME
               echo "Login successful."
-              cf zero-downtime-push opre-ops-test -f manifest.yml
+              cf push opre-ops-test -f manifest.yml
```
amymok, line 25: [question]: Why is `zero-downtime-push` being removed?

manifest.yml
```diff
@@ -1,6 +1,7 @@
 ---
 applications:
   - name: opre-ops-test
+    memory: 700M
```
amymok, line 4: [todo] Can you include the reasons why we need this in the PR description?

amymok (changes requested):

.circleci/config.yml
jskinne3: To make deployment from local as similar as possible to CircleCI, and to reduce memory usage. I didn't confirm that the change is currently needed for a successful deployment, but I thought if we are running into memory problems it's good to start trimming down unnecessary things.

manifest.yml
jskinne3: Yes, great idea, thank you

.circleci/config.yml
jskinne3: Update: at your request have have confirmed that `zero-downtime-push` blocks deployment by exceeding the memory limit. So I have restored `push` in the config file, and I have also removed the [autopilot](https://github.com/contraband/autopilot) plugin, which was [deprecated years ago](https://engineering.18f.gov/continuous-deployment/).

amymok: Okay thank you for checking, it sounds like it is necessary to remove it.

amymok (approved):
LGTM!

PR #268, opened 2022-04-13
Propose the --insecure flag for runserver

jskinne3:
## What changes

Add a flag to `runserver` command that will allow it to server files during prototyping

## Issue

CSS and JS static assets not showing up

## How to test

Look at the login page and see if the CSS is broken

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -5,7 +5,7 @@ applications:
     buildpack: python_buildpack
     # TODO: Django docs say not to use the `runserver` command in prod;
     # switch over to a production-ready webserver like gunicorn
-    command: cd opre_ops && python manage.py migrate && python manage.py runserver 0.0.0.0:8080
+    command: cd opre_ops && python manage.py migrate && python manage.py collectstatic --noinput && python manage.py runserver --insecure 0.0.0.0:8080
 
 services:
  - opre-ops-psql-db


manifest.yml
```diff
@@ -5,7 +5,7 @@ applications:
     buildpack: python_buildpack
     # TODO: Django docs say not to use the `runserver` command in prod;
     # switch over to a production-ready webserver like gunicorn
-    command: cd opre_ops && python manage.py migrate && python manage.py runserver 0.0.0.0:8080
+    command: cd opre_ops && python manage.py migrate && python manage.py collectstatic --noinput && python manage.py runserver --insecure 0.0.0.0:8080
```
amymok, line 8: [suggestion]: If we decided to use `--insecure` flag, we should put more notes around this in terms of why we are using it and why it is okay, and in what situation we really should **NOT** use it for.  This will help developers later to know this needs to be revisited.
[suggestion]: Besides writing the note here, we would need another note in the README.md or somewhere very visible as a caution note that, this set up is only intended for local development only.

amymok (changes requested):

manifest.yml
jskinne3: Excellent suggestions! I've added notes

mgwalker (approved):
Added #269 to capture the need to switch from Django `runserver` to some other app runner, like gunicorn.

manifest.yml
mgwalker: I think this is resolved!

PR #271, opened 2022-05-23
Development

jskinne3:
## What changes

Serve static assets (CSS, JS) in prototype deployment with `collectstatic` and `runserver` commands

## How to test

* The deployed (prototype) environment has non-broken-looking CSS
* Deployment logs include output like `128 static files copied to '/opre_project/opre_ops/static'`


mgwalker (approved):
do it. now.

PR #274, opened 2022-06-01
Can view

jskinne3:
## What changes

Add basic view detailing a CAN. Add links among existing views.

## How to test

Ensure `/ops/cans/` lists CANs, which can be clicked on to view details


jskinne3:
> This makes sense. I guess the HTML is linked to the view class by Django naming convention?
> 
> `CANInfoListView` ➡️ `caninfo_list`

Yup! This is [documented implied behavior ](https://docs.djangoproject.com/en/4.0/topics/class-based-views/generic-display/#generic-views-of-objects)

PR #275, opened 2022-06-01
Data model updates and expansion

mgwalker:
## What changes

This PR updates and expands the data model to support all the data displayed in the [wireframe prototype](https://www.figma.com/proto/PKFlN0Bq23KkpNI325XSZk/OPS-CAN-wireframes?node-id=1439%3A12784&starting-point-node-id=1377%3A9247), and to make maintenance a bit easier.

- [x] renames `CANInfo` to `CommonAccountNumber`
- [x] renames `CANAmount` to `CANFiscalYear`
- [x] adds `Contract`
- [x] adds `ContractLineItem`
- [x] adds `ContractLineItemFiscalYear`
- [x] updates associated admin views
   - we may want to create an issue for adding the rest of the admin views, if we think that's important/valuable. I didn't add them all, for the sake of time.
- [x] updates fake data fixtures to account for table renames, and add contracts, line items, and FYs

You will need to migrate your database using the provided migration(s) after this is merged.


docker-compose.yml
```diff
@@ -10,6 +10,8 @@ services:
   migration:
     build: .
     command: bash -c "sh wait-for-postgres.sh && python manage.py migrate && python manage.py loaddata ops_site/fixtures/fake_data.json"
+    volumes:
+      - ./opre_ops:/${CI}opre_project
```
mgwalker, line 14: Similar to the `volumes` property in the other Docker service. This one allows the `migration` container to run the most recent migrations without having to rebuild it. You can instead run `docker compose up -d` at any time to apply the latest migrations.

mgwalker, line 14: Similar to the `volumes` property in the other Docker service. This one allows the `migration` container to run the most recent migrations without having to rebuild it. You can instead run `docker compose up -d` at any time to apply the latest migrations.

opre_ops/ops_site/admin.py
```diff
@@ -41,42 +134,64 @@ class CANInfoAdmin(admin.ModelAdmin):
...
+        "can_arrangement_type",
+        "can_authorizer",
+        "display_can_leads",
+        "can_division",
+        "notes",
+    )
 
     def can_display_name(self, obj):
         return f"{obj.can.number} ({obj.can.nickname}) - {obj.fiscal_year}"
+
```
mgwalker, line 162: A lot of these whitespace changes were introduced by Black autoformatter.

opre_ops/ops_site/fixtures/fake_data.json
```diff
@@ -1,22 +1,4 @@
...
-        "first_name": "", 
-        "last_name": "", 
-        "email": "email@test.com", 
-        "is_staff": true, 
-        "is_active": true, 
-        "date_joined": "2021-09-13T23:29:07.694Z", 
-        "groups": [], 
-        "user_permissions": []
-    }
-},
```
mgwalker, line 19: We can revert this if you want, but it seemed weird to have the admin user in the fixtures data. Plus I don't know its password so I had to create my own admin user anyway.

jskinne3: Ah, sorry, I could/should have told you the password. What alternate method would you use to create at least one user following a deployment?

docker-compose.yml
jskinne3: Do you think anything should be added to the README about this?

opre_ops/ops_site/models.py
```diff
@@ -30,46 +36,65 @@ class Meta:
...
+
+    def contracts_for_fiscal_year(self, fy):
+        cid = [
+            li.fiscal_year.line_item.contract.id
+            for li in self.line_items_fy.filter(fiscal_year__fiscal_year=fy)
+        ]
+        return Contract.objects.filter(id__in=cid)
 
     class Meta:
-        verbose_name_plural = "CANs Info"
```
jskinne3, line 61: Seeing this makes me so glad we are getting rid of the name `CAN Info`

jskinne3 (approved):

opre_ops/ops_site/models.py
mgwalker: Re: the README, I don't think so because it shouldn't matter for getting up and running or anything, but it should definitely be documented somewhere! Incoming commit that adds comments to the `docker-compose` file!

opre_ops/ops_site/fixtures/fake_data.json
mgwalker:
```sh
docker compose run web python manage.py createsuperuser
```

That'll walk you through the process of creating a local admin account. And since you asked this totally reasonable question, I'm also going to update the README since that will be a necessary step for someone setting up the environment for the first time.

PR #278, opened 2022-06-10
Add Django data model visualization

mgwalker:
## What changes

This PR updates our Django install to include the extra libraries necessary to extract a [DOT file](https://en.wikipedia.org/wiki/DOT_%28graph_description_language%29) representing the data model expressed with Django models. It also generates the current visualizations, adds them to the repo, displays them in the readme, and includes instructions for regenerating them in the future.


Pipfile
```diff
@@ -7,6 +7,9 @@ name = "pypi"
 Django = "==4.0.4"
 psycopg2-binary = "==2.9.3"
 cfenv = "0.5.3"
+pyparsing = "*"
+pydot = "*"
+django-extensions = {git = "https://github.com/django-extensions/django-extensions"}
```
mgwalker, line 12: We'll want to watch this. Eventually we should switch back to a pip-based install. However, the most recent version of `django-extensions` does not support Django 4. Django 4 support was pushed to main a few days ago, but they haven't published a release yet, so if we install directly from GitHub, we'll get a version that works for us.

README.md
```diff
@@ -69,3 +69,27 @@ docker-compose run web pytest --cov-config=.coveragerc --cov=ops_site --cov-repo
 This prototype uses `runserver` as a web server, which is considered insecure
 for production use. This should be replaced with something like `gunicorn` and
 `nginx` before it is deployed beyond prototyping purposes.
+
+## Data model
+
+![OPRE prototype data model](docs/models.png)
+
+To update this visualization, use the django-extensions module to create a new
+dotfile:
```
jskinne3, line 78: Might want to provide a link explaining what a dotfile is, and/or link to documentation for libraries used

jskinne3 (approved):
Very useful! I left one comment suggesting documentation and I am approving this PR

README.md
jskinne3: Also I'm unclear on the meanings of the linkage lines among the tables. What do the circles at the end of the lines represent? Are join tables / pivot tables represented here?

jskinne3 (approved):

README.md
mgwalker: I'll open another PR with some more documentation into the readme because that's smart. I think there's a link to the dotfile documentation in the Python itself but that's not super helpful unless you're reading code. 😛 

FWIW, I also don't know what the circles there are for. They're created by the [graph_models](https://django-extensions.readthedocs.io/en/latest/graph_models.html) command and the documentation doesn't clarify either. So maybe something else for the readme is to mention that so people won't try to read too much into the diagram. Something like:

> This diagram shows all of the tables and which ones have relationships between one another. However, by itself, this diagram should not be used to try to fully understand the model or the nature of the relationships between tables. Instead, for that deeper understanding, this diagram can be used as an aid alongside the Django model code.

Issue #304, opened 2022-07-14
Story: View Funding Status of a CAN


jonnalley:
### User Story
As an CAN lead, manager, or budget officer, in order to know what funds are available, what has been spent, and what needs to be allocated, I want to see a funding status for a CAN in FY 2022.

### Pre-conditions
- A schema is in place in the database for a CAN's spending plan and other funding attributes. 
- Existing CAN data and related data structures are in place

### Acceptance Criteria
- [x] User can view a CAN's funding status total fiscal year funding (assuming FY 2022).
- [x] User can view a CAN's funding status funded YTD amount.
- [x] User can view a CAN's funding status pending funds.
- [x] User can clearly distinguish any negative dollar amounts from positive dollar amounts. - @haprin

-----------------------------------------

### Tasks

- [x] Backend
  - [x] Implement OESA in backend Django project.
  - [x] Figure out unit tests (they seem finicky when run outside of Docker)
  - [x] Get code coverage checks implemented. @tdonaworth 
  - [x] Get an end-to-end test created. @tdonaworth 
  - [x] Load testing. - @halprin 
  - [x] Figure out what `ops_site/admin.py` is all about
  - [x] Add linting (tox, PEP8?)
  - [x] Add additional code to support viewing elements mentioned in the AC and wireframes
- [x] Frontend
  - [x] Move UI to React project.
  - [x] Allow frontend to call backend in deployed environment.
  - [x] Remove the weird extra appended values to the frontend domain in cloud.gov
  - [x] Add UI elements to support viewing elements mentioned in the AC and wireframes
  - [x] Unit tests. - @halprin 
  - [x] Get code coverage checks implemented. - @halprin 
  - [x] Add linting (ESLint, Prettier?)
- [x] DevOps
  - [x] Pivot to GitHub Actions for CI.
  - [x] Pivot to GitHub Actions for CD.
  - [x] Add and configure Renovatebot.  Disable Dependabot.
  - [x] Ensure Dependabot is doing dependency vulnerability scanning.
  - [x] Have CD process do some set of testing before deploy - @halprin 
- [x] Other
  - [x] Update and create new ADRs.
  - [x] Update READMEs. - @haprin

### Definition of Done
- [x] Usability validated
- [x] UI works as designed
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests passed
- [x] ~Automated integration tests passed~
- [x] Automated quality tests passed
[...]

PR #320, opened 2022-07-20
304: Slow Renovate's Python Updates

halprin:
## What changes

Renovate now will not automatically merge Python updates.  It requires more than just the one file that Renovate updates to update Python in this project.  The PR will still be created, but we can keep it from merging, improve it, and merge it manually ourselves when we're ready.

## Issue

#304


jonnalley (approved):
seems reasonable

PR #327, opened 2022-07-22
304: Move Frontend Away from Django

halprin:
## What changes

Moved the frontend into the React frontend.  It has been removed from Django.

## Issue

#304


tdonaworth (approved):
👍

PR #337, opened 2022-07-26
304: Frontend Communicates with Backend in Deployment

halprin:
## What changed

The deployment in an actual Cloud.gov wasn't working and wouldn't be able to call the backend.  The backend that the frontend calls is now configurable and therefore able to work in a deployed environment.

This resulted in a few other things changing...
- `DISABLE_ESLINT_PLUGIN=true` was added to disable Create React Scripts default and built-in eslint configuration.  `yarn lint` still works and runs.
- Created the `backend.js` gateway that allows us to call our backend from a single place.  It is configured through the environment variable `REACT_APP_BACKEND_DOMAIN`.
  - I hate that I need to append the `REACT_APP_...` in the actual .js files.  This PR is ready for review, but I wan to figure out how to change that.
- Removed some commands from the `manifest.yml` because they don't need to run anymore since we removed the static resource serving from Django.

## Issue

#304

## How to test

Play with the [dev deployed frontend](https://opre-ops-frontend-test-sweet-sable-ir.app.cloud.gov) and notice the CANs are populated.


jonnalley:
@halprin so with the new dynamic URL in place and the old no longer functional (as expected, although not sure why it defaults to random-route) we ought to update the README which makes a reference to the URL.

Also, might need to figure out how to make it not do that

jonnalley:
> @halprin so with the new dynamic URL in place and the old no longer functional (as expected, although not sure why it defaults to random-route) we ought to update the README which makes a reference to the URL.
> 
> Also, might need to figure out how to make it not do that

I submitted a new commit to remove the readme reference

jonnalley (approved):

halprin:
> > @halprin so with the new dynamic URL in place and the old no longer functional (as expected, although not sure why it defaults to random-route) we ought to update the README which makes a reference to the URL.
> > Also, might need to figure out how to make it not do that
> 
> I submitted a new commit to remove the readme reference

Thanks!

Yeah.  Let me add that as a task.

PR #340, opened 2022-07-28
304: refactored project, updated imports, basic pytest functionality working

tdonaworth:
## What changed

- Refactored project structure - changed child `opre_ops` directory to `django`
- Updated all imports to reflect the refactored changes
- Created a basic test file, to check for pytest functionality

## Issue

#304 

## How to test

`pipenv run pytest`

## Screenshots

## Links


.pre-commit-config.yaml
```diff
@@ -29,8 +29,14 @@ repos:
...
   - repo: local
     hooks:
-      - id: nox-flake8
-        name: nox-flake8
+      - id: nox-black
+        name: nox-black
+        entry: cd ./backend/ && pipenv run nox -s black
+        language: system
+        types: [python]
+        pass_filenames: false
```
halprin, line 37: Is there a reason you want Black to actually reformat your files during a pre-commit hook?  The `lint` session already checks to see whether your files are Black formatted, but it just doesn't format them.

The way you have it here it will reformat your files, but those changes would be unstaged.  And depending on timing, the `nox-lint` pre-commit hook could run before or after reformat and either find no issues (because the Black formatter already ran due to `nox-black` pre-commit hook) or still find the issues.

backend/Pipfile
```diff
@@ -11,6 +11,7 @@ djangorestframework = "==3.13.1"
 django-cors-headers = "==3.13.0"
 
 [dev-packages]
+pytest = "7.1.2"
```
halprin, line 14: Hah, that's hilarious that we didn't have this yet.

backend/.flake8
```diff
@@ -6,4 +6,4 @@ per-file-ignores = tests/*:S101
 max-complexity = 10
 max-line-length = 120
 import-order-style = google
-application-import-names = opre_ops,ops_site,tests
+application-import-names = django,ops_site,tests
```
halprin, line 9: Actually, I think the correct thing to do here is keep `opre_ops`, but remove `ops_site` and `django`.  We especially don't want `django` in this list because then any actual import of the `django` framework will be treated as an application-local import and be grouped with our application imports.

halprin (changes requested):
This is awesome!  I have a few recommendations, though.

backend/.flake8
tdonaworth: updated and resolved

halprin (approved):

.pre-commit-config.yaml
tdonaworth: removed nox-black

PR #341, opened 2022-07-28
304: Frontend ADR

halprin:
## What changed

Created a new frontend ADR that replaced the previous one.

## Issue

#304


tdonaworth (approved):
👍

PR #348, opened 2022-08-01
304: Frontend Unit Tests with Coverage

halprin:
## What changed

Added unit tests to the frontend and check for 90% code coverage.

This required making an application context to do dependency injection.

## Issue

#304

## How to test

`yarn test` or `yarn test --watchAll=false` if you don't want to run the tests just once.


tdonaworth (approved):
A good first slice 👍

PR #368, opened 2022-08-10
Issue 304 funding status

tdonaworth:
## What changed

- Refactored the Dockerfile(s) to improve caching of dependancies
- Moved `BudgetSummary` into its own Component
- Wired up `Funding Status` to the backend
- Created Tests for `Funding Status`
- Updated the `Fiscal Year` dropdown to be usable for `Funding Status`
- Generated new model diagrams (/docs/opre_ops_models.png)
- Created Controllers for the rest of the current Models

## Issue

#304


frontend/src/pages/cans/detail/getCfy.js
```diff
@@ -0,0 +1,18 @@
+import { setCfy } from "./canFiscalYearSlice";
+import ApplicationContext from "../../../applicationContext/ApplicationContext";
+
+export const getCfy = (id) => {
```
halprin, line 4: In general, I would expand `cfy` to its full name.  I'd do that here in the frontend and the backend.

frontend/package.json
```diff
@@ -9,6 +9,7 @@
     "react-dom": "18.2.0",
     "react-redux": "8.0.2",
     "react-router-dom": "6.3.0",
+    "react-select": "^5.4.0",
```
halprin, line 12: Let's pin this dependency.

halprin (changes requested):

backend/opre_ops/ops_site/cans/controller.py
```diff
@@ -19,3 +50,143 @@ class CanListController(ListAPIView):
...
+
+
+class PersonListController(ListAPIView):
+    queryset = Person.objects.all()
+    serializer_class = PersonSerializer
+
+
+class PersonReadController(RetrieveAPIView):
+    queryset = Person.objects.all()
+    serializer_class = PersonSerializer
```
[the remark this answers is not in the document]
tdonaworth: I'll remove the Person stuff, and agree, placement was wrong, and we'll ultimately be refactoring this to fit what we actually need. 
As for the other controllers I was trying to work ahead, but based on discussions I'll pair down to only what's needed for now.

docker-compose.yml
```diff
@@ -3,6 +3,9 @@ version: "3.8"
 services:
 
   db:
+    image: "postgres:14.4-alpine3.16"
+    platform: linux/arm64/v8
```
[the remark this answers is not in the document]
tdonaworth: Probably a typo on my part `amd64`

rest-client.http
```diff
@@ -0,0 +1,37 @@
+@hostname = localhost
```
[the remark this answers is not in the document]
tdonaworth: Intended this actually, but the need for the *.http files to exist at all in the repo leans towards moving this all together.

frontend/package.json
```diff
@@ -17,7 +18,7 @@
     "eslint-plugin-prettier": "4.2.1",
     "eslint-plugin-react": "7.30.1",
     "prettier": "2.7.1",
-    "react-scripts": "5.0.1"
+    "react-scripts": "^5.0.1"
```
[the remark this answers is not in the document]
tdonaworth: acknowledged.

frontend/package.json
tdonaworth: fixed

frontend/src/pages/cans/detail/getCfy.js
tdonaworth: Refactored

backend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.9
+FROM amd64/python:3.9-slim
```
halprin, line 1: Do we need the `amd64/` on this since we added `platform: linux/amd64` in the `docker-compose.yml`?

frontend/src/pages/cans/detail/BudgetSummary.jsx
```diff
@@ -48,18 +48,18 @@ const BudgetSummary = () => {
                                     <th>Amount</th>
                                 </tr>
                                 <tr>
-                                    <td>Total FY {cfy[0]?.fiscal_year || "--"} Funding</td>
-                                    <td>{cfy[0]?.total_fiscal_year_funding || "n/a"}</td>
+                                    <td>Total FY {can_fiscal_year[0]?.fiscal_year || "--"} Funding</td>
```
halprin, line 51: I noticed that there's always `[0]` being used.  Is it always an array of size 1?  Can we just return the first item without it being wrapped in an array?

frontend/src/pages/cans/detail/BudgetSummary.jsx
```diff
@@ -48,18 +48,18 @@ const BudgetSummary = () => {
...
+                                    <td>{can_fiscal_year[0]?.total_fiscal_year_funding || "n/a"}</td>
                                 </tr>
                                 <tr>
                                     <td>Funded YTD</td>
-                                    <td>{cfy[0]?.amount_available || "n/a"}</td>
+                                    <td>{can_fiscal_year[0]?.amount_available || "n/a"}</td>
```
halprin, line 56: I also noticed that `?` is always being used.  Why is that?  Is it because it's possible nothing could come back from the backend?

frontend/src/pages/cans/detail/BudgetSummary.jsx
```diff
@@ -1,12 +1,12 @@
...
-import { getCfyByCan } from "./getCfy";
+import { getCanFiscalYearByCan } from "./getCanFiscalYear";
 import { useEffect, useState } from "react";
 import Select from "react-select";
 import { useParams } from "react-router-dom";
 
 const BudgetSummary = () => {
     const dispatch = useDispatch();
-    const cfy = useSelector((state) => state.cfyDetail.cfy);
+    const can_fiscal_year = useSelector((state) => state.canFiscalYearDetail.canFiscalYear);
```
halprin, line 9: "Snake case" isn't really used in JavaScript.  I'd rename this to `canFiscalYear`.

backend/opre_ops/ops_site/cans/controller.py
```diff
@@ -19,3 +50,143 @@ class CanListController(ListAPIView):
 class CanReadController(RetrieveAPIView):
     queryset = CommonAccountingNumber.objects.all()
     serializer_class = CommonAccountingNumberSerializer
+
+
+class FundingPartnerSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = FundingPartner
+        fields = "__all__"
+        dept = 1
```
halprin, line 59: I think you were going to remove a bunch of these?

frontend/src/pages/cans/detail/BudgetSummary.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+const BudgetSummary = () => {
+    const dispatch = useDispatch();
+    const can_fiscal_year = useSelector((state) => state.canFiscalYearDetail.canFiscalYear);
+    const urlPathParams = useParams();
+    const [selectedValue, setSelectedValue] = useState(0);
+    const handleChange = (e) => {
+        setSelectedValue(e.value);
+        // console.log(cfy[0].fiscal_year);
+    };
```
halprin, line 15: We shouldn't need to use React's `useState` because we're using Redux for state management.

So, in `handleChange`, you would do...

```javascript
dispatch(setFiscalYear(e.value));
```

That would require a `setFiscalYear` reducer and its own state in `initialState`.  Lastly, you'd have to add a `useSelector` for the fiscal year, and update the `useEffect` dependency list.

frontend/src/pages/cans/detail/BudgetSummary.jsx
tdonaworth: Yeah, I went back and forth on whether to do the null handling here, or simply ensure the api endpoint would handle it and return an empty (fully structured) object.

frontend/src/pages/cans/detail/BudgetSummary.jsx
tdonaworth: fixed

backend/Dockerfile
tdonaworth: good catch

frontend/src/pages/cans/detail/BudgetSummary.jsx
tdonaworth: I was trying to find away around this, as the Controller is returning an `ListAPIView` so it's always a list. Can't use `RetrieveAPIView` or generic `APIView` (I don't think) as they require the lookup by (pk), unless we wanted to go in and overwrite everything. 

With that said, without knowing more about the full data structure, and whether there's a chance of multiple return values when looked up by `can_id` and `fiscal_year`, I'm not sure yet. The Model does indicate unique on those two values though, so it could be fairly constrained.

halprin (approved):

PR #374, opened 2022-08-17
304: Read Me

halprin:
## What changed

Updated the `README.md` based on all the changes we've made.  I also slimmed it down a little bit.

## Issue

#304.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,8 @@ This is the OPRE's Research Portfolio Management System, or OPS. The finished pr
 
 ## Dependencies
 
-At a bare minimum, you need [Docker](https://www.docker.com) installed to run the application locally.  If you want to do
+At a bare minimum, you need [Docker](https://www.docker.com) and
+[Docker Compose](https://docs.docker.com/compose/install/) installed to run the application locally.  If you want to do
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,177 +1,204 @@
...
-Technologies used in this prototype:
-* Python
-* Django 4.x
-* React
-* PostgreSQL
-* Cloud.gov
+## Dependencies
 
-## Getting Started
+At a bare minimum, you need [Docker](https://www.docker.com) installed to run the application locally.  If you want to do
```
tdonaworth, line 11: I don't recall, but does `Docker` include `Compose` now? I know it didn't in the past, and it was a separate install on some systems.

halprin: Oh, good point.  I know it does with Docker Desktop for ____.  Let me check.

halprin: I added a bit that we also require Docker Compose.  The webpage linked to mentions that you might already have it through Docker Desktop and also includes instructions on how to install it if you just have the engine/CLI installed.

tdonaworth: 👍

tdonaworth (approved):
Nice cleanup!

PR #379, opened 2022-08-17
304: Remove the outer describe blocks from the Cypress tests

halprin:
## What changed

Removed the outer describe blocks from the Cypress tests

## Issue

#304.

## How to test

`yarn test:e2e` still passes and runs all four tests.


tdonaworth (approved):
Ah, thanks, I totally forgot to do that earlier!

PR #384, opened 2022-08-19
Issue 304 automate security

tdonaworth:
## What changed

- Added Semgrep Scans to CI
- Cleaned up any prior existing Semgrep findings

## Issue

#304

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,19 +1,20 @@
 FROM python:3.9-slim
 RUN apt-get update && apt-get clean && rm -rf /var/lib/apt/lists/*
-WORKDIR /opt/local/
+RUN pip install --no-cache-dir --upgrade pip==22.2.2 pipenv==2022.7.24 && useradd -ms /bin/bash app
+USER app
+WORKDIR /home/app
 
-ENV PYTHONPATH=/opt/local/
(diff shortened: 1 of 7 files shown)


backend/Dockerfile
```diff
@@ -1,19 +1,20 @@
 FROM python:3.9-slim
 RUN apt-get update && apt-get clean && rm -rf /var/lib/apt/lists/*
-WORKDIR /opt/local/
+RUN pip install --no-cache-dir --upgrade pip==22.2.2 pipenv==2022.7.24 && useradd -ms /bin/bash app
+USER app
+WORKDIR /home/app
 
-ENV PYTHONPATH=/opt/local/
+ENV PYTHONPATH=/home/app
```
jonnalley, line 7: is this just a preference thing or was that a finding?

jonnalley (approved):
all the security! :-)

backend/Dockerfile
tdonaworth: preference, and mostly to ensure we're running in user-space. The overall fix was to add the `app` user. So running out of that user's home just ensures correct permissions, etc.

jonnalley: Yes indeed.  Splendid, thank you

PR #386, opened 2022-08-22
Remove duplicate WORKDIR in frontend Dockerfile

halprin:
## What changed

Remove the duplicate `WORKDIR` from the frontend `Dockerfile`.


tdonaworth (approved):
👍

PR #389, opened 2022-08-22
Specify /home/app in the docker-compose file

halprin:
## What changed

We recently moved to copy over our application code to `/home/app` in the `Dockerfile`s.  Now we specify that for the volumes in our `docker-compose.yml` file.


tdonaworth (approved):
Good catch. This does raise one issue, in that none of our workflows or checks use the containers now, so we have no full sense of "Dev" breakage 🤔

Issue #376, opened 2022-08-17
Story: View Portfolio Details Primer


jonnalley:
### User Story
As an OPRE staff member, I want to see a non-polished view of portfolios and limited details about a specific Portfolio (aka Research Portfolio) so that I can understand aspects of it in preparation for future iteration.

### Acceptance Criteria
- [x] Basic developer UIs exist for a list of portfolios and details thereof. 
- [x] Backend objects exist for the notion of portfolios in later development

-----------------------------------------

### UX Tasks
- [x] Share research on user needs with team - @andressacurvelo @kimschulke 
- [x] Iterate on design options based on user research in #388
- [x] Share design options for feedback with all stakeholders and end users - @andressacurvelo @kimschulke 
- [x]  Prioritize functionality to build now vs later

#### Frontend

- [x] Create a dumb list view anchored on an a URL path - @halprin 
- [x] Create simple link to the dumb list view at the homepage - @halprin
- [x] Create a dumb detail view
- [x] Only install non-dev dependencies when running in Docker - @halprin 
- [x] Fix the deployment not understanding direct URL paths - @halprin #401
- [x] Remove 18F CSS styling @halprin 

#### Backend

- [x] Migrate to use [Uvicorn](https://www.uvicorn.org) - @halprin 
- [x] Develop DB schema additions and changes required - @tdonaworth 
- [x] Make some fake data to go into the local DB
- [x] Create new controller/serializer
- [x] Hook up a URL to the controller
- [x] Conduct overhaul of GitHub Actions workflow - @tdonaworth 
- [x] Implement Zap for DAST - @tdonaworth  
- [x] Add additional code coverage and e2e testing improvements @halprin 
- [x] Add axe-core a11y testing @halprin 
- [x] Review data model and rename elements as needed @johndeange 

#### Other

- [x] ADR for a11y - @halprin 

### Definition of Done Checklist
- [x] Usability validated
- [x] UI works as designed
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests passed
- [x] Automated integration tests passed
- [x] Automated quality tests passed
- [x] Automated load tests passed
- [x] Automated a11y tests passed
- [x] Automated security tests passed
- [x] 90%+ Code coverage achieved
- [x] PR(s) have been merged to main
- [x] Design/tech debt eliminated
- [x] Build process updated
- [x] Documentation updated or added
- [x] Feature flags/toggles created

[...]

PR #394, opened 2022-08-22
376: Uvicorn & Gunicorn

halprin:
## What changed

We now use the [Uvicorn](https://www.uvicorn.org) as the application (ASGI) server.  We also use [Gunicorn](https://gunicorn.org) as the worker supervisor.

## Issue

#376


jonnalley:
whos-awesome

tdonaworth (approved):

PR #397, opened 2022-08-23
376: Only install non-dev dependencies when running in Docker

halprin:
## What changed

Only install non-dev dependencies when building our Docker container for the frontend.  This makes the frontend Docker image much smaller and also reduces the amount of time needed to install.

## Issue

#376.

## How to test

`docker compose up`.


.github/workflows/dev_deploy.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           curl -L "https://cli.run.pivotal.io/stable?release=linux64-binary" | tar xzv -C $HOME/bin
       - name: Install frontend dependencies
         working-directory: ./frontend
-        run: yarn install --frozen-lockfile
+        run: yarn install --production --frozen-lockfile
```
tdonaworth, line 36: Curious, if for the actual `deployment` if we'd rather simply set the `NDOE_ENV=production` env, to get all `Production` benefits or not. I'm fine limiting this to just the `install`. Just a thought.

halprin: We could set `NODE_ENV` to `production` to make `yarn install` do the same as `yarn install --production`, but `react-scripts build` (which is executed on line 48 by `yarn build`) doesn't care what `NODE_ENV` is set to because it automatically forces it to `production` anyway.  So, the only thing where setting `NODE_ENV` to `production` makes a difference is for `yarn install`, but eh... I guess I chose to do it through a CLI option.

Would you prefere I do it through the environment variable?

tdonaworth: No, it was more of a curiosity question, and curious if we'll want to set it at some point later anyway, especially for actual production. "multiple birds" type of thing. I'm okay as-is for this specific purpose.

tdonaworth (approved):
👍

.github/workflows/dev_deploy.yml
halprin: Yeah, might as well add it so if anything does trigger off of it, it just uses the prod version.

PR #403, opened 2022-08-25
remove web-vitals

tdonaworth:
## What changed

- Identified the inclusion of `web-vitals`; which is not yet being used. Cleaning up until the library is needed.

## Issue

#376


halprin (approved):
💣

PR #408, opened 2022-08-29
Fix Frontend Linting

halprin:
## What changed

Our frontend linting was not actually linting anything.  It wasn't traversing the directory structure.  Now it does.

Now that ESLint traverses everything, I need to tell it to ignore CSS and SVG files since it doesn't know how to lint those files.

I also fixed some of the findings.

## How to test

`yarn lint`

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -13,6 +13,10 @@
     },
     "plugins": [
     ],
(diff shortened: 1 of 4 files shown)


tdonaworth (approved):
👍

jonnalley (approved):

PR #409, opened 2022-08-29
Upgrade flake8 to 5 and flake8-bandit to 4.1.1

halprin:
## What changed

Combined PR #345 and #407 since they depend on each other.  Renovate didn't combine them because they were both major updates and we tell Renovate to keep major updates separate.


tdonaworth (approved):
🔥

PR #412, opened 2022-08-31
Remove the Read Only and Throttling Restrictions

halprin:
## What changed

Moved the restrictions that our ReST framework uses for anonymous users.  Specifically, the throttling was moved to only exist in a deployed environment.

This required changing how Semgrep scans and handles ignored findings.


tdonaworth:
I'm not opposed, I upped the limit in my current code edits for the same reasons (100 was way too low). But I'd like to keep not, or some other "before production" reminders to ensure we review and set these to some reasonable value before going to "Production". That's mostly why they ended up in here so early. Do we need them now, no, but we will want them at some point.

johndeange (approved):

halprin:
> I'm not opposed, I upped the limit in my current code edits for the same reasons (100 was way too low). But I'd like to keep note, or some other "before production" reminders to ensure we review and set these to some reasonable value before going to "Production". That's mostly why they ended up in here so early. Do we need them now, no, but we will want them at some point.

@tdonaworth, If you think we will actually use this later on, I'm fine with keeping them and upping the number.  What did you set yours to or what do you think is a better number?

tdonaworth:
> 

Mine currently is set to 2500 each; we'll need to play with these numbers once we have real data obviously. 

Alternatively, it might simply be better to overwrite these for `settings.local` directly, as that's the main issue right now correct? Or move them to their respective `settings.cloudgov` and `settings.local` - thoughts?

tdonaworth (approved):
I have tickets open with both Semgrep and GitHub CodeQL about the bug we found, so will allow the work-around for now, begrudgingly 😄

PR #413, opened 2022-09-01
376: Adopt USWDS on Main Page

halprin:
## What changed

I tried to keep the styling of the other pages in tact while changing others.  I can't seem to do that.  So, I removed the reference to the 18F CSS file and let the other pages look weird for now.  They will be fixed in subsequent PRs.  At that point, I will remove the `index.css` 18F file.

- Replaced the navigation table with the USWDS styled navigation table.
- Removed the family image.
- Moved the `BudgetSummary` React component to use CSS modules so its red negative styling doesn't possibly conflict with USWDS's global styling.
- Removed the old `gulpfile.js`.

## Issue

#376.

## How to test

Navigate to http://localhost:3000 when running locally.

## Screenshots

Main page using USWDS styling


frontend/src/App.jsx
```diff
@@ -19,44 +18,33 @@ function App() {
...
-                                    </td>
-                                </tr>
-                            </tbody>
-                        </table>
-                    </div>
-                </div>
-            </div>
-        </main>
+
+            <h1>This is the OPRE OPS system prototype.</h1>
```
jonnalley, line 22: I think the word system is redundant as that's what the last S in the acronym stands for. 

Frankly, OPRE is as well as @kimschulke has mentioned but I think it's also clarifying to some extent---that one feels more more grey.

tdonaworth (approved):
It's a good start

frontend/src/App.jsx
halprin: Hmmm.  I'll change it in the next PR.

PR #414, opened 2022-09-01
376: Added JWT support to the API

tdonaworth:
## What changed

- added the basics for JWT support to the backend api
- added a static user to the `fixture`
- re-added settings for `staticfiles` support within the `/admin`

## Issue

#376 

## How to test

JWT Testing: 
POST call to `/api/token` with a payload for a username/password

```curl \
  -X POST \
  -H "Content-Type: application/json" \
  -d '{"username": "admin", "password": "admin"}' \
  http://localhost:8000/api/token/
```


halprin (commented):
~Do we have an endpoint that needs to be protected by JWTs yet?~

Eh... ignore this review.

backend/Pipfile
```diff
@@ -12,6 +12,7 @@ django-cors-headers = "==3.13.0"
 django-filter = "==22.1"
 uvicorn = {version = "==0.18.3", extras = ["standard"]}
 gunicorn = "==20.1.0"
+djangorestframework-simplejwt = "*"
```
halprin, line 15: Can you pin this dependency to the specific version?

halprin (commented):
Does this change automatically lock down all of the endpoints?

tdonaworth:
> 

No, this doesn't have any enforcement on using the JWT yet, simply supports generating a basic JWT, as a way to authenticate to the backend. Step 1 of many.

backend/Pipfile
tdonaworth: > 

good catch. Pinning to 5.2.0

halprin (approved):

PR #423, opened 2022-09-05
Refactor breadcrumb component

johndeange:
## What changed

Added BreadcrumbList and BreadcrumbItem React Components.  Added aria-current and aria-label.

## Issue

N/A

## How to test

https://opre-ops-frontend-test.app.cloud.gov/


tdonaworth (approved):
👍

frontend/src/pages/UI/Breadcrumb.jsx
```diff
@@ -0,0 +1,32 @@
+import { Link } from "react-router-dom";
+
+const BreadcrumbList = (props) => {
```
halprin, line 3: Mind if we split `props`?  I.e. something like this...
```suggestion
const BreadcrumbList = ({isCurrent, children}) => {
```

frontend/src/pages/UI/Breadcrumb.jsx
```diff
@@ -0,0 +1,32 @@
...
+                        Home
+                    </Link>
+                </li>
+                {props.children}
+            </ol>
+        </nav>
+    );
+};
+
+const BreadcrumbItem = (props) => {
```
halprin, line 18: Same here.  Can we split `props` here too?

frontend/src/pages/UI/Breadcrumb.jsx
```diff
@@ -0,0 +1,32 @@
+import { Link } from "react-router-dom";
```
halprin, line 1: Maybe move the `UI` directory outside `pages`?  Maybe just call it `components`?

halprin (commented):
Nice changes, @johndeange.  I've got some suggestions.  What do you think?

frontend/src/pages/UI/Breadcrumb.jsx
johndeange: Using `props` I think is the idiomatic way of passing it in: https://reactjs.org/docs/components-and-props.html
Our existing code uses `props` this way as well.

frontend/src/pages/UI/Breadcrumb.jsx
johndeange: Same reply as above.

frontend/src/pages/UI/Breadcrumb.jsx
johndeange: That would be my preference too - I will move it.

frontend/src/pages/UI/Breadcrumb.jsx
halprin: Mmmm.  I'm going to push back just a bit more.  If you really think `props` is the way to go, I'll approve.  But...  I see a number blob posts mention destructuring the arguments is a good practice, we did this to great effect at my previous client (US Tax Court), and I think this helps with us understanding specifically what is needed for a component to operate with a quick glance.

johndeange: I found a couple examples of it being done this way online as well - I'll apply it since I don't have a strong preference.

frontend/src/pages/portfolios/detail/PortfolioDetail.jsx
```diff
@@ -3,10 +3,10 @@ import { getPortfolio } from "./getPortfolio";
 import { useEffect } from "react";
 import { useParams, Link } from "react-router-dom";
 
-const CanList = (props) => {
+const CanList = ({ id, name }) => {
     return (
         <li>
-            <Link to={`/cans/${props.id}`}>{props.name}</Link>
+            <Link to={`/cans/${id}`}>{name}</Link>
```
halprin, line 9: Thanks for this!

halprin (approved):

PR #424, opened 2022-09-06
adding code-ql scans; added `development` to ci

tdonaworth:
## What changed

- Added CodeQL to Security Scans Action
- Added `development` branch to CI workflow for `pull`s

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,8 @@ jobs:
 
   codeql-analysis:
     permissions:
+      actions: read
+      contents: read
       security-events: write
     name: CodeQL Scan
     uses: ./.github/workflows/security_codeql.yml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,13 @@ jobs:
     name: Unit Tests
     uses: ./.github/workflows/unit_test_reusable.yml
 
-  semgrep:
+  CodeQL Analysis:
+    permissions:
+      security-events: write
+    name: CodeQL Scan
+    uses: ./.github/workflows/security_codeql.yml
+
+  Semgrep Analysis:
     permissions:
       security-events: write
     name: Semgrep Scan
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,8 +5,7 @@ on:
     branches:
       - main
   push:
-    branches:
-      - main
+    branches: [main, development]
 
 jobs:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,13 +13,13 @@ jobs:
     name: Unit Tests
     uses: ./.github/workflows/unit_test_reusable.yml
 
-  CodeQL Analysis:
+  codeql-analysis:
     permissions:
       security-events: write
     name: CodeQL Scan
     uses: ./.github/workflows/security_codeql.yml
 
-  Semgrep Analysis:
+  semgrep-analysis:
     permissions:
       security-events: write
     name: Semgrep Scan
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
       - main
   push:
     branches: [main, development]
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
 
 jobs:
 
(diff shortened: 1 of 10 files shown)


johndeange (approved):

backend/opre_ops/django_config/settings/local.py
```diff
@@ -5,9 +5,17 @@
...
 from opre_ops.django_config.settings.helpers.random_string import generate_random_string
 
+__all__ = [
+    "DATABASES",
+    "SECRET_KEY",
+    "DEBUG",
+    "ALLOWED_HOSTS",
+    "CORS_ALLOWED_ORIGIN_REGEXES",
+    "INSTALLED_APPS",
+]
```
halprin, line 18: Did you mean to keep this in?  We didn't do the same in `common.py`.

.github/workflows/security_codeql.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
```
halprin, line 29: Instead of making this action run twice (once for JavaScript and once for Python), you can run it once with both languages.

`languages: python, javascript`

See the code snippet right above [this anchor](https://github.com/github/codeql-action#configuration-file).

.github/workflows/ci.yml
```diff
@@ -5,16 +5,26 @@ on:
     branches:
       - main
   push:
-    branches:
-      - main
+    branches: [main, development]
```
halprin, line 8: @tdonaworth, I'm late for the review, but is there a reason that we are in including pushes to `development`?

The `dev_deploy.yml` already run the backend/frontend unit tests and end-to-end tests before allowing the deploy.

halprin (commented):
I'm late.  😦

.github/workflows/ci.yml
tdonaworth: unit-tests and e2e tests are run, but no additional security checks are run; This specifically occurs for dependency related security findings that may pop up in the time between when `main` was pushed, and when `deploy` is performed; there could be a few days, when the security posture has changed, etc. 

Under this same line of thinking; I plan on creating a nightly job which runs those type of security checks.

backend/opre_ops/django_config/settings/local.py
tdonaworth: Yes you can, and I thought about that, but actually liked the idea of having it separate; since we effectively break our code/language by area (python:backend, js:frontend). The time required is the same with minimum overhead on each; so might save 10 seconds on the build/cleanup. I'll play with a few other cleanup steps to and see where it goes; There are a few other break-outs I'd prefer to do across the pipeline as well.

.github/workflows/ci.yml
halprin: Ah, OK.  Should we then actually put these two security checks as part of a requirement before deployment then?  E.g. add it in the `dev_deploy.yml` file.  That way it will actually stop the deployment if something is found?  Right now if I push to `development`, the deploy will keep happening and if a security finding is found, it isn't reallly reported anywhere except perhaps somehwere in the Security tab.

Albeit, there's a discussion to be had that this is the development environment and used for quick and dirty testing, and therefore one may not care to have development blocked by a finding.

.github/workflows/security_codeql.yml
tdonaworth: Yes you can, and I thought about that, but actually liked the idea of having it separate; since we effectively break our code/language by area (python:backend, js:frontend). The time required is the same with minimum overhead on each; so might save 10 seconds on the build/cleanup. I'll play with a few other cleanup steps to and see where it goes; There are a few other break-outs I'd prefer to do across the pipeline as well.

halprin: Ah word. Sounds good.

.github/workflows/ci.yml
tdonaworth: Much larger discussion; but yes it should happen and and be required there. Won't lie I haven't looked too closely at our `dev_deploy` yet; especially from a `path to production` standpoint; so it will change.

PR #425, opened 2022-09-06
issue-minor-cleanup

tdonaworth:
## What changed

- cleanup from late reviews by @halprin ;) 
- tested, and looks like combining `language` on `codeql` saves ~30 seconds; mostly from standup/teardown of the job. I still like the separation; but will let it run-as is (combined) for now.
- cleaned up `local.py` - my goof


halprin (approved):
😭  I'm on time now!  😉

johndeange (approved):

PR #449, opened 2022-09-13
376: A11y Testing ADR

halprin:
## What changed

An ADR for a11y testing.

## Issue

#376.

### DIFF
--- a/docs/adr/015-a11y-testing.md
+++ b/docs/adr/015-a11y-testing.md
@@ -1,0 +1,28 @@
+# 15. Accessibility (a11y) Testing
+
+Date: 2022-09-13
+
+## Decision
+
+A11y is important.  To ensure OPS is accessible, we will use [aXe](https://github.com/dequelabs/axe-core) in
+[combination](https://www.npmjs.com/package/cypress-axe) with [Cypress](https://www.cypress.io) to do a11y testing.
+We will strive to a11y test every view of OPS.  We will accomplish this by inlining a11y checks in our existing
+end-to-end tests.
+
+## Status
+
+Accepted.
+
+## Context
+
+A11y is important.  Not only that, it is mandated by federal law.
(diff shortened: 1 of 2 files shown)


docs/adr/015-a11y-testing.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Status
+
+Accepted.
+
+## Context
+
+A11y is important.  Not only that, it is mandated by federal law.
+
+[aXe](https://www.deque.com/axe/) is a major player in the a11y testing and training.  They make an open-source tool called axe-core that we'll
```
tdonaworth, line 20: Do we have any specific reason for choosing `axe` over something else? I feel like thats one of the main purposes of these records as well is to enforce why the decision was made over something else.

halprin: That's a good point.  I _heard_ that aXe is better than pa11y from Brice.  Let me dig that up, do some more research, and expand on the ADR.

docs/adr/015-a11y-testing.md
```diff
@@ -0,0 +1,38 @@
...
+
+A11y is important.  Not only that, it is mandated by federal law.
+
+[aXe](https://www.deque.com/axe/) is a major player in the a11y testing and training.  They make an open-source tool
+called axe-core that we'll be utilizing through Cypress.  That makes it easy to a11y test because Cypress is...
+
+- Adept at interacting and navigating websites to get them into all the different states that need to be a11y tested.
+- Already used for our [end-to-end testing](./014-use-cypress-for-testing.md).
+
+We' are using aXe because it...
```
tdonaworth, line 26: "We're"?

tdonaworth (approved):
fix typo and I'm good

johndeange (approved):

Issue #499, opened 2022-10-24
DevEx: Flask Migration


tdonaworth:
### Goals
* Reduce backend bloat and footprint

### Tasks
- [x] Get docker container issues resolved @tdonaworth 
- [x] Migrate Auth endpoints @tdonaworth 
- [x] Migrate Portfolio Model @xlorepdarkhelm 
- [x] Migrate Portfolio dependencies @xlorepdarkhelm 
- [x] Handle Tests @xlorepdarkhelm 
- [x] Restructure as needed. @xlorepdarkhelm 
- [x] Author new ADR and update & deprecate existing ADR 5 for the move from Django to Flask @xlorepdarkhelm 
- [x] Update repo README @xlorepdarkhelm 
- [x] Repopulate DB
- [x] PR issued and merged @xlorepdarkhelm 
- [x] Cutover frontend
- [x] Remove Django remnants

PR #546, opened 2022-11-16
499 - Change Backend to use Flask

tdonaworth:
## What changed

* Switch to using a Flask backend
* Switch to using SQLAlchemy (ORM)
* Change frontend to call new backend

## Issue

#499 

## How to test

`docker compose up` --> App should work as before.

## Notes
* Additional refactoring will be needed, as well as removing all `django` structure. 
* CI/CD Needs updating to adhere to new structure or refactored to be agnostic of project structure.


backend/ops_api/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM python:3.9-slim
```
johndeange, line 1: Do we want to rev to 3.10 here since setup-python is using it?

johndeange (approved):

backend/ops_api/Dockerfile
tdonaworth: good call 👍

PR #551, opened 2022-11-18
Ops 461 portfolio description

johndeange:
## What changed

Add Portfolio Name, Division, Description, and URL to the Portfolio Details page.

## Issue

https://app.zenhub.com/workspaces/opre-ops-62d197b4c468970013bcd2ec/issues/hhs/opre-ops/461

## How to test

Currently can only test locally at: http://localhost:3000/portfolios/6

## Screenshots

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,6 +55,7 @@ services:
     container_name: ops-backend
     ports:
       - "8080:8080"
+    command: ["python", "-m", "flask", "run", "--host=0.0.0.0", "--port=8080"]
     environment:
       - JWT_PRIVATE_KEY
       - OPS_CONFIG=environment/local.py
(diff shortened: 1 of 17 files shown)


docker-compose.yml
```diff
@@ -55,6 +55,7 @@ services:
     container_name: ops-backend
     ports:
       - "8080:8080"
+    command: ["python", "-m", "flask", "run", "--host=0.0.0.0", "--port=8080"]
```
tdonaworth, line 58: I know this is only used in dev. Do we prefer to simply use the flask engine in dev? We should keep some parity with live though, during testing, etc

tdonaworth (changes requested):

docker-compose.yml
johndeange: yes the flask dev web server provides better error message and context logging than the production web server - so this helps for development work

tdonaworth (approved):

PR #584, opened 2022-12-05
Demo-data-update

tdonaworth:
## What changed

* Updated data records to be even round numbers
* small ui tweak to create a dummy card

## Images

image


frontend/src/components/RoundedBox/RoundedBox.jsx
```diff
@@ -0,0 +1,41 @@
+import { library } from "@fortawesome/fontawesome-svg-core";
+import { faSquare } from "@fortawesome/free-solid-svg-icons";
+import cssClasses from "./styles.module.css";
+
+library.add(faSquare);
+
+const styles = {
```
johndeange, line 7: Should be able to delete this variable

tdonaworth: good catch, missed that one.

PR #648, opened 2023-01-27
Data import etl experiment

johndeange:
Refactored models to common component (separated dependencies)
data tools now using ORM models rather than DB reflection.
Using Postgres DB for unit tests in data tools instead of sqlite.


backend/data_tools/environment/cloudgov.py
```diff
@@ -0,0 +1,30 @@
...
+        # Note: if or when we have more than one db per application instance,
+        # we will need to do something smarter than taking the first db in the list
+        database_service = vcap_services["aws-rds"][0]
+        database_creds = database_service["credentials"]
+
+        return f'postgresql+psycopg2://{database_creds["username"]}:{database_creds["password"]}@{database_creds["host"]}:{database_creds["port"]}/{database_creds["db_name"]}'
+
+    @property
+    def opre_excel_connection_string(self):
+        return "excel:///?Excel File='/Users/jdeangelis/PycharmProjects/OPRE-OPS-2/data-tools/data/REDACTED - FY22 Budget Summary -10-12-22.xlsm'"
```
tdonaworth, line 26: Do we want to check in a user specific directory here?

backend/data_tools/environment/dev.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class DevConfig(DataToolsConfig):
+    @property
+    def db_connection_string(self):
+        return "postgresql://postgres:local_password@localhost:5432/postgres"  # pragma: allowlist secret
+
+    @property
+    def opre_excel_connection_string(self):
+        return "excel:///?Excel File='/Users/jdeangelis/PycharmProjects/OPRE-OPS-2/data-tools/data/REDACTED - FY22 Budget Summary -10-12-22.xlsm'"
```
tdonaworth, line 11: Save here - default to a user directory?

backend/data_tools/tests/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    container_name: unit-test-db
+    security_opt:
+      - no-new-privileges:true  # Resolve semgrep https://sg.run/0n8q
+    environment:
+      - POSTGRES_PASSWORD=local_password  # pragma: allowlist secret
+    read_only: true  # Resolve semgrep https://sg.run/e4JE
+    tmpfs: /var/run/postgresql/
+    ports:
+      - "54321:5432"
+    healthcheck:
```
tdonaworth, line 17: We technically don't need this here, if there's no other services in the stack that needs to wait on this healthcheck - especially since the pytest has a built-in "wait" function already. It doesn't hurt to keep it, just noting for reference.

tdonaworth (changes requested):

backend/data_tools/environment/cloudgov.py
johndeange: Checked in accidentally - backed it out.

backend/data_tools/environment/dev.py
johndeange: Checked in accidentally - backed it out.

backend/data_tools/tests/docker-compose.yml
johndeange: Removed.

tdonaworth (approved):

Issue #656, opened 2023-02-07
Decide on & onboard with authn provider for OPS


jonnalley:
### Goals
* Have a decision reached on which authentication service provider we'll use as the go-forward solution for OPS and get onboarded to it

### Tasks
- [x] Gather info from ACF OCIO on if they have a direction or mandate on login.gov, HHS AMS, or others (OPRE Tech Lead is coordinating)
- [x] If applicable, gather HHS AMS cost and onboarding info including registration, etc...
- [x] #970
~- [ ] #1077~
~- [ ] #1078~
~- [ ] If necessary, discuss any configuration choices and factors with OPRE stakeholders and reach consensus~

### Additional Context
*

AmareB441:
As per the suggestion OPRE Data Security Manager, an email has been sent to the ACF OCIO Deputy Cyber Security Manager for guidance on Login.gov/HHS AMS services including the request for requirement information or documentations that we should follow in the design and development process.

jonnalley:
Following up on Amare's previous comment, it would appear the direction for the time being is to use the HHS-provided AMS. More direction and interaction with OCIO to come on this.

jonnalley:
AMS meeting scheduled for next week, yay :-)

jonnalley:
met with reps from ACF/OCIO to get a client added for us on their end. Right now it just points to localhost for local dev while all the various AMS options and configuration is tweaked optimally as well as changes being made to accommodate on the OPS end. Following all that, promotion on upward to other environments will be planned and implemented

jonnalley:
Trying to schedule another working session for next week

jonnalley:
Testing with AMS is complete and has been validated in `localhost` and `staging`. Targeting a final cutover on 7/10

jonnalley:
waiting on cutover until further OCIO action around non-PIV accountholders.

jonnalley:
Related issue is #1206 about supporting both AMS and login.gov. We're already ready to cutover/add support for any environments so this may be kinda sorta moot.

jonnalley:
Moving ADR task to #1206 and will create a new backlog item to get onboarded to the production environment

Issue #448, opened 2022-09-12
Investigate & Experiment with Permission Frameworks


tdonaworth:
### Goals
* Have enough knowledge to make a decision on **what "permissions" model / package to use** based on our functional and security needs.
* Generate an ADR about the final decision.

### Tasks
- ~~[ ] Research [Django REST - Access Policy](https://github.com/rsinger86/drf-access-policy)~~
- ~~[ ] Research [Composed Permissions](https://www.django-rest-framework.org/api-guide/permissions/#composed-permissions)~~
- ~~[ ] Research [DRY Rest Permissions](https://www.django-rest-framework.org/api-guide/permissions/#dry-rest-permissions)~~
- ~~[ ] Research [Django REST Framework API Key](https://www.django-rest-framework.org/api-guide/permissions/#django-rest-framework-api-key)~~
- ~~[ ] Research [Django Rest Framework Roles](https://www.django-rest-framework.org/api-guide/permissions/#django-rest-framework-roles)~~
- [x] Research [Open Policy Agent (OPA)](https://www.openpolicyagent.org/), along with Flask Extension [flask-opa](https://github.com/EliuX/flask-opa)
- [x] Research [OSO](https://docs.osohq.com/index.html)
- [x] Research [SqlAlchemy-Auth](https://github.com/dhiltonp/sqlalchemy_auth)
- [x] Research [flask-authorize](https://flask-authorize.readthedocs.io/en/latest/)
- [x] Author the ADR
- [x] Clean up some `path to production` items like enforcing login, checking for JWT validation for each endpoint, manual list of allowed login.gov users, etc..

### Additional Context

The idea is to spin up a few branches experimenting with each different permission framework, and determine which we "feel" is better.

Updated to remove Django specific options, after the switch to Flask. Added new Flask options to look into.

jonnalley:
Team intends today to mob on creating initial structure of the backend endpoint that would do the second call. See [Lucidchart diagram](https://lucid.app/lucidchart/d36f95e3-7e29-4182-95ea-f7829d777470/edit?viewport_loc=-166%2C18%2C1611%2C764%2C0_0&invitationId=inv_e89efce7-3a69-4d26-ae41-1301d8213e3c#)

jonnalley:
@tdonaworth what's left here?  Are these tasks up to date in terms of what needs to be done in particular?

jonnalley:
[...]

Issue #443, opened 2022-09-12
Story: Enforce Initial Authentication


jonnalley:
### User Story
As an OPS System Admin, I want to require authentication so that I can limit who is able to initially access OPS and initially prevent undesired user access.

### Acceptance Criteria
- [x] Users must not be able to access any OPS page via direct URL or make any other HTTPS request (GET, POST, etc) without authenticating first
- [x] Have a provision for session state to expire after the prescribed time (15 minutes)

-----------------------------------------

#### Dev: 
- [x] Procure login banner text from OPRE tech lead [or potentially steal from MAPS](https://map.acf.hhs.gov/map/app/)
- [x] Create any persistent store schema related to the concept of a user account logging in
- [x] Create backend code related to the log in action
- [x] Create backend code related to refreshing of a user's session ?
- [x] Create backend code related to expiry of a user's session
- [x] #711 - Frontend code and experience for expired log in session
- [x] Create frontend code/UX related to the log in interaction

- [x] #715
- [x] Write more awesome unit tests, related to #634
- [x] Fix e2e tests which now fail because they don't authenticate

### Definition of Done Checklist
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests passed
- [x] Automated integration tests passed
- [x] Automated quality tests passed
- [x] Automated load tests passed
- [x] Automated a11y tests passed
- [x] Automated security tests passed
- [x] 90%+ Code coverage achieved
- [x] PR(s) have been merged to main
- [x] Design/tech debt eliminated
- [x] Build process updated
- [x] Documentation updated or added
- [x] Feature flags/toggles created

### Additional Context & Resources
* There are ACF-specified login banners and messages that need to be used. 
* There are also USWDS styles/components to use for those messages 
* Inactivity/refresh tokens will be handled later

tdonaworth:
Current `Warning` used: 
![image](https://user-images.githubusercontent.com/56687505/221642865-786f7e51-1a86-4667-9eb2-bc94092680eb.png)

jonnalley:
opre review on 3/6

PR #658, opened 2023-02-08
Ops 448 jwt auth z

tdonaworth:
## What changed

- Flushed out the creation of Access and Refresh Tokens from the `backend`
- Updated the `frontend` to support storing the Access token, and provide them with each request to the `backend`
- Updated `BaseView` to require valid tokens `jwt_required()`, which propagates up to all endpoints
- Added unit tests to test for valid tokens on endpoint requests, and ensure they return 401 when not present
- Created a `auth_client` for use in pytest; which facilitates providing a pre-authenticated client (Authorization header populated)
- Wired up the unit tests to hit the PostgreSQL DB instead of the fallback SQLite
- Updated all unit tests / e2e tests to reflect the new data from our dev data
- Created a basic (developer) `frontend` page for `UserDetails` - mostly for testing
- Updated the Login button header, to include a link for the logged in user.

## Issue

#443 
#448
#656 

## How to test

`PyTest`
Manual - Login should now work, and accessing an `Portfolios` or `CANs` shouldn't return results unless you're logged in.

## Screenshots

image


backend/models/users.py
```diff
@@ -1,14 +1,16 @@
...
 from sqlalchemy import Column, DateTime, ForeignKey, Integer, String, func
+from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import column_property, relationship
 
 
 class User(BaseModel):
     """Main User model."""
-
     __tablename__ = "users"
-    id = Column(Integer, primary_key=True)
```
johndeange, line 11: I think it would be OK to do it this way but having id as an Integer and the only PK on the table may be preferable since it is a little simpler.

backend/ops_api/1
```diff
@@ -0,0 +1,8 @@
+: 1675801665:0;pytest
```
johndeange, line 1: not sure what this file is

backend/ops_api/Pipfile
```diff
@@ -27,6 +29,7 @@ pytest-flask = "==1.2.0"
 black = {extras = ["d"], version = "==22.12.0"}
 flake8-black = "==0.3.6"
 pytest-cov = "*"
```
johndeange, line 31: maybe fix versions for * in file here - also you probably don't need to explicitly include "six" here

backend/ops_api/ops/utils/auth.py
```diff
@@ -23,20 +23,26 @@ def user_lookup_callback(_jwt_header: dict, jwt_data: dict) -> Optional[User]:
...
-    payload = {
-        "iss": client_id,
-        "sub": client_id,
+    # client_id = current_app.config["AUTHLIB_OAUTH_CLIENTS"]["logingov"]["client_id"]
+    _payload = payload or {
+        "iss": current_app.config["AUTHLIB_OAUTH_CLIENTS"]["logingov"]["client_id"],
+        "sub": current_app.config["AUTHLIB_OAUTH_CLIENTS"]["logingov"]["client_id"],
         "aud": "https://idp.int.identitysandbox.gov/api/openid_connect/token",
         "jti": str(uuid.uuid4()),
         "exp": int(time.time()) + 300,
```
johndeange, line 44: wasn't this defined as an environment variable?

backend/ops_api/ops/utils/auth_views.py
```diff
@@ -1,43 +1,44 @@
...
 
 import requests
-from flask import Response, jsonify, request
+from flask import Response, current_app, jsonify, request
 from flask_jwt_extended import create_access_token, create_refresh_token, get_jwt_identity, jwt_required
-from ops_api.ops.utils.auth import get_jwt, oauth
+from ops_api.ops.utils.auth import create_oauth_jwt, oauth
 from ops_api.ops.utils.user import process_user
 
+logging.basicConfig(stream=sys.stderr, level=logging.DEBUG)
```
johndeange, line 12: instead of independently configuring logging here it would be better to get the logger from the current_app

backend/models/users.py
tdonaworth: Yeah, I've thought about this, and I think my main reason for swapping it to use the UUID as the PK, was because that's how we'd plan on looking up the User most times anyway. With the BaseView, etc - we'd have to create additional endpoints or query string filters to search by `oidc_id` specifically. So just thought it easier to swap it out. 

But I do see where it could create some conflict.

backend/ops_api/ops/utils/auth.py
tdonaworth: we have:

- JWT_ACCESS_TOKEN_EXPIRES
- JWT_REFRESH_TOKEN_EXPIRES
which are both used for our API Token. This is for the `authentication_code` Token sent back to login_gov. We probably should make this a config, or align it with our own `JWT_ACCESS_TOKEN_EXPIRES` - since it should be short-lived anyway.

backend/ops_api/1
tdonaworth: cleaned up

backend/models/users.py
tdonaworth: I will update this in the next PR

backend/ops_api/Pipfile
```diff
@@ -15,6 +15,7 @@ flask-cors = "==3.0.10"
 psycopg2-binary = "==2.9.5"
 gunicorn = "==20.1.0"
 sqlalchemy-mixins = "==1.5.3"
+black = "==23.1.0"
```
johndeange, line 18: already a dev dep

backend/ops_api/ops/environment/default_settings.py
```diff
@@ -1,17 +1,27 @@
...
+JWT_REFRESH_TOKEN_EXPIRES = timedelta(hours=12)
+
 AUTHLIB_OAUTH_CLIENTS = {
     "logingov": {
         "server_metadata_url": "https://idp.int.identitysandbox.gov/.well-known/openid-configuration",
+        "user_info_url": "https://idp.int.identitysandbox.gov/api/openid_connect/userinfo",
         "client_id": "urn:gov:gsa:openidconnect.profiles:sp:sso:hhs_acf:opre_ops",
-        "client_kwargs": {"scope": "openid"},
-    }
+        "client_kwargs": {"scope": "openid email"},
```
johndeange, line 25: I think we will want the first and last name from `profile`

backend/ops_api/ops/utils/auth_views.py
```diff
@@ -48,17 +45,22 @@ def login() -> Union[Response, tuple[str, int]]:
...
-        return jsonify(access_token=access_token, refresh_token=refresh_token)
+        response = jsonify(access_token=access_token, refresh_token=refresh_token)
+        response.headers.add("Access-Control-Allow-Origin", "*")
+        return response
 
     except Exception as err:
+        logging.error(err)
         traceback.print_exc()
-        return f"You screwed up!: {err}", 400
+        return f"Login Error: {err}", 400
```
johndeange, line 63: do we need to add the response header, i.e. `Access-Control-Allow-Origin` here as well?

frontend/src/pages/portfolios/detail/PortfolioDetail.jsx
```diff
@@ -77,7 +77,7 @@ const PortfolioDetail = () => {
     return (
         <>
             <App>
-                <Breadcrumb currentName={portfolio.name} />
+                <Breadcrumb currentName={"Tim-is-awesome"} />
```
johndeange: haha

johndeange (approved):

backend/ops_api/ops/environment/default_settings.py
tdonaworth: We ultimately will, and that locks us into using AAL2 as well, maybe flush that out more when we fully lockdown our "User" needs.

frontend/src/pages/portfolios/detail/PortfolioDetail.jsx
tdonaworth: Already updated that 😉

backend/ops_api/ops/utils/auth_views.py
tdonaworth: Not sure it was needed, but updating to make it consistent - we need to figure out why the global CORS isn't working - and eventually set correct origins

backend/data_tools/environment/pytest.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class PytestConfig(DataToolsConfig):
+    @property
+    def db_connection_string(self):
+        return "postgresql://postgres:local_password@unittest_db:5432/postgres"  # pragma: allowlist secret
+
+    @property
+    def opre_excel_connection_string(self):
+        return "excel:///?Excel File='/Users/jdeangelis/PycharmProjects/OPRE-OPS-2/data-tools/data/REDACTED - FY22 Budget Summary -10-12-22.xlsm'"
```
jonnalley, line 11: @tdonaworth @johndeange was this intentional to have this path specific to this local environment in here ?

Issue #594, opened 2022-12-12
Carry-forward Funds Update


jonnalley:
### User Story
As an OPRE staff member, when looking at a portfolio and its CANs, I want to better understand how much of the funding is from the currently-selected fiscal year & how much is from prior fiscal years so that I can make better decisions about the timing of fund usage.

### Acceptance Criteria
- [ ] When viewing a portfolio's details, the user can quickly discern if a CAN has carry-forward funds.
- [ ] When viewing a portfolio's summary, the user can quickly discern and see the amount of funds new to the FY and carried forward from a fiscal prior to the selected one.
- [ ] When viewing a portfolio's details (CANs), the user can quickly get an idea for the ratio of the CAN's received funds to expected funds in the currently-selected FY.
- [ ] When viewing a portfolio's summary, the user can quickly get an idea for the ratio of funds carried forward from prior fiscal years to funds new to the fiscal year.

-----------------------------------------

#### UX
- [x] Make any adjustments/edits to the design
- [x] Provide design assets to developers

#### Dev
- [x] In the backend, build in the math necessary to sum up all the prior FYs carry-forward numbers (received and expected)
- [x] Update the PortfolioSummaryCards components
- [x] Update the CanCard

### Definition of Done Checklist
- [ ] UI works as designed
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests updated and passed
- [x] Automated integration tests updated and passed
- [x] Automated quality tests updated and passed
- [x] Automated load tests updated and passed
- [x] Automated a11y tests updated and passed
- [x] Automated security tests updated and passed
- [x] 90%+ Code coverage achieved
- [x] PR(s) have been merged to main
- [x] Design/tech debt eliminated
- [x] Build process updated
- [x] Documentation updated or added
- [x] Feature flags/toggles created

### Additional Context & Resources
* Carry-forward funds have already been introduced at the overall portfolio level and possibly the CAN level on the backend. 
* 1 year CANs by definition wouldn't have any carry-forward funds.

kimschulke:
Updated designs in Figma

jonnalley:
Was just re-listening to the recording from 2/2 and wanted to clarify that for the PortfolioSummaryCards, that should be expected money, not just received.

fpigeonjr:
@xlorepdarkhelm the demo today looked good.

kimschulke:
[...]

PR #714, opened 2023-02-28
594 carry forward funds update

xlorepdarkhelm:
## What changed

- Added new field to the CANFiscalYearCarryForward model, replacng amount with "received_amount" and "expected_amount".
- Changed "current_funding" to "received_funding" for CANFiscalYear model.
- Summing up the received_amount & expected_amount for carry forwards into a single value, also have the backend send the text for the carry-forward tag for the front-end, with the correct FY put in it.
- Refactored "carry_over" and "CarryOver" to "carry_forward" and "CarryForward" respectively.
- Added new Carry-Forward tag for can cards that appears if that can has any carry-forward amount in it for the current FY.

## Issue

[Carry-forward Funds Update #594](https://app.zenhub.com/workspaces/opre-ops-62d197b4c468970013bcd2ec/issues/gh/hhs/opre-ops/594)

## How to test

- Log into site
- Go to portfolios
- Choose "Healthy Marriage & Responsible Fatherhood"
- See that there is carry-forward in the portfolio summary bar.
- Scroll down through CANs, and 2 of them have carry-forward tags.

## Screenshots

362FC414-5BE1-4AAB-BDB8-929F9FE652CD

FB33830D-614D-43D5-AF05-2279C813EB61


backend/ops_api/ops/utils/cans.py
```diff
@@ -1,43 +1,63 @@
 from typing import Optional, TypedDict
 
-from models.cans import CAN, BudgetLineItem, BudgetLineItemStatus, CANFiscalYear, CANFiscalYearCarryOver
+from models.cans import CAN, BudgetLineItem, BudgetLineItemStatus, CANFiscalYear, CANFiscalYearCarryForward
 
 
 class CanFundingSummary(TypedDict):
     """Dict type hint for total finding"""
```
tdonaworth, line 7: comment typo - should be `funding` 😉

backend/ops_api/ops/utils/cans.py
```diff
@@ -1,43 +1,63 @@
...
+    received_funding: float
     expected_funding: float
     total_funding: float
+    carry_forward_funding: float
+    carry_forward_label: str
     planned_funding: float
     obligated_funding: float
     in_execution_funding: float
-    available_funding: float
+    available_funding: str
```
tdonaworth, line 18: Curious why this is a `str`.

frontend/src/components/CANs/CANFundingYTD/CANFundingYTD.jsx
```diff
@@ -26,6 +33,14 @@ const CANFundingYTD = ({ total_funding = "0", current_funding = "0", expected_fu
     ];
     const [activeId, setActiveId] = React.useState(0);
 
+    const CarryForwardTag = ({ funding, label }) => {
+        if (funding > 0) {
```
tdonaworth, line 37: I forget, did we have a use-case where the carry forward could legit be negative (indicating an over spending)? I might be thinking of something else.

johndeange (approved):

backend/ops_api/ops/utils/cans.py
xlorepdarkhelm: I am not sure, we are doing formatting on the server side. I don't think that's what we should be doing. I can adjust it and make sure things work their way through correctly on the front-end.

backend/ops_api/ops/utils/cans.py
xlorepdarkhelm: Fixed

frontend/src/components/CANs/CANFundingYTD/CANFundingYTD.jsx
xlorepdarkhelm: This is just rendering the tag for those cans that have carry-forward ( greater-than 0), vs those that don't (equal-to 0).

PR #820, opened 2023-04-05
Ops 674 bl front end ps hookup

xlorepdarkhelm:
Hooking up backend to frontend for Procurement Shops.


fpigeonjr (approved):
Great job @xlorepdarkhelm. Pulled down locally and tested in local env. 
Works as expected.

Issue #844, opened 2023-04-11
Add Project and Agreement Summary Box 


fpigeonjr:
### Goals
Add Project and Agreement Summary Box.

### Tasks
- [X] add agreement data from state
- [X] add project data from state
- [X] add procurement shop data from state

### Additional Context
- Displays the project and agreement selected from the previous page.
- We also removed the procurement shop data field from creating the budget line user flow.
- If the agreement has existing/current budget lines, the procurement shop will auto-select what procurement shop is being used. The user will be unable to edit or change it. All budget lines on an agreement have to use the same procurement shop.

### Resources
<img width="936" alt="image" src="https://user-images.githubusercontent.com/4629398/231304668-420d324c-4188-45c1-bcdb-ea661396715a.png">

fpigeonjr:
:thinking: I noticed that the Procurement Shop select is not present anymore.
Does that mean that it will be set by the Agreement selected in Step One?

jonnalley:
> 🤔 I noticed that the Procurement Shop select is not present anymore.
> Does that mean that it will be set by the Agreement selected in Step One?

Yes, thank you!

PR #857, opened 2023-04-17
Task 844

fpigeonjr:
## What changed

Adds Project / Agreement Summary Card

## Issue

#844 

## How to test

Select and Project and Agreement on step 1, and then see the summary card on step 2.

## Screenshots

image

## Links


package.json
```diff
@@ -0,0 +1,5 @@
+{
```
tdonaworth, line 1: Is this in the right location? I think you mean to do the `yarn add` within the `frontend` dir.

yarn.lock
tdonaworth, line 1: ditto - wrong yarn.lock file

tdonaworth (changes requested):
Fix the package.json please

package.json
fpigeonjr: doh!

yarn.lock
fpigeonjr: will fix

package.json
fpigeonjr: these are fixed

yarn.lock
fpigeonjr: fixed

Issue #891, opened 2023-04-25
Create frontend components for Agreement/BLI Approval Request Page


tdonaworth:
- [x] Create a Agreement Summary Card Component
  - [x] Create (or reuse) BLI Funding Table Component
  - [x] Create/Add BLI Funding Totals Cards
  - [x] Add Edit Button
  - [x] Add Send-For-Approval Button 

Ref: 
![image](https://user-images.githubusercontent.com/56687505/234379803-8229cda7-448b-43b9-834a-2ef8c47efe81.png)

PR #893, opened 2023-04-25
adds Agreement Review page

fpigeonjr:
## What changed

adds Agreement Review page

## Issue

#891 

## How to test

for now, goto Agreement create steps and this will show as the last step in the wizard

### DIFF
--- a/frontend/src/pages/agreements/CreateAgreement.jsx
+++ b/frontend/src/pages/agreements/CreateAgreement.jsx
@@ -7,6 +7,7 @@ import { getProcurementShopList } from "../../api/getProcurementShopList";
 import { StepSelectProject } from "./StepSelectProject";
 import { StepCreateAgreement } from "./StepCreateAgreement";
 import { StepCreateBudgetLines } from "./StepCreateBudgetLines";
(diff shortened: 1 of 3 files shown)


frontend/src/pages/agreements/CreateAgreement.jsx
```diff
@@ -42,6 +43,7 @@ export const CreateAgreement = () => {
                 <StepSelectProject wizardSteps={wizardSteps} />
                 <StepCreateAgreement wizardSteps={wizardSteps} />
                 <StepCreateBudgetLines wizardSteps={wizardSteps} />
+                <StepReviewAgreement />
```
tdonaworth, line 46: I'm not sure if we want to add it as part of CreateAgreement or for it to be a stand alone page 🤔

tdonaworth (changes requested):
Lets discuss where this will live, and how it fits into the overall flow, as a team.

frontend/src/pages/agreements/CreateAgreement.jsx
tdonaworth: Lets roll with it for now, and then update once the AgreementList is added, we'll move it there.

PR #902, opened 2023-04-27
PATCH: Added a health check endpoint to API `/health/`

tdonaworth:
## What changed

- Added a `/health` endpoint to the API which returns a `status: "OK"` if healthy, can expand later to flush out further details like DB health, etc.


jonnalley (approved):
Health is good. Thank you!

Issue #920, opened 2023-04-28
After Adding a new Project, the Project Select isn't updated.


tdonaworth:
When going through the `Create Project` form, a Project is created as expected.

If you navigate to the `Create Agreement` form, the newly created `Project` is not in the `Project Select` list.

Issue #928, opened 2023-04-30
BLs from created Contract don't show up in review/send for approval screen


jonnalley:
After successfully creating a contract with 2 budget lines, when I am then directed to the list of agreements and I click the icon to send that agreement for review/approval, only one of the two budget lines show up in the table. 

## Expected Behavior
I expect for both BLs I created to then show up for review/approval

## Current Behavior
Only one BL shows in the table on /agreements/approve/

## Possible Cause
idk, not sure if the second BL was actually persisted or if it's just not showing for some reason, but it said the second BL was created. 

## Steps to Reproduce
Exact steps from [the demo script](https://docs.google.com/document/d/1o6bOk-LL2RrWO2L2-B-Q5b2N41CUO1mf7oCfzBSuqdE/edit?pli=1), no variance. In Chrome 

## Context
Dev cloud.gov environment

## Detailed Description

https://user-images.githubusercontent.com/928984/235475208-cd2d04fe-6e62-4cb7-92d5-eafeda67864d.mov

fpigeonjr:
think @tdonaworth has resolved this in [PR939](https://github.com/HHS/OPRE-OPS/pull/939).

jonnalley:
I concur regarding the resolution. Kudos @tdonaworth

PR #935, opened 2023-05-01
Patch create agreement project loading

tdonaworth:
## What changed

- Removed `{ }` from around the `ProjectSelect` prop `projects`; I think it was left in by mistake during testing.
- Reworked how the CreateProject object was getting the `id` back, post creation
- Cleared CreateProject form after submission

## Issue

#920


fpigeonjr (approved):
pulled down locally and works great. May solve #928?

Issue #940, opened 2023-05-02
Total Agreement amount is wildly inaccurate on Agreement List page


jonnalley:
When going through our script for the upcoming demo, when the user clicks `Continue` after adding all BLs, the user is sent to the list of agreements. On that list of agreements, the newly-created agreement has a wildly different total dollar amount than was displayed on the previous screen when adding BLs and also than the amount shown when clicking the arrow to send for review/approval.  

## Expected Behavior
For two CANs, one at $500 and one at $502, with a 4.8% fee, on the agreements list page, it should show $1,050.10.

## Current Behavior
Instead, for that agreement given the dollar amounts in the previous section, it shows a total of $5,811.60 in that total column on the agreements list. This example is currently in place in the live dev environment (freshly deployed from `main` Monday night). `Hope Testing Excellence` - id 10. Weirdly enough, when doing this again, I got very a different total amount (see attached video, it's highlighted at the end - $2,911.60 

## Possible Cause
idk

## Steps to Reproduce
1. Follow the script in the Google doc for our demo

## Context
cloud.gov development environment both before and after today's PRs/merges.   Chrome on macOS. 

https://user-images.githubusercontent.com/928984/235573847-272616c6-c570-415b-b7dd-37f2ad3b3b1e.mov

## Detailed Description
Proposing we use correct/accurate math on the agreements list just like on the cards on the adding BLs page

## Possible Implementation
1. Identify the problem
1. Fix it
1. Repeat until all problems are fixed.

PR #945, opened 2023-05-02
Ops-940 Proc Shop Fee

tdonaworth:
## What changed

 - Adjusted the `Procurement_Shop` fee to be a decimal value in the Data / DB
   - The intent here is to make it safer for calculations to use the decimal form, instead of relying on a `/ 100` each time it's used.
 - Added unit tests for AgreementTableRow
 - Updated the displayed Fee on CreateAgreement flow
 - Update to Readme, removed unused sections. 

## Issue

#940 

## How to test

`yarn test`


xlorepdarkhelm (approved):
LGTM

johndeange (approved):

PR #980, opened 2023-05-10
fixes table border when expanded

fpigeonjr:
When the agreement is expanded, there is no line separating the content

## What changed

removed border top and bottom for when it is expanded.

## Screenshots

image


johndeange (approved):
Is it possible to set the class to "border-bottom-none" at the parent component level and inherit it so that you don't have to dup it in each child component?

fpigeonjr:
> Is it possible to set the class to "border-bottom-none" at the parent component level and inherit it so that you don't have to dup it in each child component?

Looking into this @johndeange

fpigeonjr:
> Is it possible to set the class to "border-bottom-none" at the parent component level and inherit it so that you don't have to dup it in each child component?

😭 there doesn’t seem an easy way to do this since the style is conditionally added based on if the row is expanded.

PR #987, opened 2023-05-10
add health checks for the database and OAuth server (to /api/v1/health/)

stevtek:
## What changed

The health check endpoint, /api/v1/health/, now performs simple checks for the database and the auth server.
This is a experimental feature, but I don't think the /health endpoint is being used yet.
The pool_status may not make sense since it's only for the node it's executed on and you might have more than one in prod.  However, I put a timeout on the test query, so when the pool is busy it shouldn't wait for ages to respond.  It such a case seeing the pool is all checked out when the db conn test failed would be meaningful.

## How to test

Run the app and check http://127.0.0.1:8081/api/v1/health/

With both the database and OAuth server working the response would be

```json
{
  "alarm_level": 0,
  "checks": {
    "auth_services": {
      "alarm_level": 0,
      "status_code": 200
    },
    "database_connection": {
      "alarm_level": 0,
      "db_conn_is_ok": true,
      "pool_status": {
        "checked_in": 0,
        "checked_out": 1,
        "overflow": -4,
        "size": 5
      }
    }
  },
  "status": "OK"
}
```

With the database down (and the OAuth server working), the response would be

```json
{
  "alarm_level": 2,
  "checks": {
    "auth_services": {
      "alarm_level": 0,
      "status_code": 200
    },
    "database_connection": {
      "alarm_level": 2,
      "db_conn_is_ok": false,
      "pool_status": {
        "checked_in": 0,
        "checked_out": 0,
        "overflow": -5,
        "size": 5
      }
    }
  },
  "status": "BAD"
}
```

If you alter server_metadata_url in default_settings.py to something invalid (and if the database is working) the response would be 

```json
{
  "alarm_level": 2,
  "checks": {
    "auth_services": {
      "alarm_level": 2,
      "status_code": null
    },
    "database_connection": {
      "alarm_level": 0,
      "db_conn_is_ok": true,
      "pool_status": {
        "checked_in": 0,
        "checked_out": 1,
        "overflow": -4,
        "size": 5
      }
    }
  },
  "status": "BAD"
}
```

### DIFF
--- a/backend/ops_api/ops/resources/health_check.py
+++ b/backend/ops_api/ops/resources/health_check.py
@@ -9,18 +9,8 @@ from ops_api.ops.utils.health_check import check_all
 class HealthCheckAPI(MethodView):
     def get(self) -> Response:
         checks = asyncio.run(check_all())
-        max_alarm_level = max(checks.values(), key=lambda obj: obj["alarm_level"])[
-            "alarm_level"
-        ]
-        status = (
-            "OK"
-            if max_alarm_level == 0
-            else "BAD"
-            if max_alarm_level > 1
-            else "Maybe problems"
-        )
-        response = jsonify(
-            {"status": status, "alarm_level": max_alarm_level, "checks": checks}
-        )
+        max_alarm_level = max(checks.values(), key=lambda obj: obj["alarm_level"])["alarm_level"]
+        status = "OK" if max_alarm_level == 0 else "BAD" if max_alarm_level > 1 else "Maybe problems"
+        response = jsonify({"status": status, "alarm_level": max_alarm_level, "checks": checks})
         response.headers.add("Access-Control-Allow-Origin", "*")
         return response
--- a/backend/ops_api/ops/resources/health_check.py
+++ b/backend/ops_api/ops/resources/health_check.py
@@ -1,9 +1,26 @@
+import asyncio
+
 from flask import Response, jsonify
 from flask.views import MethodView
+
+from ops_api.ops.utils.health_check import check_all
 
 
 class HealthCheckAPI(MethodView):
     def get(self) -> Response:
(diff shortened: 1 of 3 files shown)


backend/ops_api/ops/utils/health_check.py
```diff
@@ -0,0 +1,69 @@
...
+    """check the health of the OAuth server with a GET of the server_metadata_url
+    Returns:
+        dict: with alarm_level>0 if there is a problem
+    """
+
+    resp = {"status_code": None, "alarm_level": 0}
+    try:
+        authlib_client_config = current_app.config["AUTHLIB_OAUTH_CLIENTS"]["logingov"]
+        server_metadata_url = authlib_client_config["server_metadata_url"]
+        print(server_metadata_url)
```
johndeange, line 19: would be better to use `current_app.logger` here

backend/ops_api/ops/resources/health_check.py
```diff
@@ -1,9 +1,26 @@
...
+            "OK"
+            if max_alarm_level == 0
+            else "BAD"
+            if max_alarm_level > 1
+            else "Maybe problems"
+        )
+        response = jsonify(
+            {"status": status, "alarm_level": max_alarm_level, "checks": checks}
+        )
         response.headers.add("Access-Control-Allow-Origin", "*")
```
johndeange, line 25: `ops_api.ops.utils.response.make_response_with_headers` is also available for this

johndeange (approved):

backend/ops_api/ops/resources/health_check.py
stevtek: k, I went ahead and changed it to use that common function.

tdonaworth (approved):

Issue #1007, opened 2023-05-15
Fix Expansion Row on Agreements List


kimschulke:
### Goals
* When the agreement is expanded, there is no line separating the content

### Tasks
- [x] Remove line in expansion row

kimschulke:
**Currently in Staging**
<img width="992" alt="Screenshot 2023-05-15 at 1 56 09 PM" src="https://github.com/HHS/OPRE-OPS/assets/65555918/4c5e5462-eec1-4b64-9371-57a47a99892c">


**Design in Figma**
<img width="802" alt="Screenshot 2023-05-15 at 1 56 31 PM" src="https://github.com/HHS/OPRE-OPS/assets/65555918/c96c1b4b-8989-4b37-bc78-9a0180a5fe59">

kimschulke:
Looks good!

PR #1015, opened 2023-05-16
removes table row border top/bottom when expanded

fpigeonjr:
## What changed

removes table row border top/bottom when expanded

## Issue

#1007 

## How to test

Goto agreements list and expand a table row.

## Screenshots

image


tdonaworth:
I kinda preferred the old look.. but not my call..   ✅

tdonaworth (approved):

Issue #975, opened 2023-05-09
User object not being populated or updated correctly.


xlorepdarkhelm:
The user object is not populated correctly on the front-end.

## Expected Behavior

The user object has relevant fields populated and is usable from the front-end.

## Current Behavior
Most if not all of the user object fields are null (like full_name, or first_name, etc).

## Possible Cause
?

## Steps to Reproduce
Look at the object and see.

## Context
This impacted the "Created By" issue that I had just address in #934

PR #1020, opened 2023-05-16
Ops 975 user object fix

xlorepdarkhelm:
## What changed

Made User object get loaded with RTK, added some tests around the created_by component to make sure it always has a value in it.

## Issue

#975 

## How to test

N/A Bug fix.

## Screenshots

N/A

## Links

N/A

### DIFF
--- a/frontend/src/components/UI/Form/CreateBudgetLinesForm.jsx
+++ b/frontend/src/components/UI/Form/CreateBudgetLinesForm.jsx
@@ -171,6 +171,7 @@ export const CreateBudgetLinesForm = ({
                     </div>
                 ) : (
                     <button
+                        id="add-budget-line"
                         className="usa-button usa-button--outline margin-top-2 float-right margin-right-0"
                         onClick={handleSubmitForm}
                     >
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -44,3 +44,11 @@ export const convertCodeForDisplay = (list_name, code) => {
     }
     return code;
 };
+
+export const loggedInName = (activeUser) => {
+    let loggedInUser = "Unknown User";
+    if (activeUser) {
+        loggedInUser = activeUser.full_name ? activeUser.full_name : activeUser.email;
+    }
+    return loggedInUser;
+};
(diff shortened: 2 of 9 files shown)


frontend/src/pages/budgetLines/PreviewTable.jsx
```diff
@@ -20,7 +20,9 @@ export const PreviewTable = ({ handleDeleteBudgetLine = () => {}, readOnly = fal
         .sort((a, b) => Date.parse(a.created_on) - Date.parse(b.created_on))
         .reverse();
 
-    let loggedInUser = useSelector((state) => state.auth.activeUser.full_name);
+    let loggedInUser = useSelector((state) =>
+        state.auth?.activeUser ? state.auth.activeUser.full_name : "Not logged in"
```
fpigeonjr, line 24: Maybe we can make a reusable helper function to display the user name. 1st try `full_name`, if not available then use `email`, and last option something like `unknown user`.

fpigeonjr (changes requested):
would be nice to have a helper function to display logged in user's full name, email, or unknown.

frontend/src/pages/budgetLines/PreviewTable.jsx
```diff
@@ -11,6 +11,14 @@ import { TotalSummaryCard } from "./TotalSummaryCard";
 import { formatDate } from "../../helpers/utils";
 import "./PreviewTable.scss";
 
+export const loggedInName = (activeUser) => {
```
fpigeonjr, line 14: Would it be possible to move this to our helpers file `frontend/src/helpers/utils.js`?

fpigeonjr (changes requested):
Would it be possible to move this to our helpers file `frontend/src/helpers/utils.js`?

frontend/src/pages/budgetLines/PreviewTable.jsx
xlorepdarkhelm: Done.

frontend/src/components/UI/Form/CreateBudgetLinesForm.jsx
```diff
@@ -171,6 +171,7 @@ export const CreateBudgetLinesForm = ({
                     </div>
                 ) : (
                     <button
+                        id="add-budget-line"
```
fpigeonjr, line 174: learned that the `data-cy` property is the recommended way to target elements for cypress tests.  So here we could do something like `data-cy="add-budget-line-btn"`

frontend/src/helpers/utils.js
```diff
@@ -44,3 +44,11 @@ export const convertCodeForDisplay = (list_name, code) => {
     }
     return code;
 };
+
+export const loggedInName = (activeUser) => {
```
fpigeonjr, line 48: thank you

Issue #996, opened 2023-05-12
Consistent casing on drop-downs


jonnalley:
### Goals
* Users need to see consistent casing in drop-downs in the UI, regardless of how the values comes in from the backend
* No underscores or hyphens please in drop-down values shown to users. 

### Additional Context
* For Procurement Shop we use Sentence Case and then for Reason for Agreement we use ALL CAPS and some words are abbreviated like “NEW_REQ”

PR #1021, opened 2023-05-16
OPS-996 convert type codes to display text

stevtek:
## What changed

Codes in ALL_CAPS (from ENUMs) were used as the display text of select boxes.  This changes those to use standardized friendly text using a method that can be used for any other situation where text needs to converted for display. 

## Issue

[OPS-996](https://github.com/HHS/OPRE-OPS/issues/996)

## How to test

View the Agreement Type and Agreement Reason select boxes in the create agreement form to see value like "CONTRACT" has been replace with "Contract".   And complete the create agreement process to verify that the valid codes are still being submitted.

### DIFF
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -20,3 +20,27 @@ export const formatDate = (date) => {
 
     return date.toLocaleDateString("en-US", options);
 };
+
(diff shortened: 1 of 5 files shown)


fpigeonjr (approved):
tested locally and all worked great!

PR #1058, opened 2023-05-29
Added workflow pre-condition to deploy on main push

tdonaworth:
## What changed

* Updated the `dev_deploy` to also trigger on the completion of CI workflow in `main`, resulting in the deployment to main, on any successful push to main, after CI completes. i.e.: All merged PRs will be deployed to DEV

### DIFF
--- a/.github/workflows/dev_deploy.yml
+++ b/.github/workflows/dev_deploy.yml
@@ -1,4 +1,4 @@
-name: Deployment
+name: Deployment DEV
 
 on:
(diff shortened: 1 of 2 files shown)


fpigeonjr (approved):
👍🏿

jonnalley (approved):
@tdonaworth Seems like this would also necessitate an update to the root README section on deployment instructions. Unless we anticipate this is such a temporary state it's not warranted?

PR #1066, opened 2023-06-01
Patch minor workflow updates

tdonaworth:
## What changed

- Updated ci.yml so that changes to `workflows` won't trigger `ci`, therefore won't trigger a `dev` deployment.
- Updated same rule change to `dev_deployment` and `staging_deployment`
  - If we determine later there are some conditions where we want workflow changes to trigger deployments, we can update it then.

## Issue

- TechDebt

## Testing

This shouldn't trigger a `dev` deployment when it's merged 🤞


fpigeonjr (approved):
👍🏿

johndeange (approved):

PR #1097, opened 2023-06-09
Ops 1089 ensure agreements filterable by status

xlorepdarkhelm:
## What changed

Agreement data now has a status. The status is calculated as the "lowest ranked" status for all BLIs on the agreement. So if there is at least 1 BLI for the agreement that is in `DRAFT`, then the agreement is also in `DRAFT`. This new field is simply "status". It is made as a read-only field that cannot be returned back through the API endpoints for PUT, PATCH, etc.

This field is also filterable, like other fields for agreements. specifying `/api/v1/agreements?status=PLANNED` will only show agreements that are in a `PLANNED` state.

## Issue

#1089 

## How to test

A pytest entry has been added to validate that searching for a specific status works.

## Screenshots

N/A

## Links

N/A

### DIFF
--- a/backend/models/cans.py
+++ b/backend/models/cans.py
@@ -111,46 +111,6 @@ class ProductServiceCode(BaseModel):
     agreement = relationship("Agreement")
 
 
-class BudgetLineItem(BaseModel):
-    __tablename__ = "budget_line_item"
-
-    id = Column(Integer, Identity(), primary_key=True)
-    line_description = Column(String)
-    comments = Column(Text)
-
-    agreement_id = Column(Integer, ForeignKey("agreement.id"))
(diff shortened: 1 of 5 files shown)


backend/models/cans.py
```diff
@@ -150,6 +151,14 @@ class Agreement(BaseModel):
 
     notes = Column(Text, nullable=True)
 
+    @property
+    def status(self):
+        subq = select(
```
tdonaworth, line 156: We may change this. We can keep it for now, but we will want to review this before utilizing it in the frontend or for any other business logic calculations.

johndeange (approved):
discussed with team

Issue #1146, opened 2023-06-26
investigate and experiment with authz provider


jonnalley:
### Goals
* Experiment with a custom/roll-your-own authz provider that fills the need that OPA would fill but recent consensus is that OPA is overkill. 

### Tasks
- [ ] 

### Additional Context
* #653 is a separate item for front-end permission framework and solves separate needs.
* #448 was a previous iteration of this effort when OPA seemed hopeful
 
### Resources
*

PR #1151, opened 2023-06-27
Ops 1146 poc authz

xlorepdarkhelm:
## What changed

Implemented basic authorization using the existing User/Roles and updating the permissions accordingly. All relevant endpoints (minus `/health` and `/auth`) now require JWT and has the new `is_authorized()` decorator added to them. All of the developer users have been given admin permissions which should unlock all endpoints.

Note: `/cans` still is left as not needing authorization. This is to allow an e2e test to continue to be used, and because the `/cans` endpoint isn't really finished.

Note 2: For making backend endpoints, the new `@is_authorized()` decorator *includes* the `@jwt_required()` decorator, so you do not need to specify both.

---

Permissions are split into 2 enums. A PermissionType, which specifies GET, PUT, POST, etc. and a Permission, which is tied to the kind of data/request that would be needed (AGREEMENT, BUDGET_LINE_ITEM, etc). These are combined together, and stored in the database's Roles as strings like "GET_AGREEMENT", or "PUT_RESEARCH_PROJECT", etc. This gives some flexibility.

Each endpoint is assigned to a particular permission. As of right now, it does not appear we need more than 1 permission per endpoint. But that could be needed in the future, and can be iterated on later.

## Issue

#1146 

## How to test

pytest runs all the checks, everything should be green.

Logging into the frontend, you should have no 401 errors.

cypress e2e tests also all work.

## Screenshots

N/A

## Links

N/A


backend/data_tools/data/user_data.json5
```diff
@@ -1,126 +1,126 @@
 {
   users: [
-    {
+    { // 1
       first_name: "Chris",
       last_name: "Fortunato",
       division: 1,
       email: "chris.fortunato@example.com",
       role: "COR",
```
johndeange, line 8: I think the `role` can be removed from the test User data here

backend/ops_api/ops/utils/authorization.py
```diff
@@ -11,17 +15,18 @@ class BasicAuthorizationPrivider(AuthorizationProvider):
...
+    def is_authorized(self, oidc_id: str, permissions: list[str]) -> bool:
+        stmt = select(User).where(User.oidc_id == oidc_id)
+        users = current_app.db_session.execute(stmt).all()
+        if users and len(users) == 1:
+            user = users[0][0]
+
+            for role in user.roles:
+                for permission in permissions:
+                    if permission in role.permissions:
+                        return True
```
johndeange, line 27: Do we want to ensure the User has all `permissions` here rather than one of them?

backend/ops_api/ops/resources/research_type.py
```diff
@@ -1,6 +1,12 @@
...
 from models.research_projects import ResearchType
 from ops_api.ops.base_views import EnumListAPI
+from ops_api.ops.utils.auth import is_authorized
+from typing_extensions import override
 
 
 class ResearchTypeListAPI(EnumListAPI, enum=ResearchType):  # type: ignore [call-arg]
-    pass
+    @override
+    @is_authorized("GET_RESEARCH_PORJECT", "GET_RESEARCH_PROJECTS")
```
johndeange, line 10: would be better if these were enums: `GET_RESEARCH_PORJECT`

backend/ops_api/ops/utils/authorization.py
xlorepdarkhelm: Well, I was seeing permissions a little differently - so if an endpoint has 2 permissions, like GET_USER and GET_USERS, then that means the endpoint allows for both options, not restricting someone to needing both to use that endpoint.

The way you're describing, it would be more and more restrictive on the endpoint, rather than permissive. That might be worth talking about as a team to come up with a consensus.

backend/ops_api/ops/resources/research_type.py
xlorepdarkhelm: I thought about that as well. Went back and forth a lot on it. Might be worth making them enums for clarity and some safety from mis-typing.

xlorepdarkhelm: I decided for now to make it 1 permission per endpoint. As we didn't need anything more than that. If we ever do, we can revisit this question and come with a good solution.

backend/ops_api/ops/utils/authorization.py
xlorepdarkhelm: I decided for now to make it 1 permission per endpoint. As we didn't need anything more than that. If we ever do, we can revisit this question and come with a good solution.

tdonaworth (approved):
Kudos 👍

stevtek (approved):

Issue #856, opened 2023-04-17
Edit Draft Agreement Metadata


jonnalley:
### User Story
As an OPRE COR or team lead, I want to alter some attributes of an agreement itself before it is in planned status.

### Preconditions
No budget lines are in "planned" status and thusly the agreement has not previously been approved. 

### Acceptance Criteria
- [ ] User can change any agreement-level metadata and have the changes persistently stored

-----------------------------------------

#### UX
- [x] Explore design options
- [x] Share designs with stakeholders and other OPRE team members for feedback
- [x] Refine the design based on feedback
- [x] Provide design assets with annotations to developers

#### Dev
***Backend***
- [x] #1017
- [x] #1043 @stevtek 
  - [x] Create Business Rules (gherkin features) for required field checks.
  - [x] Ensure Business Rules for metadata are checked for validity.

***Frontend***
- [x] #1088 - @stevtek @fpigeonjr 
- [x] #1115  - @fpigeonjr 
- [x] wire up pencil on agreements list
- [x] wire up agreement review edit button
- [x] Edit Agreement Design v2
   - [x] Agreement Details page #1156 <-- Lots of stuff in here 😉 
   - [x] Edit Agreement Details from/in details page
     - [x] #1209
   - [x] change edit links to new edit page (instead of wizard)

### Definition of Done Checklist
- [x] UI works as designed
- [ ] OESA: Code refactored for clarity
- [ ] OESA: Dependency rules followed
- [x] Automated unit tests updated and passed
- [x] Automated integration tests updated and passed
- [x] Automated quality tests updated and passed
- [x] Automated load tests updated and passed
- [x] Automated a11y tests updated and passed
- [x] Automated security tests updated and passed
- [x] 90%+ Code coverage achieved
- [x] PR(s) have been merged to main
- [x] Design/tech debt eliminated
- [x] ~Build process updated~
- [x] Documentation updated or added
- [x] ~Feature flags/toggles created~

### Additional Context & Resources
* Editing an agreement's budget lines is a separate story #681
* reduce value - changes tracked. contract mods?
* change metadata is a "mod" i.e. vendor, etc... at some point in the contract's status/lifecycle
* Once a contract is awarded, the PoP. Contract number, type, procurement shop, cannot be edited.
* Once BLs go into "in execution", a budget officer should change metadata, not other users

kimschulke:
[...]

PR #1159, opened 2023-06-28
disable edit agreement UI based on status (for OPS-856)

stevtek:
## What changed

Disable agreement edit based on status

## Issue

#856 

## How to test

View the agreements list page and verify that the edit icon is not shown if the agreement is not DRAFT or UNDER_REVIEW.
Also if try to directly go to the edit page via URL like http://localhost:3000/agreements/edit/2, then editing will be disabled.


frontend/src/pages/agreements/EditAgreement.js
```diff
@@ -47,6 +48,17 @@ const EditAgreement = () => {
     if (errorAgreement) {
         return <div>Oops, an error occurred</div>;
     }
+    if (agreement.status !== "DRAFT" && agreement.status !== "UNDER_REVIEW") {
```
fpigeonjr, line 51: no hackers 😄

fpigeonjr (approved):
tested locally and looks great @stevtek

johndeange (approved):

PR #1175, opened 2023-07-07
Ops 1171 backend logout

xlorepdarkhelm:
## What changed

Logout now has a handler in the backend, which currently just logs that the user logs out.

## Issue

#1171 

## How to test

Log in, log out, and then check the events table in the database to see the event is written.

## Screenshots

N/A

## Links

N/A


fpigeonjr (approved):
👍🏿

Issue #1183, opened 2023-07-11
Remove agreement-level statuses from existing models and pages


jonnalley:
### Goals
* Remove any UX, frontend validation, backend logic or validation, data models, or anything else  that assumes, implies, or enforces a "status" on the agreement or contract level. Recent research and determinations are such that statuses should only live at the budget line object.

### Tasks
- [x] If applicable, alter various components of the portfolio and research project page(s) (and models, etc..) that reference statuses for agreements and 
- [x] Update agreements listing page to remove the column for agreement status
- [x] Update logic and handling for creating new agreements such that a new agreement shouldn't try to set a status and the concept is removed entirely.
- [x] Update Gherkin rules for references to agreement-level statuses
- [x] Update e2e and cypress tests
- [x] Update the backend data model to remove the attribute/column

### Additional Context
* See figma for references and examples

### Resources
* _Links to useful related info (e.g., articles, help docs, etc)_

PR #1196, opened 2023-07-12
Ops 1183 remove agreement status

xlorepdarkhelm:
## What changed

Removed agreement.status from frontend and backend.

## Issue

#1183 

## How to test

pytest & e2e, can also verify by logging in and checking the agreement list to see there is nothing using or showin an agreement-level status.

## Screenshots

Screenshot 2023-07-12 at 10 21 16 AM

## Links

N/A

### DIFF
--- a/frontend/src/pages/agreements/EditAgreement.js
+++ b/frontend/src/pages/agreements/EditAgreement.js
@@ -49,7 +49,11 @@ const EditAgreement = () => {
         return <div>Oops, an error occurred</div>;
     }
 
-    if (agreement?.budget_line_items.every((item) => item.status !== "DRAFT" && item.status !== "UNDER_REVIEW")) {
+    const agreementStatus = agreement?.budget_line_items?.find((bli) => bli.status === "UNDER_REVIEW")
+        ? "In Review"
+        : "Draft";
+
+    if (agreementStatus !== "Draft" && agreementStatus !== "In Review") {
         return (
             <App>
                 <SimpleAlert
--- a/frontend/src/pages/agreements/EditAgreement.js
+++ b/frontend/src/pages/agreements/EditAgreement.js
@@ -5,7 +5,7 @@ import { useParams } from "react-router-dom";
 import { useGetAgreementByIdQuery } from "../../api/opsAPI";
 import { useEffect, useState } from "react";
 import { getUser } from "../../api/getUser";
-import { SimpleAlert } from "../../components/UI/Alert/SimpleAlert";
+import SimpleAlert from "../../components/UI/Alert/SimpleAlert";
 
 const EditAgreement = () => {
     const urlPathParams = useParams();
@@ -67,7 +67,7 @@ const EditAgreement = () => {
     return (
         <App>
             <CreateAgreementProvider agreement={agreement} projectOfficer={projectOfficer}>
-                <CreateEditAgreement existingBudgetLines={agreement.budget_line_items} isEditMode={true} />
+                <CreateEditAgreement existingBudgetLines={agreement.budget_line_items} />
             </CreateAgreementProvider>
         </App>
     );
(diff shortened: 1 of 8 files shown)


frontend/src/pages/agreements/list/AgreementTableRow.jsx
```diff
@@ -101,15 +99,13 @@ export const AgreementTableRow = ({ agreement }) => {
                         onClick={() => handleDeleteAgreement(agreement.id)}
                     />
 
-                    {(agreement.status === "DRAFT" || agreement.status === "UNDER_REVIEW") && (
```
johndeange, line 104: do we really want to remove this requirement?

frontend/src/pages/agreements/EditAgreement.js
```diff
@@ -48,17 +48,6 @@ const EditAgreement = () => {
     if (errorAgreement) {
         return <div>Oops, an error occurred</div>;
     }
-    if (agreement.status !== "DRAFT" && agreement.status !== "UNDER_REVIEW") {
```
johndeange, line 51: do we really want to remove this requirement?

xlorepdarkhelm: There is no agreement-level status any more. I believe Steve said he was working on something to fix it in his current issue. We'll definitely need to figure out how to limit this, but for the time being the purpose here was to remove the agreement level status from everything.

frontend/src/pages/agreements/list/AgreementTableRow.jsx
stevtek: yeah, I haven't sort out how this is supposed to work in the new design yet, but I thought for Cliff's ticket it would just be removed.  I'd like something at the Agreement level for this, if we are removing status, then I don't know what it is.  Can we include something like "is_editible" or "is_locked" in the GET Agreement(s) response?

frontend/src/pages/agreements/EditAgreement.js
stevtek: yeah, I haven't sorted out how this is supposed to work in the new design yet, but I thought for Cliff's ticket it would just be removed. 

I'd like something at the Agreement level for this, if we are removing status, then I don't know what it is.  It be nice if there was something that encapsulated the BLI.status logic so clients and server and trying to both looking through all the BLIs and compute the same thing.  Can I add something like "is_editible" or "is_locked" in the GET Agreement(s) response?

stevtek:
I just noticed while looking at the new design
https://www.figma.com/file/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?type=design&node-id=4658-143521&mode=design&t=3UHuInxWJrI8oyrh-4
It looks like the change icons (edit, delete, send) should be moved under the `Next Need By` column on hover the way they were under `Status` before.  Eliminating the status column altogether.
We had just renamed `Status` to `Actions` yesterday not being sure what to do.

frontend/src/pages/agreements/EditAgreement.js
```diff
@@ -5,7 +5,7 @@ import { useParams } from "react-router-dom";
 import { useGetAgreementByIdQuery } from "../../api/opsAPI";
 import { useEffect, useState } from "react";
 import { getUser } from "../../api/getUser";
-import SimpleAlert from "../../components/UI/Alert/SimpleAlert";
+import { SimpleAlert } from "../../components/UI/Alert/SimpleAlert";
```
fpigeonjr, line 8: don't think you need the curly bois here

frontend/src/pages/agreements/EditAgreement.js
```diff
@@ -59,7 +64,6 @@ const EditAgreement = () => {
             </App>
         );
     }
-
     return (
         <App>
             <CreateAgreementProvider agreement={agreement} projectOfficer={projectOfficer}>
```
fpigeonjr, line 69: isEditMode is not needed since it is derived from the query params

fpigeonjr (approved):
have some minor suggestions

stevtek (approved):

tdonaworth (approved):

Issue #1201, opened 2023-07-12
Create the GET endpoint for the ops_db_history


stevtek:
see also https://github.com/HHS/OPRE-OPS/issues/676#issuecomment-1631414989

Create a GET (list) endpoint for OpsDbHistory which supports querying by class_name and row_key and ordering by created_on desc so that we can query to find changes for a record with most recent changes first.

So URL might be like /ops-db-histories?class_name=BudgetLineItem&row_key=21
And maybe it should just default to recent changes first (created_on desc)

Also these histories could get big, so we'll need some sort of paging params
?class_name=BudgetLineItem&row_key=21&offset=10&limit=10
with appropriate defaults (offset=0, limit=10 ?)

We may want other queryables in the future like querying by user ID (created_by).

PR #1233, opened 2023-07-18
Ops 1201 get ops db history endpoint

xlorepdarkhelm:
## What changed

/ops-db-histories endpoint created as needed for the task.

## Issue

#1201 

## How to test

pytest has been added to validate that the endpoint runs.

## Screenshots

N/A

## Links

N/A


stevtek:
How do I set pagination params that will query with offset and limit?
If that's not already built-in to base lists, we will at least need that here.

xlorepdarkhelm:
> How do I set pagination params that will query with offset and limit? If that's not already built-in to base lists, I think we will at least need that here.

I don't believe that's set up. So I'll see about getting it in on this too.

xlorepdarkhelm:
> How do I set pagination params that will query with offset and limit? If that's not already built-in to base lists, I think we will at least need that here.

I have added query parameters for limit and offset, which are passed into the SQL query, which will give pagination capabilities.

stevtek (approved):

tdonaworth (approved):

PR #1269, opened 2023-07-27
Ops 1232 agreement filter button

johndeange:
## What changed

Added HTML/CSS for filter button.

N.B. The filter functionality has not been added yet, i.e. only the visual appearance for now.  That work with be next.

## Issue

https://github.com/HHS/OPRE-OPS/issues/1232

## How to test

Verify the visual appearance of the filter modal matches the design.

## Screenshots

image

### DIFF
--- a/frontend/cypress/e2e/reviewAgreement.cy.js
+++ b/frontend/cypress/e2e/reviewAgreement.cy.js
@@ -129,14 +129,14 @@ it("review an agreement", () => {
         // TODO: check to ensure the group date is in the future
         cy.get("#enteredMonth").select(month);
         cy.get("#enteredDay").clear();
-        cy.get("#enteredDay").type(day);
+        cy.get("#enteredDay").type(blData[0].day);
         cy.get("#enteredYear").clear();
         cy.get("#enteredYear").type(`${year - 1}`);
         // check for date to be in the future  which should error
         cy.get('[data-cy="date-group-errors"] .usa-error-message').should("exist");
         // fix by adding a valid date
         cy.get("#enteredDay").clear();
-        cy.get("#enteredDay").type(day + 1);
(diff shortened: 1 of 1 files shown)


frontend/src/pages/agreements/list/AgreementsFilterButton.module.css
```diff
@@ -0,0 +1,33 @@
+.filterButton {
+  background-color: #162E51;
```
fpigeonjr, line 2: wonder if we have global colors we can use?

frontend/src/pages/agreements/list/AgreementsFilterHeaderSection.jsx
```diff
@@ -39,9 +40,18 @@ export const AgreementsFilterHeaderSection = () => {
         );
     });
 
+    // useEffect(() => {
```
fpigeonjr, line 43: imagine this was left in for next step?

frontend/src/pages/agreements/list/AgreementsFilterButton.module.css
johndeange: I looked for it and didn't see it used elsewhere in the app.  In USWDS it isn't used other than for links where in this case it wouldn't semantically make sense.

johndeange:
@fpigeonjr I will look for a way to close the modal when you click outside of it - not sure if it's supported with the `react-modal` library

frontend/src/pages/agreements/list/AgreementsFilterButton.module.css
fpigeonjr: np, was hoping we had something like `bg-primary` or what have you.

xlorepdarkhelm (approved):

Issue #681, opened 2023-02-14
Modify existing draft budget lines (within contract)


jonnalley:
### User Story
As any OPS user, I want to update the attributes of a pre-existing draft budget line on a contract so that data in OPS is accurate and reliable.

### Preconditions
* A contract is already created and in existence with draft budget lines and no budget lines in any other status.

### Acceptance Criteria
- [x] The user can access and locate the draft contract containing the budget lines they wish to alter.
- [x] The user can choose to edit that contract's budget lines
~- [ ] The budget team, the division director (via CAN relationship), as well as any OPS user that's added as a lead or team member on the contract should be able to update any attribute of the draft BL.~ pending further discussions and work on authz and MVP scope
~- [ ] Other OPS users not in the list of users in the prior AC should not be able to edit the draft BL.~ pending further discussions and work on authz and MVP scope
- [x] User can successfully submit a change to values for the existing draft budget line and that change is persisted.
- [x] User should receive an indication of some kind that the edit was successful

-----------------------------------------

#### UX
- [ ] Identify the best way for OPS users to iterate/communicate on changes
- [x] Share designs with stakeholders and other OPRE team members for feedback
- [x] Refine the design based on feedback
- [x] Create clickable prototype
- [x] Provide design assets with annotations to developers

#### Dev
- [x] Validate this functionality is done (?) 
- ~~Update backend components to satisfy the AC~~
- ~~Build frontend interactions and components to satisfy the AC~~
- [x] Given the modified UI Designs, refactoring of the previous implementation of the Edit From should occur - tracked here: #1209 
- [x] Conduct experiment with newly-released SQLAlchemy features #1204 @johndeange 
- [x] #1274

### Definition of Done Checklist
- [ ] UI works as designed
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests updated and passed
- [x] Automated integration tests updated and passed
- [x] Automated quality tests updated and passed
- [x] Automated load tests updated and passed
- [x] Automated a11y tests updated and passed
- [x] Automated security tests updated and passed
- [x] 90%+ Code coverage achieved
- [x] PR(s) have been merged to main
- [x] Design/tech debt eliminated
- ~~Build process updated~~
-  ~~Documentation updated or added~~
-  ~~Feature flags/toggles created~~

### Additional Context & Resources
#677

kimschulke:
[...]

PR #1271, opened 2023-07-27
Agreement metadata and budget lines editing outside of wizard (from details page)

stevtek:
## What changed

* agreement metadata editing and budget lines editing has been refactoring into reusable forms
* new agreement details/metadata stand alone editor (from the details page)
* new budget lines stand alone editor (from the details page)
* edit link on the agreements list now goes to the new agreement editor

## Issue

#1209 
#681 
#856

## How to test
* create a test CONTRACT agreement which will verify that the refactored components work in the wizard UI
* on the agreements list, click on the title to go the Agreement details page
* click on the "Edit" link on the right of page
* makes change to the agreement metadata and click "Save Changes" at the bottom
* click on the "Budget Lines" tab
* click on the "Edit" link on the right of page
* edit the budget lines and click "Save Changes" at the bottom
  * NOTE: you cannot delete previously save budget lines yet

### Screenshot: Agreement Details Editor
image

### Screenshot: Budget Lines Editor
image


tdonaworth (approved):

xlorepdarkhelm (approved):

frontend/src/api/opsAPI.js
```diff
@@ -24,6 +24,76 @@ export const opsApi = createApi({
...
+                    method: "POST",
+                    headers: { "Content-Type": "application/json" },
+                    body: cleanData,
+                };
+            },
+            invalidatesTags: ["Agreements", "BudgetLineItems"],
+        }),
+        updateBudgetLineItem: builder.mutation({
+            query: ({ data }) => {
+                const cleanData = { ...data }; // make a copy
```
johndeange, line 78: The problem with this approach is that it makes assumptions about `data`, i.e. if another part of the app wanted to call `updateBudgetLineItem` it would be forced to pass in `data` in this particular format (which is awkward).  It would be better to normalize `data` in the callers at not here IMHO.  We can talk more in office hours on Monday.

johndeange: Not blocking the PR - this came up earlier too when Frank mentioned in Slack about discussing `opsAPI`

stevtek: Ok, we can merge this and discuss on Monday and do a small follow up PR.
The date_needed === "--" -> null seems like it doesn't belong in the API, but some of the cleaning is actually  removing fields that were in a GET that aren't allow in a PUT/POST, so that might be different.

PR #1290, opened 2023-08-03
Ops 1232 agreement filter button 2

johndeange:
## What changed

Added filter functionality to Agreements List page.

## Issue

https://github.com/HHS/OPRE-OPS/issues/1232

## How to test

Navigate to the Agreements List page and click on the Filters button.

Interact with the various filters, tags, and buttons to verify it works as expected.

## Screenshots

image


frontend/src/components/UI/Form/ProjectOfficerReactSelect.jsx
```diff
@@ -0,0 +1,113 @@
+import { useEffect, useState } from "react";
+import { useGetUsersQuery } from "../../../api/opsAPI";
+import Select from "react-select";
```
fpigeonjr, line 3: is this the new ComboBox?

frontend/src/pages/agreements/list/AgreementsTable.jsx
```diff
@@ -9,28 +9,32 @@ import { AgreementTableRow } from "./AgreementTableRow";
...
-                {agreements?.map((agreement) => (
-                    <AgreementTableRow key={agreement?.id} agreement={agreement} />
-                ))}
-            </tbody>
-        </table>
-    ) : (
-        <div className="padding-1 text-bold">No agreements found.</div>
+    return (
+        <div>
+            <table className="usa-table usa-table--borderless width-full ">
```
fpigeonjr, line 14: maybe don't show the table if there are no agreements?

fpigeonjr (changes requested):
Pulled down locally and filters work great.
A few suggestions: 
- Can we add the ability to tab into the React-Select components as a keyboard user.
- Forward Tabbing traps the focus inside the filter box but Shift-Tabbing (tabbing backwards) you can get out of the filter box
- 🤔Wonder if the React-Select has a cancel option to clear out the selection?

frontend/src/components/UI/Form/ProjectOfficerReactSelect.jsx
fpigeonjr: if so, should we remove the non-react-select `<ProjectSelect />` and `<ProjectOfficerSelect />`?

fpigeonjr:
🎥keyboard navigation video:
https://github.com/HHS/OPRE-OPS/assets/4629398/9ac7e70e-34f4-42b1-8c9d-1d9306e65a3b

frontend/src/components/UI/Form/ProjectOfficerReactSelect.jsx
johndeange: Yes that is the proposed new combobox.  If we are happy with `react-select` we should replace the other implementations, e.g. `<ProjectSelect />` and `<ProjectOfficerSelect />` with it.  Our current implementation has a number of bugs and UX issues and wasn't working in the filter which led me to pick this alternative.  `react-select` should be accessible and seems to be working well.

frontend/src/pages/agreements/list/AgreementsTable.jsx
johndeange: That was how it was implemented originally but the design was updated recently to show the table.

tdonaworth (approved):

stevtek (approved):

frontend/src/components/UI/Form/ProjectOfficerReactSelect.jsx
fpigeonjr: sounds good @johndeange. Sounds like we can use `react-select` to replace ComboBox across the app. May be a tech-debt item user story.

PR #1296, opened 2023-08-07
adds api call and wires up trash icon

fpigeonjr:
## What changed

- adds api call and wires up trash icon for deleting Agreements
- does not include:
  -  Create a notification, which reflects the status of the deletion (success or fail)
  - Enforce a subset of the rules on the frontend as well
  - Optional - give Tooltip feedback on why DELETE is disabled (?)

## Issue

#1294 

## How to test

1. create an agreement or add yourself to an agreement that only has DRAFT budget lines.
2. In the Agreement list, click on the trash can icon
3. 💰 profit

### DIFF
--- a/frontend/src/pages/agreements/list/AgreementsList.jsx
+++ b/frontend/src/pages/agreements/list/AgreementsList.jsx
@@ -36,11 +36,12 @@ export const AgreementsList = () => {
         data: agreements,
         error: errorAgreement,
         isLoading: isLoadingAgreement,
-        refetch,
+        refetch, // is this needed?
     } = useGetAgreementsQuery({ refetchOnMountOrArgChange: true });
 
     const activeUser = useSelector((state) => state.auth.activeUser);
 
+    // FSP@Flexion: Not sure if this is needed since we have the refetchOnMountOrArgChange: true
     useEffect(() => {
         refetch();
         // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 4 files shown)


frontend/src/pages/agreements/list/AgreementsList.jsx
```diff
@@ -36,11 +36,12 @@ export const AgreementsList = () => {
         data: agreements,
         error: errorAgreement,
         isLoading: isLoadingAgreement,
-        refetch,
+        refetch, // is this needed?
```
fpigeonjr, line 39: @johndeange Do you know if we need this?

frontend/src/pages/agreements/list/AgreementsList.jsx
```diff
@@ -36,11 +36,12 @@ export const AgreementsList = () => {
...
     } = useGetAgreementsQuery({ refetchOnMountOrArgChange: true });
 
     const activeUser = useSelector((state) => state.auth.activeUser);
 
+    // FSP@Flexion: Not sure if this is needed since we have the refetchOnMountOrArgChange: true
```
tdonaworth, line 44: I agree, worth a test, but the `refetchOnMountOrArgChange` should do the same thing with the cache/data if defined anyway.

frontend/src/pages/agreements/list/AgreementsList.jsx
johndeange: We have needed that previously - not sure if it still is post `opsApi` changes.

johndeange (approved):

PR #1311, opened 2023-08-08
OPS-1310 - delete BLIs when agreement is deleted.

xlorepdarkhelm:
## What changed

(Draft) BLIs are all deleted from the agreement when the agreement is deleted.

## Issue

#1310 

## How to test

Delete an agreement, and the BLIs connected to it are deleted. All pytest tests are working.

## Screenshots

N/A

## Links

N/A


fpigeonjr (approved):

backend/ops_api/ops/resources/agreements.py
```diff
@@ -271,6 +271,9 @@ def delete(self, id: int) -> Response:
                 elif any(bli.status != BudgetLineItemStatus.DRAFT for bli in agreement.budget_line_items):
                     raise RuntimeError(f"Agreement {id} has budget line items not in draft status.")
 
+                for bli in agreement.budget_line_items:
```
tdonaworth, line 274: should we catch any specific conditional errors that might occur during the BLI deletion, that would result in the Agreement not getting deleted? (thought process, if something goes wrong during the BLI.delete, then we probably shouldn't continue on deleting the Agreement (orphaned BLIs at that point). I presume the `try` would capture that, but it doesn't look explicit.

open to thoughts

stevtek (approved):

fpigeonjr (approved):

johndeange (approved):
I think this can be done more cleanly in a CASCADE DELETE in the SQLAlchemy model.

xlorepdarkhelm:
> I think this can be done more cleanly in a CASCADE DELETE in the SQLAlchemy model.

Good point, I adjusted it to do just that.

Issue #1300, opened 2023-08-07
View an agreement's audit trail/history


jonnalley:
### User Story
As an OPRE staff member, I want to view the history of changes and a record of the lifecycle of an agreement as part of other agreement details so that I can understand the full scope of all alterations to a contract.

### Acceptance Criteria
- [x] The user can view and see a list of all changes that occurred to a contract since its inception including the date of the change, what user made the change, and a succinct definition of what was altered.
- [x] The retention of the history should be exhaustive with no limit as to the age of a change
- [x] The history/list of all changes shouldn't be editable by regular users accessing the OPS interface
- [x] The list of items in the history should be sorted in descending order chronologically

-----------------------------------------

#### UX
- [x] Design already provided and discussed within #1156

##### Backend
- [x] #1361
  - [x] track team_members, support_contacts
  - [x] make history for BLIs, etc queryable by agreement ID
  - [x] fix issues with unwanted histories - BLIs, User getting history records with no changes with an Agreement is saved
- [ ] ~#1377~ saving for later 
- [x] agreement history endpoint that provides combined history of agreement with the history of it's BLIs, etc.
- [x] OpenAPI spec documentation

##### Frontend
- [x] #1385
- [x] Infinite scroll
- [x] simplify prepared changes https://github.com/HHS/OPRE-OPS/pull/1383#discussion_r1311727716

### Definition of Done Checklist
- [x] UI works as designed
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests updated and passed
- [x] Automated integration tests updated and passed
- [x] Automated quality tests updated and passed
- [x] Automated load tests updated and passed
- [x] Automated a11y tests updated and passed
- [x] Automated security tests updated and passed
- [x] 90%+ Code coverage achieved
- [x] PR(s) have been merged to main
- [x] Design/tech debt eliminated
- [x] Build process updated
- [x] Documentation updated or added
- [x] Feature flags/toggles created

kimschulke:
**Quotes from user research / feedback on agreement details design**
__* “Anytime there is an addition or subtraction, a notification would be good”_
_* “There's just all these, like responding to changes”_
_* “Seeing the changes in real time would be really helpful”_
_* “Right now I have to communicate the amount to project officers and then someone has to communicate it to the budget team to make the changes in MAPS but its nice to have that be a shared responsibility with CORS so they can do those changes themselves but I can still see what's happening”_
[...]

PR #1312, opened 2023-08-08
OPS-1361 Agreement History backend - including OpsDbHistory changes

stevtek:
## What changed

Agreement History with team members, budget lines, etc included

### OpsDbHistory
* new "changes" column storing data from SqlAlchemy's get_history
  * includes lists of added, deleted values so collections like team_members can track changes here 
  * for simple column values this has old, new values like description: { "old": "old desc", "new":"new desc"}
  * for relationships like team_members which are collections, this has arrays of add, deleted values
  * the original and diff columns might not be needed anymore
* adds tracking of changes to secondary, non-viewonly relationships like team_members
* omit history records for UPDATE with no changes
  * otherwise BLIs, USERs were getting changeless UPDATE records saved with Agreements

### Agreement History endpoint
* /agreements/\<id\>/history/
* specialized form of OpsDbHistory that queries Agreements, BLIs and any object that has an agreement_id

## Issue

#1361
part of #1300 

## How to test

* run tests
* add/edit/delete records and view the ops_db_history table
* for a modified agreement GET it's history /agreements/\<id\>/history/


backend/models/history.py
```diff
@@ -25,6 +25,8 @@ class OpsDBHistory(BaseModel):
     row_key = sa.Column(sa.String)
     original = sa.Column(JSONB)
     diff = sa.Column(JSONB)
+    changes = sa.Column(JSONB)
+    agreement_id = sa.Column(sa.Integer)
```
tdonaworth, line 29: Is there a way to do this without tying the `agreement_id` directly to the record? Thinking future, what if we wanted to query history for other specific models, like `CAN`? Would we need to embed the `can_id` as well? User History - `user_id`? etc. 

Is this data already accessible from the `original`, `diff`, or `changes` JSON blobs?

stevtek: I think it's possible without this column, but I haven't actually tried writing that alternative.  You need to query the even_details JSONB to find the agreement ID in different place depending on the class - ID for class_name in 'Agreement', 'ContractAgreement', 'GrantAgreement', ... or agreement_id for BLIs, etc.  I wasn't sure about creating an index for that later on.
I was thinking that Agreement was important enough to justify this, but not everything.  So far this Agreement-Details the only place in the UI where we collect all the history.  I wondered in Can or project might be.  User is probably different since it's already a column if you are looked at who made the change.

I can try the alternative using JSONB queries if that's preferred.

stevtek: WIP: here's a query using event_details JSONB to replace what's in agreement_history.py

```
stmt = select(OpsDBHistory).join(OpsDBHistory.created_by_user, isouter=True).add_columns(User)
stmt = stmt.where(
    and_(
        or_(
            and_(
                OpsDBHistory.event_details['id'].astext.cast(Integer) == id,
                OpsDBHistory.class_name.in_(class_names),
            ),
            OpsDBHistory.event_details['agreement_id'].astext.cast(Integer) == id
        ),
        OpsDBHistory.event_type.in_(
            [
                OpsDBHistoryType.NEW,
                OpsDBHistoryType.UPDATED,
                OpsDBHistoryType.DELETED,
            ]
        ),
    )
)
stmt = stmt.limit(limit)
if offset:
    stmt = stmt.offset(int(limit))
stmt = stmt.order_by(OpsDBHistory.created_on.desc())
```

If we want to go this route and eliminate agreement_id, we could consider eliminating row_key as well.  Although it is also being used to handle composite PKs, I'm not sure that any of the tables like that have Classes that will be logged in OpsDbHistory.

I am a little unsure about setting up indexes for this later on.  Before it was just three columns, now it's using a mix of stuff inside JSONB with other columns.

stevtek: agreement_id removed and JSONB query used instead

tdonaworth (approved):

johndeange (commented):
strangely you have Frank's recent changes in this PR too

johndeange (approved):

PR #1329, opened 2023-08-10
Ops 909 replace team member select

johndeange:
## What changed

Replace TeamMemberSelect with new ComboBox
Fix bugs

## Issue

https://app.zenhub.com/workspaces/opre-ops-62d197b4c468970013bcd2ec/issues/gh/hhs/opre-ops/909

## How to test

See original issue and attempt to reproduce.


fpigeonjr (changes requested):
💯 pulled locally and looks and functions great @johndeange 
one small request is to increase the width since its a little squishy.
can we make it the same width as in the incumbent field above it?

image

tdonaworth (approved):

stevtek (approved):

PR #1383, opened 2023-08-22
View Agreement History

stevtek:
## What changed

Added panel that displays the Agreement history on the Agreement details page.

This PR does not include infinite scrolling, it just gets the first 20 most recent history records for now.  Also there is a bug with get_history that reports certain properties as changed when they haven't, for example agreement_reason.  I'll do other PRs for that.

## Issue
#1385 (part of #1300)

## How to test

Create and edit an Agreement and it's Budget Lines and view the History panel on the Agreement details page.

There is also some unit and e2e tests.

## Screenshots

image


frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
```diff
@@ -0,0 +1,173 @@
...
+            const added = change.collection_of == "User" ? usersToNames(change.added) : change.added;
+            const deleted = change.collection_of == "User" ? usersToNames(change.deleted) : change.deleted;
+            return {
+                propertyLabel: getPropertyLabel(historyItem.class_name, key),
+                isCollection: true,
+                added: added,
+                deleted: deleted,
+            };
+        } else if (key === "procurement_shop_id") {
+            const new_val = historyItem.event_details?.procurement_shop?.name;
```
tdonaworth, line 55: Suggested Refactor for future updates: 
I could see breaking this bit out into a function like `generateChangeObject` or something to `DRY` it out a bit. This could also help reduce the number of `if/else if`, since there's a clear correlation between the `key === {name}` and the `event_details.{property}`, so a simple mapping lookup or something, to do actual output, since the logic is the same across most.

tdonaworth (approved):
Comment for future refactor, but I know this is an ongoing WIP 👍

frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
```diff
@@ -0,0 +1,173 @@
...
+    if (historyItem.class_name === "BudgetLineItem") {
+        return historyItem.event_details.line_description;
+    } else {
+        return historyItem.event_details.name;
+    }
+};
+
+const summaryMessage = (historyItem) => {
+    const className = convertCodeForDisplay("className", historyItem.class_name);
+    const user_name = historyItem.created_by_user_full_name;
```
fpigeonjr, line 17: should user_name be userName to be consistant camelCase where possible?

frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
```diff
@@ -0,0 +1,173 @@
...
+        if ("collection_of" in change) {
+            const added = change.collection_of == "User" ? usersToNames(change.added) : change.added;
+            const deleted = change.collection_of == "User" ? usersToNames(change.deleted) : change.deleted;
+            return {
+                propertyLabel: getPropertyLabel(historyItem.class_name, key),
+                isCollection: true,
+                added: added,
+                deleted: deleted,
+            };
+        } else if (key === "procurement_shop_id") {
```
fpigeonjr, line 54: think a switch statement will make this a little easier to read

frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
```diff
@@ -0,0 +1,173 @@
...
+            return {
+                propertyLabel: getPropertyLabel(historyItem.class_name, key),
+                from: change.old,
+                to: change.new,
+            };
+    });
+
+    return (
+        <dl>
+            {prepChanges.map((change) => (
```
fpigeonjr, line 92: `prepChanges` is a little ambiguous...maybe a rename to something a little clearer like `logList`?

frontend/src/pages/agreements/details/AgreementDetailsView.js
```diff
@@ -46,25 +47,7 @@ const AgreementDetailsView = ({ agreement, projectOfficer }) => {
...
-                                    title={note.created_by}
-                                    createdOn={note.created_on}
-                                    message={note.message}
-                                />
-                            ))}
-                        </ul>
-                    ) : (
-                        <p>Sorry no history</p>
-                    )}
+                    <AgreementHistoryPanel agreementId={agreement.id} />
```
fpigeonjr, line 50: good name for this component

fpigeonjr (approved):
💪looks great @stevtek. Added some non-blocking suggestions.

frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
stevtek: yeah, it can be improved.  I got part way into an idea and reverted to `if/else-if` as I wondered if maybe a little backend work could help simplify this.

Two things keep this from being almost a universal one liner, not just here but anywhere we want to turn a FK ID into a display string.  All but `project_officer`, use `<related_object_name>_id` for the FK ID.  And then the property name to display varies for classes -  User -> `full_name`, Agreement -> `name`, ResearchProject -> `title`, BudgetLineItem -> `line_description`.  So I am wondering if maybe there should be some back end changes first.  Changing project_officer to project_officer_id would be a breaking change and the project_officer user isn't included anyway - so maybe not that.  But can we add something to the base model, say `display_name` with a getter implemented on each class in a way that it's included in the JSON without messing with the ORM/DB?  If so, I could also use this for the summary message to say which Agreement/BLI was edited and use it for common `to_slim_dict()` - just ID + display_name.

frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
stevtek: done

frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
stevtek: see Tim's comment and my reply about some refactoring coming later

frontend/src/components/Agreements/AgreementDetails/AgreementHistoryPanel.jsx
stevtek: I made a slight clarification.  I didn't go with logList since to me that's sounds like list of history records rather than the field-level changes in a single history record.

Issue #1406, opened 2023-08-25
created by user for BLs might be showing incorrectly?


jonnalley:
When I go to view the budget lines tab of any contract and expand out the details of any budget line, it shows my user as the user that created that budget line, even for data created by the `data_tools` container.  This info might be hardcoded to show my user, or, it could just be showing my user in the absence of a real user who created it. 

## Expected Behavior
I would expect that a correct non-hard-coded user be shown for who actually create the BL and if not known, for it to just be blank or say unknown or NA or something.

## Current Behavior
Currently all BLs show that the currently logged in user created that BL.

## Possible Cause
hardcoding?   BLs without any created_by user ?  Just guessing

## Steps to Reproduce
See above

https://ops-staging.app.cloud.gov/agreements/1/budget-lines

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaHM2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--4232c68c909e554de74820f6a36d12e7389db668/Screenshot%202023-08-25%20at%2003.27.52.png)

## Context
Staging environment as recently as 08/24

PR #1422, opened 2023-09-01
Hotfix/bug 1406

fpigeonjr:
## What changed

adds custom hook `useGetFullNameFromId` to return full name from user id.
- return `unknown` if id is `null`

## Issue

#1406


jonnalley:
👏

stevtek (approved):

PR #1435, opened 2023-09-06
Agreement History - Infinite Scrolling

stevtek:
## What changed

Implemented infinite scrolling for the Agreement History panel.
This gets pages of 20 records as one scrolls to the bottom of the history panel.
This implementation uses Axios instead of RTK due to issues with this approach to infinite scrolling and RTK.

## Issue

#1385 #1300 

## How to test

Edit an agreement and/or it's BLIs (multiple times) and view the history on the details page.  Scroll in the history panel to load more history records.  You can edit the pageSize in frontend/src/api/getAgreementHistory.js if you want to test with smaller pages and more loads.

## Screenshots

image


johndeange (approved):

fpigeonjr (approved):
pulled locally and works good

PR #1481, opened 2023-09-19
Ops 1462 refactor tokens frontend

tdonaworth:
## What changed

* Created `getAccessToken()` and `getRefreshToken()` as wrappers for accessing the tokens; future updates may change where the tokens live, so this makes it easier to refactor that later.
* Created checks, on the frontend, so that if the `access_token` is expired, it will actually call the backend's `/refresh` endpoint, and get a new access token
* Added `iss` and `aud` claims to our tokens (via default configs in `flask_jwt_extended` config
* Added an `iss` check, with temporary values, simply to help fill out the logic
* Created a `ValidateTokenStatus` object, to capture the current, and future status of tokens when checked.
* Created unit-tests for the `isValidToken` function
* Updated the `callBackend()` function, now takes an optional parameter, which determines if the `access_token` is used or `refresh_token`; defaults to `false` which is `access_token`; :information: I thought about simply passing in the functions themselves, it would be more scalable, but in reality we only have two token types.

## Issue

#1462

## How to test

1. Login with your preferred provider (`FakeAuth`)
2. Wait the desired time that the access_token should live (default is 30 min). 
3. After that time period, you should still have a valid session, but a new access_token.
4. The access_token["fresh"] value should be `false`


frontend/src/components/Auth/auth.test.js
```diff
@@ -1,4 +1,11 @@
-import { getAuthorizationCode } from "./auth";
+import { getAuthorizationCode, isValidToken } from "./auth";
+
+const expiredToken =
+    "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJmcmVzaCI6dHJ1ZSwiaWF0IjoxNjk1MTQ5NTkxLCJqdGkiOiJlOGUwMTY2ZS1lNTYyLTQ3N2UtOWJiMy05MjA1OTFiNmEyMjUiLCJ0eXBlIjoiYWNjZXNzIiwic3ViIjoiMDAwMDAwMDAtMDAwMC0xMTExLWExMTEtMDAwMDAwMDAwMDE4IiwibmJmIjoxNjk1MTQ5NTkxLCJleHAiOjEwMDAwMDAwMDEsImlzcyI6Imh0dHBzOi8vb3ByZS1vcHMtYmFja2VuZCIsImF1ZCI6Imh0dHBzOi8vb3ByZS1vcHMtZnJvbnRlbmQiLCJyb2xlcyI6WyJhZG1pbiJdfQ.RWvGtcZCxGGaqX_28LSfFqcza32sU4zZnfC_niCJP3X6V9vTRGjONEfLoZS9_TjGbx0TZHHMdZRnCWF2LW_84mTJEhizqad8nmVHnagvehfrjUcS9Fe1wn6ZvjlQfSiX9iibgfXjV0WjlNeKuNx3nLhcSt0yKiwpw-7BoPUjQKYQcjmNYfL6rzqxj85MIsHo09ifxKDmmOePiFqZEgyUDakovSMrS-nGFif3sArUC7dmJRuEO3KdFGD9Dc9rY1RItlBYl4xSlq3bZDwDmH5DWlJ_22LjW2Xq04LJwLFbagOjcsDQhRIOHevviX6S3xBfzZf_NDx8ZFhj3FXIItQ4QQw";
```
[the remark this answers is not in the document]
johndeange: seems like we need to add a semgrep exception here?

backend/ops_api/ops/utils/auth_views.py
```diff
@@ -108,6 +114,9 @@ def _get_token_and_user_data_from_internal_auth(user_data: dict[str, str]):
         # authZ, given a valid login from the prior AuthN steps above.
 
         additional_claims = {}
+        additional_claims["iss"] = "https://opre-ops-backend"  # Placeholder
```
johndeange, line 117: not sure of the function of the placeholder but the string is duplicated - can it be put in a constant instead?  Does the value, e.g. `https://opre-ops-backend` have any meaning or can empty string or None be used instead?

frontend/src/components/Auth/MultiAuthSection.jsx
```diff
@@ -29,8 +29,9 @@ const MultiAuthSection = () => {
                 console.error("API Login Failed!");
                 navigate("/login");
             } else {
-                console.log(`DEBUG:::ACCESS_TOKEN: ${response.access_token}`);
+                // console.log(`DEBUG:::ACCESS_TOKEN: ${response.access_token}`);
```
johndeange, line 32: please remove commented code

frontend/src/components/Auth/auth.test.js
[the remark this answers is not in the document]
tdonaworth: I did add one, but it's not picking it up, I did flag them as "Test code only" on the finding.

johndeange (approved):

backend/ops_api/ops/utils/auth_views.py
tdonaworth: Basically, anything is allowed; it simply represents principal that issued the JWT, so that you can validate it wherever needed (so ensure in the frontend, that a specific entity generated the JWT. The placeholder is currently used and checked; Some value is preferred in a default (enabled) check state, versus an empty string (tyranny of defaults). It probably should be in the config or a constant; but wasn't sure "what" to use yet 🤔

fpigeonjr (approved):

backend/ops_api/ops/utils/auth_views.py
tdonaworth: moved these to the config via: `JWT_ENCODE_ISSUER` and `JWT_ENCODE_AUDIANCE`

stevtek (approved):

Issue #1429, opened 2023-09-05
Hover State on Table Rows Not Working Properly


kimschulke:
## Expected Behavior
When a user hovers their mouse over a single row within a table, that row should appear with a light grey hover state

## Current Behavior
Currently, when you mouse over the table quickly, the hover states gets stuck on multiple rows simultaneously

## Possible Cause
No idea what the cause could be

## Steps to Reproduce
1. Navigate to any page with a table (~~Agreements List~~ or Budget Lines List)
2. Move your mouse across the table (up and down, back and forth)
3. This will cause you to see multiple rows in a hover state at one time

## Context
 Dev, Admin User
 The hover state is intended to show the user where their mouse is hovering over. If multiple rows are showing a hover state it becomes impossible to know which one is accurate and may be confusing to the user.

## Detailed Description
 Improve hover state so only row will be in hover state at a single time.

kimschulke:
Example here: 
<img width="1059" alt="Screenshot 2023-09-05 at 3 35 12 PM" src="https://github.com/HHS/OPRE-OPS/assets/65555918/8d3b7b74-7dd5-4bd3-b14a-43243852cfbe">

tdonaworth:
Updated the Bug to reflect only BLI List page.

fpigeonjr:
@jonnalley will move on to more pressing issues if you would like to return this bug to the proper placement in the backlog

jonnalley:
9/28

PR #1516, opened 2023-09-27
fix: :bug: fixes table row hover

fpigeonjr:
## What changed
fix: :bug: fixes table row hover
uses JSX expression and passes into slots

issue #1429


jonnalley:
Fascinating. I haven't looked at it more to understand, but I'm curious how this fixes the display issue

xlorepdarkhelm (approved):
LGTM, don't see hover problems any more.

tdonaworth (approved):

Issue #1502, opened 2023-09-25
Update `/login` to store `refresh_token` somewhere (along with `user_id` and expiration)


tdonaworth:
Determine where to store the `refresh_token`:
- ( ) Store in backend, attached to user record?
- (x) Store in frontend, but in secure `HttpOnly` cookie?

Justification: Currently, we're storing the `refresh_token` openly in the `localStorage`, which isn't very secure. Since the `refresh_token` is used to generate new `access_token`s, it should be secured.

jonnalley:
Curious how much overlap there is between this and the intent of #663

PR #1551, opened 2023-10-05
Ops 1502 token storage

tdonaworth:
## What changed

- Fixed past issues with flask_cors, so CORS works correctly
  - This potentially makes the `make_response_with_headers` useless; since the addition of the ACAO header there was actually breaking the CORS() initialization; basically any headers need to be managed via CORS() now. Will evaluate the continued need as we progress.
- Added `OPS_FRONTEND_URL` to backend configuration, to support the `origin` parameter for CORS().
- Introduced the creation of cookies for `access_token` and `refresh_token`
  - Only the refresh_token cookie is going to be used via this PR
  - Additional refactoring would need to occur before we can switch over to the `access_token` cookie; since we currently rely on access to the JWT data within the frontend; if we wanted to properly secure the cookie (Secure=true; HttpOnly=true) then we wouldn't have access to the JWT anymore, and would need to pull in that data another way (API Calls).
- Removed `refresh_token` from localStorage.
- Adjusted NotificationCenter to only update every 1 minute (60 seconds)
  - This is based off feedback we heard from the client, indicating it doesn't need to be as often.

## Issue

#1502 

## How to test

1. Login per your favorite method.
2. Check your browsers cookies, you should have both `access_token` and `refresh_token` cookies.
3. Validate that the `refresh_token` cookie is working correctly, by letting the `access_token` expire (default 30 min). You should continue to get a new `access_token`, indicating that the `/refresh` endpoint is now validating with the `refresh` cookie instead of the `localStorage` value.

## Screenshots
N/A

- [] 90%+ Code coverage achieved


fpigeonjr (approved):

stevtek (approved):

frontend/src/components/UI/NotificationCenter/NotificationCenter.jsx
```diff
@@ -23,7 +23,7 @@ const NotificationCenter = () => {
         data: notifications,
         error,
         isLoading
-    } = useGetNotificationsByUserIdQuery(userId, { pollingInterval: 5000 });
+    } = useGetNotificationsByUserIdQuery(userId, { pollingInterval: 60000 });
```
jonnalley, line 26: we all thank you

PR #1635, opened 2023-11-02
Using a pinned version of the newly-upgraded semgrep container

jonnalley:
## What changed

Using a pinned version (based on sha) of the `semgrep` container rather than relying on `latest`. Among other benefits, it helps us avoid bugs they inadvertently introduce...

## Issue

https://github.com/returntocorp/semgrep/issues/9091

## How to test

See if CI passes 😉 

## Links

https://hub.docker.com/layers/returntocorp/semgrep/1.47.0/images/sha256-e7216f0ec1eaa032f470733e33d4a49d556f04b82fdce5e066599c38c1334c32

### DIFF
--- a/.github/workflows/security_semgrep.yml
+++ b/.github/workflows/security_semgrep.yml
@@ -9,7 +9,7 @@ jobs:
     name: Semgrep Analyze
     runs-on: ubuntu-latest
     container:
-      image: returntocorp/semgrep@sha256:4ddf4c7908f6cd56bd703ddcac2dcf4ad6e000dba7835a75fe0b9a2181f4a325 # 1.47.0
+      image: returntocorp/semgrep@sha256:e7216f0ec1eaa032f470733e33d4a49d556f04b82fdce5e066599c38c1334c32 # 1.47.0
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/security_semgrep.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     name: Semgrep Analyze
     runs-on: ubuntu-latest
     container:
-      image: returntocorp/semgrep
+      image: returntocorp/semgrep@sha256:e7216f0ec1eaa032f470733e33d4a49d556f04b82fdce5e066599c38c1334c32 # 1.47.0
```
tdonaworth, line 12: shouldn't we also swap over to the `semgrep/semgrep` version?

tdonaworth (approved):

.github/workflows/security_semgrep.yml
jonnalley: @tdonaworth So I looked at and tried that but it seemed to bomb out because even though their GitHub org has changed, in the docker registry it's still `returntocorp`. A 404 error is returned. I could have been missing something but that's what I experienced anyway

PR #1866, opened 2024-01-16
testing successive bun updates > 1.0.16

jonnalley:
## What changed

PR #1754 is attempting to update several packages to newer versions. The particular issue going on with the CI workflows on that PR is that the runner can never fully get the full installer for `cypress` completely downloaded and fails to get completely downloaded. On occasion when re-attempting, it complains about other similar symptomatic issues related to performance and timeouts. This PR is an attempt to isolate just the bun upgrade without upgrading anything else to see what behavior is experienced with each version after the current one (1.0.16). 

## How to test

See if PR checks succeed


jonnalley:
accomplished what I needed to with this testing, will likely close

PR #1905, opened 2024-01-22
chore: Just changed the source to relative path for now.

devops-chris:
## What changed

Changed the source to relative path until we can figure out a better solution. 

## Issue

Renovate changed the tags for the context module to something pre-terraform.


tdonaworth (approved):

jonnalley (approved):

fpigeonjr (approved):
👍

PR #1941, opened 2024-02-07
fix: Add NODE_ENV

devops-chris:
Should work without, but tested to require some NODE_ENV

## What changed

Added NODE_ENV to the entrypoint where the runtime env is set. 

## Issue

Runtime env injection stopped working. Documentation indicates the NODE_ENV is not required for this to work, but testing proves it must be set to something... anything but development.


jonnalley (approved):
good luck!

tdonaworth (approved):

PR #2072, opened 2024-03-14
Cml/add alembic dev

devops-chris:
## What changed

- Built for Alembic to run in data-tools for cloud
- Updated the deployment to support. 
- Moved to non-terraform deployments
- Cleaned up


backend/alembic/env.py
devops-chris, line 1: @johndeange - I particularly wanted your eyes on this, just for awareness. Alembic was just defaulting to the local env with no access to the env variables used for deployments, so I added those here.

tdonaworth (approved):

johndeange (changes requested):

johndeange (approved):

PR #2140, opened 2024-03-29
linting along with black 24 upgrade

jonnalley:
## What changed

Upgrades `black` to v24 and re-lints the backend

Also disables E704 in `.flake` config due to disagreement between linters/formatters about whether you can have multiple statements on one line. See inline review for more info

## Issue

[CVE-2024-21503](https://nvd.nist.gov/vuln/detail/CVE-2024-21503)

### DIFF
--- a/backend/ops_api/.flake8
+++ b/backend/ops_api/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 select = B,B9,BLK,C,E,F,S,W
-ignore = E203,E501,W503
+ignore = E203,E501,W503,E704
 per-file-ignores = tests/*:S101
 
 max-complexity = 10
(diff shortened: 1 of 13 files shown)


backend/ops_api/.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 select = B,B9,BLK,C,E,F,S,W
-ignore = E203,E501,W503
+ignore = E203,E501,W503,E704
```
jonnalley, line 3: Without this change, there were rampant complaints about ` E704 multiple statements on one line (def)`

From reading I did, this was a common complaint about black 24 and mixing multiple linters/formatters together that don't agree at the moment. Recommending disabling unless there is a better path forward

fpigeonjr (approved):

devops-chris (approved):

PR #2225, opened 2024-04-23
Cml/automate versioning

devops-chris:
## What changed

- Added a community action for bumping and tagging package.json for front end.
- Created an action based off of the same features as the front end but for openapi.yml and backend. 
- Added the actions to the workflow so that after bumping and tagging occurs, deployments to dev are made on the same versions. 

## Issue

Version automation is needed in order to keep versions in-sync with their deployments and to create point in time tagging.


fpigeonjr (approved):
🤩

stevtek (approved):

johndeange (approved):

PR #2335, opened 2024-05-21
feat: ✨ adds vitest-ui

fpigeonjr:
## What changed

Adds vitest ui

## How to test

`bun run test:ui`

## Screenshots

image

## Links

https://vitest.dev/guide/ui.html#vitest-ui


fpigeonjr:
🤔 oddness, FE tests are working locally?
image

fpigeonjr:
seems like adding a package breaks the build?

jonnalley:
@fpigeonjr definitely seems like testing is behaving differently and it's being more picky about things than before.  I presume you've checked out all the complaints about things being undefined
https://github.com/HHS/OPRE-OPS/actions/runs/9176573976/job/25250174242

fpigeonjr:
yes but not clear on the path forward 😞

PR #2505, opened 2024-06-28
hiding login.gov authn option for anything but local dev

jonnalley:
## What changed

Given [John's recent PR](https://github.com/HHS/OPRE-OPS/pull/2494) that ripped out the config for login.gov in azure and given that our `redirect_uri `on the login.gov side only works for `localhost` anyway and not any hosted environment, I modified the conditional to only show the login.gov authn option when doing local dev. The intent is to create a better UX so that an unknowing user doesn't try to innocently auth with their login.gov sandbox account.

However, this presumes that we don't set the environment variable to something for the non-static FE container build or for local, which seems to be the current practice for the time being. 

## Issue

N/A

## How to test

Hit the front login page.  For any azure non-prod environment, you should see only AMS or AMS plus FakeAuth.  On localhost you should see all 3.

### DIFF
--- a/frontend/src/components/Auth/MultiAuthSection.jsx
+++ b/frontend/src/components/Auth/MultiAuthSection.jsx
@@ -111,7 +111,7 @@ const MultiAuthSection = () => {
                         You can access your account by signing in with one of the options below.
                     </p>
                 </div>
-                {!import.meta.env.PROD && (
+                {import.meta.env  && ( //testing for existence of the env being set at all since we don't seem to do that for local environments
                     <p>
                         <button
                             className="usa-button usa-button--outline width-full"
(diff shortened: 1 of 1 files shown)


fpigeonjr (approved):

johndeange (changes requested):
When I tested this locally it didn't work for me.  I don't think `import.meta.env` is valid syntax.  https://vitejs.dev/guide/env-and-mode.html

This is the only reliable way I was able to do what I think is intended:

- In local dev display all 3 options
- In Azure non-prod show 2 options
- In Azure prod show 1 option

`{import.meta.env.MODE === "development" && (
                    <p>
                        <button
                            className="usa-button usa-button--outline width-full"
                            onClick={() => handleSSOLogin("logingov")}
                        >
                            Sign in with Login.gov
                        </button>
                    </p>
                )}`

jonnalley:
@johndeange it comes back as valid syntax in my local env but I see what you mean and yeah that parent object's existence isn't what I'm trying to achieve. I don't really want to test for the dev environment though because we don't want it visible for the azure dev environment either.  I switched it up to hopefully be more accurate although not statically built.  Please re-review and let me know what you think.

johndeange (approved):

Issue #2059, opened 2024-03-11
Add model `CANAppropriationDetails`


johndeange:
Based on comments from 3/6/24 meeting:

> CAN.APPROP_PREFIX, CAN.APPROP_POSTFIX, CAN.APPROP_YEAR can be considered together as the appropriation details.

rajohnson90:
Should this have a foreign key to any existing tables?

PR #2526, opened 2024-07-02
Add CANAppropriationDetails model and CANStatus enum

rajohnson90:
## What changed

Adding CANAppropriationDetails model and table. Additionally, added the CANStatus enum with two values: Active and Inactive. This is now present on the CAN as the status property, which is calculated based on a combination of funding and expiration date. 
## Issue

[OPS-2059](https://github.com/HHS/OPRE-OPS/issues/2059)

## How to test

Start the app and look at the can_appropriation_details table in the database. There should be 3 rows.
Run unit tests


fpigeonjr (commented):
looks good but will defer to other devs

backend/data_tools/data/can_data.json5
```diff
@@ -623,5 +623,22 @@
       current_fy_funding_ytd: 100000,
       can_fiscal_year_id: 5004
     }
+  ],
```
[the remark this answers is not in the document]
rajohnson90: Adding it to tests data for now. If things are different when we get to CANs we can modify the models as needed later.

frontend/src/tests/data.js
```diff
@@ -436,3 +436,62 @@ export const changeRequests = [
         updated_on: "2024-06-17T18:55:41.826716"
     }
 ];
+
+export const canFiscalYearFundingDetails = [
```
fpigeonjr, line 440: ⭐

fpigeonjr (approved):
🟢 tests pass

backend/models/cans.py
```diff
@@ -690,6 +693,26 @@ class CAN(BaseModel):
         "Project", secondary="project_cans", back_populates="cans"
     )
 
+    @property
+    def status(self):
+        if object_session(self) is None:
+            return CANStatus.INACTIVE
```
johndeange, line 699: I think it would be OK to throw an exception if `object_session(self) is None` - that will never happen when the app is running normally

backend/models/cans.py
```diff
@@ -690,6 +693,26 @@ class CAN(BaseModel):
...
+                and_(
+                    CANFiscalYear.can_id
+                    == self.id,
+                    CANFiscalYear.fiscal_year == current_year
+                )
+            )
+        )
+        if can_fiscal_year is None:
+            return CANStatus.INACTIVE
+        can_status = CANStatus.INACTIVE if can_fiscal_year.expected_funding <= 0 and self.expiration_date < datetime.now() else CANStatus.ACTIVE
```
johndeange, line 712: I would create a method `can_fiscal_year.available_funding()` (or something like that) that uses the calculations I pointed you to this morning - it should be something like `total funding - amount spent`

backend/models/cans.py
rajohnson90: Removed this code instead of having a specific exception handler. I expect that trying to reference None will just throw the exception as you suggest.

backend/models/cans.py
```diff
@@ -711,3 +757,15 @@ class CANFiscalYearFundingDetails(BaseModel):
...
+
+class CANAppropriationDetails(BaseModel):
+    """
+
+    """
+
+    __tablename__ = "can_appropriation_details"
+
+    id: Mapped[int] = BaseModel.get_pk_column()
+    appropriation_prefix: Mapped[Optional[str]] = mapped_column(String)
```
johndeange, line 769: The `= mapped_column(String)` is redundant but doesn't hurt anything if you keep it in.

johndeange (approved):

Issue #2504, opened 2024-06-28
Handle concurrent uses of FakeAuth users


jonnalley:
Currently, our session management and security practices would block users from different IPs trying to have sessions for the same FakeAuth user.

## Expected Behavior
Legitimate human users should be able to use our limited set of FakeAuth users 

## Current Behavior
After the first person using a FakeAuth user at any given time, session management will be kicking in to block additional uses of the same user from a distinct  IP address.

## Possible Cause
Intentional logic to prevent valid/active sessions for the same user from a different IP. 

## Possible Implementation
* Eliminate the limitation of multiple sessions for FakeAuth users (AND not inside a unit test)   OR:
* Add additional distinct FakeAuth users for each role

jonnalley:
7/11

PR #2555, opened 2024-07-10
Ops 2504 concurrent fake users

johndeange:
## What changed

Disable `UserSession` check for `FakeAuth`

N.B. This effectively causes the `UserSession` to be "shared" with other computers/browsers logged in as the same "Fake" user.

## Issue

https://github.com/HHS/OPRE-OPS/issues/2504

## How to test

1. Log in as the Admin User
2. Find the active `UserSession` record in the DB and change the IP address
3. Log in as the Admin User using a different browser, e.g. Edge or Firefox

Previous to this change, step 3 would have prevented you from logging in using the second browser and logged you out in the first.  The `UserSession` record in the DB would be set to `is_active=false`.

With this change, the ip address isn't checked for fake users and so theoretically you should be able to log in using multiple browsers/computers.


fpigeonjr (approved):
tested locally and admin logon worked on both Firefox and Edge browsers. Not sure but believe this was working locally before the changes were put in place so not sure local tests are the best.

fpigeonjr:
> tested locally and admin logon worked on both Firefox and Edge browsers. Not sure but believe this was working locally before the changes were put in place so not sure local tests are the best.

switched back on `main` and was also able to logon with Admin user on both browser in `localdev`

backend/ops_api/ops/__init__.py
```diff
@@ -173,4 +171,6 @@ def before_request_function(app: Flask, request: request):
     ]
     if request.endpoint in all_valid_endpoints and request.method not in ["OPTIONS", "HEAD"]:
         verify_jwt_in_request()  # needed to load current_user
-        check_user_session_function(current_user)
```
jonnalley, line 176: 💥

johndeange:
I agree @fpigeonjr that the local testing options are tricky.  What we are trying to simulate is different people/computers logging in as the same fake user - using different browsers is a poor approximation of that.  I will update the testing instructions to make the test a little better but still this will be from the same computer - the actual test will be when the change is deployed to the dev environment in Azure and different testers can log in from different machines.

maiyerlee (approved):
Kudos for the straightforward solution! The log before checking user session will be helpful when debugging as well!

PR #2563, opened 2024-07-12
Change `appropriation_term` to be a calculated property instead of stored field

rajohnson90:
## What changed

I updated the appropriation_term to be a calculated property instead of a value stored in the database. I had to update the CAN's to_dict method to add the appropriation_term in order for everything relying on the can funding summary to still work without any changes. 

## Issue

[OPS-2062](https://github.com/HHS/OPRE-OPS/issues/2062)

## How to test

Run unit tests
## Screenshots

## Links


fpigeonjr (approved):
🟢 tests pass

backend/models/cans.py
```diff
@@ -770,10 +769,24 @@ def status(self):
         )
         return can_status
 
+    @property
+    def appropriation_term(self):
+        if self.expiration_date is None or self.appropriation_date is None:
```
johndeange, line 774: It may be slightly better to return `None` when `self.appropriation_date is None` - `0` year money means the funding doesn't expire and so has a different meaning from if `appropriation_date` is `NULL`/unknown

johndeange (approved):

backend/models/cans.py
```diff
@@ -770,10 +769,24 @@ def status(self):
...
+            return 0
+        return self.expiration_date.year - self.appropriation_date.year
+
```
stevtek, line 777: Do we know this should be calendar year and not fiscal year?
Or does the calculation end up the same anyway?

stevtek (approved):

backend/models/cans.py
rajohnson90: I do not know if it's fiscal year or calendar year. That would be a bit harder to calculate because I'd need to define the fiscal year somewhere but it's definitely not impossible.

backend/models/cans.py
rajohnson90: I'll make this change.

Issue #441, opened 2022-09-12
Story: Disable User Accounts


jonnalley:
### User Story
As an OPS System Admin, I want to disable or remove named users for OPS so that I can better control who is accessing and using OPS and meet all requisite compliance needs.

### Acceptance Criteria
- [x] A user account can be marked as disabled and then re-enabled an indefinite number of times with appropriate log or event history entries recording metadata of each action
  - [x] A notable exception to this should be that the user (User Admin) may not disable their own user account if there are no other active users possessing the System Admin role.
- [x] A user can view a list of pre-existing user accounts which are active
- [x] A user can view a list of pre-existing user accounts which include both active/enabled and inactive/disabled accounts.
- [x] A user that's marked as disabled will not succeed in an attempt to log in to OPS

-----------------------------------------

#### Dev:

- [x] Create persistent store schema related to the concept of a user account's binary status
- [x] Create backend code related to the retrieval, listing, and updating of a user account's status
- [x] Create frontend code related to the retrieval, listing, and updating of a user account's status

### Definition of Done Checklist

- [x]  UI works as designed (UX team)
- [x]  PR(s) have been merged to main
- [x]  Design/tech debt eliminated
- [x]  New design/tech debt documented (if applicable)
- [x]  Build process updated
- [x]  Documentation updated or added
- [x]  Feature flags/toggles created

### Additional Context & Resources
* This is "locking" in the database. As opposed to "inactive"
* This story does not cover topics related to using the users for anything or tying to 3rd party authn/IdP services, etc...
* A "user" in this context mentioned in the Acceptance Criteria refers to a human consumer of OPS, not a logged in or authenticated user.  This is a bootstrapping effort to get RBAC/User management in place
* Will we need outbound email to confirm the action?

kimschulke:
Users that don't log in for 60 days will become inactive users

jonnalley:
I updated #2104 to handle the 60-day auto disable

jonnalley:
I updated #2104 to handle the 60-day auto disable

jonnalley:
8/12

PR #2610, opened 2024-07-24
Ops 441 disable user accounts refactoring

johndeange:
## What changed

Added missing functionality to `/users` endpoint
Added USER_ADMIN role
Re-designed endpoint according to service layers.

## Issue

https://github.com/HHS/OPRE-OPS/issues/441

## How to test

Automated tests pass.


jonnalley:
@johndeange first off, 👏 🎉  but secondly, it seems at this point the notion of a "admin" wouldn't map to any known human at OPRE. Maybe the better thing to do is rename the current "admin" role the Budget Team

backend/data_tools/data/user_data.json5
```diff
@@ -137,8 +135,8 @@
         "POST_RESEARCH_PROJECT",
 
         "GET_USER",
-        "GET_USERS",
```
jonnalley, line 140: am struggling to remember why we would have had GET_USERS (plural) in here

backend/data_tools/data/user_data.json5
```diff
@@ -218,6 +216,12 @@
         "GET_UPLOAD_DOCUMENT",
         "POST_UPLOAD_DOCUMENT"
       ]
+    },
+    { // 5 User Admin
+      name: "USER_ADMIN",
+      permissions: [
+        "POST_USER",
+      ],
```
jonnalley, line 224: @johndeange so..... why not additional roles here?  I saw a few places in the code doing checks for `is_user_admin` on modifying or creating a user which seems to just be checking if a user has a role but it's not actually a check of the permission....    Is that by design?  I recognize there's a high chance of me asking out ignorance and/or out of context.

backend/data_tools/data/user_data.json5
```diff
@@ -137,8 +135,8 @@
         "POST_RESEARCH_PROJECT",
 
         "GET_USER",
-        "GET_USERS",
-
+        "PUT_USER",
+        "PATCH_USER",
```
jonnalley, line 139: @johndeange Up to this point I had presumed that `USER` only affected the the same user as the one doing the verb whereas `USERS` would be something affecting a user other than themselves. I'm guessing that's not correct or at least not any longer. ?

johndeange: The Permissions are currently all singular.  I think eventually `GET_USER` will allow access to the `GET` `/users/<id>` endpoint and `GET_USERS` will allow access to the `GET` `/users` endpoint.

johndeange:
@jonnalley

Necessity of "admin" role - the idea here is the System Owner will be assigned the "admin" role and the "USER_ADMIN" would be delegated by the System Owner to a couple team members to create/update users.   I think there is still a need for a role for the Budget Team.

backend/data_tools/data/user_data.json5
johndeange: See answer below

backend/data_tools/data/user_data.json5
jonnalley: Ah, thank you. I'm better informed now :-)

maiyerlee (approved):
Looks good! Thanks for keeping the event type and initial data up to date. I like the concept of a safe user!

fpigeonjr (approved):
🟢 tests pass

PR #2668, opened 2024-08-12
An ode to types-all in Pipfiles

jonnalley:
## What changed

Removing `types-all` in backend and data-tools and explicitly adding in the `numpy` dependency for backend

## Issue

In our Pipfiles `dev-packages` for backend and data-tools, we were referencing `types-all`.  The [source repo was archived a couple years ago](https://github.com/asottile-archive/types-all) and considered abandoned and deprecated by various folks consuming it. Additionally, types-all itself had some dependencies that are now also deprecated and being yanked from various repos. Including but not limited to `types-pkg-resources`.  Because of this, fresh attempts to build and update the `Pipfile.lock` were failing locally and here in GHA. As a result, many renovatebot PRs are not successful as they need to be.

~A [fork](https://github.com/Morry98/types-all) of `types-all` has emerged as `types-all-v2` and [is recognized by](https://pypi.org/search/?q=typeshed) `pypi` and others keeping most everything about the original `types-all` but with  more up-to-date dependencies. In this PR  I'm recommending we at least update to it but would also like to suggest we consider more holistic approaches of whether we need to enumerate out individual [stubs](https://github.com/python/typeshed/tree/main/stubs) from [typeshed](https://github.com/python/typeshed) or do something entirely different.  Presuming this dependency is even needed still at all....~

After some discussion, it seems this package was only needed due to limitations of older versions of python that we're no longer restricted by. When removing types-all, numpy needed to be added explicitly as a new dependency as it's no longer coming over to us.

## How to test

See actions/tests in this PR and once merged, try out renovate PRs again.

## Links

In addition to links above:
https://snyk.io/advisor/python/types-all
https://stackoverflow.com/questions/78833765/no-matching-distribution-found-for-types-pkg-resources
https://pypi.org/project/types-pkg-resources/
https://github.com/python/typeshed/blob/main/README.md


maiyerlee (approved):
Thanks for investigating and explaining the issues. Let's remove it!

johndeange (approved):

Issue #2852, opened 2024-09-25
Backend should not allow approvals for change requests from division directors from other divisions


fpigeonjr:

jonnalley:
10/9

PR #2876, opened 2024-10-02
Lock Down Change Requests to DD and DDD

rajohnson90:
## What changed

Locking down the patch change-requests endpoint to only allow a Division Director or Deputy Division Director to approve/deny change requests from within their own division. 

## Issue

#2857 #2852 

## How to test

1. Run python tests
2. Run E2E tests


backend/ops_api/tests/ops/can/test_budget_line_item.py
```diff
@@ -106,7 +106,7 @@ def test_post_budget_line_items(loaded_db, auth_client, test_can):
         "proc_shop_fee_percentage": 1.23,
         "services_component_id": 1,
     }
-    response = auth_client.post("/api/v1/budget-line-items/", json=data)
+    response = division_director_auth_client.post("/api/v1/budget-line-items/", json=data)
```
johndeange, line 109: I don't think we want to limit operations on BLIs, e.g. creating a new BLI, to only the division director.  These tests should still work with the regular unit test user, i.e. the `auth_client` user.

johndeange (changes requested):
I think the unit test changes should probably be limited to the `test_change_requests.py` tests, i.e. it doesn't seem like the agreement history tests or the BLI tests should require a division director user for them to pass.

backend/ops_api/tests/ops/can/test_budget_line_item.py
rajohnson90: Good call. I over-corrected originally when switching the auth client. All the posts are using the basic auth_client now and only the patch requests use the division director auth client.

rajohnson90: Also the reality is that the tests in test_budget_line_items are only updated because when I changed the change requests these tests were breaking. If I didn't include these fixes my backend tests would not pass.

rajohnson90: Sorry my fixes for test_budget_line were not needed and a knockoff error caused by my other configuration problems.

fpigeonjr (commented):
talked it over with Reed but unable to test on the FE. 
- Will defer to others.

fpigeonjr (commented):
talked it over in ENG Collab 10/3/24 and sounds like had found a solution with Director Derreck. 
- will defer to others

maiyerlee (approved):

PR #2966, opened 2024-10-22
fix: remove budget pills with state using defaults

maiyerlee:
One way to potentially solve not showing the budget pills when there is no user interaction with the budget range slider. 

https://github.com/user-attachments/assets/35668d70-5c98-4893-af6c-361d0e661cfc


fpigeonjr (approved):
this is working so great! I have been trying to use the activateBudget state option but could not get that working. This is a much simpler and elegant solution.

PR #3023, opened 2024-11-04
chore: adding requested user stakeholders for non-prod environments

jonnalley:
## What changed

Adding in a couple of real human users with AMS accounts for non-prod environments

## Issue

N/A

## How to test

Verify if the named users can log in


fpigeonjr (approved):

johndeange (approved):
In order for the new user to log in, they need to have at least 1 role, e.g. `roles: [{"tablename":  "role", "id": 1}],`

jonnalley:
> In order for the new user to log in, they need to have at least 1 role, e.g. `roles: [{"tablename": "role", "id": 1}],`

I knew I was forgetting something

PR #3382, opened 2025-01-28
Ops 3294/load can history

johndeange:
## What changed

- Add `CANHistory` to `CAN` load script.
- Move `CANHistory` to `models` so that it is accessible outside of the API.
- Add `FISCAL_YEAR` to the test data format.

## Issue

https://github.com/HHS/OPRE-OPS/issues/3294

## How to test

Automated tests pass.


fpigeonjr (approved):
🧪 tests pass

maiyerlee (approved):

backend/data_tools/src/load_cans/utils.py
```diff
@@ -194,6 +234,9 @@ def get_or_create_funding_details(data: CANData, sys_user: User, session: Sessio
             funding_source=funding_source,
             funding_partner=funding_partner,
             created_by=sys_user.id,
+            updated_by=sys_user.id,
+            created_on=datetime(data.FISCAL_YEAR, 10, 1),
```
rajohnson90, line 238: Wouldn't this logic lead to your fiscal year always being FISCAL_YEAR + 1? 

Just thinking about it, but if you are loading up fiscal year 2025, but place the created_on date on 10/1, won't that make this a can for Fiscal Year 2026? Maybe that's fine, I just wanted to make sure that is intentional

rajohnson90 (changes requested):
Just have one point to check in on. Everything else about the PR, including the refactor, looks fine to me

backend/data_tools/src/load_cans/utils.py
johndeange: Yes you are correct - we add the `FISCAL_YEAR` column to the spreadsheet though and so we can set it to whatever we want.  I think it's ambiguous as to what the intention is if you set it to 2025, i.e. do you mean the start of the FY or the end?  But maybe it would be better if it was the start of the FY so I think I will make that change...

rajohnson90 (approved):

PR #3391, opened 2025-01-30
Hide portfolios UI menu option temporarily

jonnalley:
## What changed

Removes the Portfolios menu option from the horizontal navigation in OPS based on discussions this week

## Issue

N/A

## How to test

Log in and see if the Portfolios menu option is displayed

### DIFF
--- a/frontend/cypress/e2e/portfolioDetail.cy.js
+++ b/frontend/cypress/e2e/portfolioDetail.cy.js
@@ -1,19 +1,20 @@
 /// <reference types="cypress" />
 import { terminalLog, testLogin } from "./utils";
 
-    beforeEach(() => {
-        testLogin("system-owner");
-        cy.visit("/portfolios/1");
-        cy.get("#fiscal-year-select").select("2023");
-    });
+beforeEach(() => {
+    testLogin("system-owner");
+    cy.visit("/portfolios/1");
+    cy.get("#fiscal-year-select").select("2023");
+});
 
-    afterEach(() => {
-        cy.injectAxe();
-        cy.checkA11y(null, null, terminalLog);
-    });
+afterEach(() => {
+    cy.injectAxe();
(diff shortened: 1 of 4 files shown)


fpigeonjr (changes requested):
wonder if using Vite's environment variables may be a better approach:

```
import.meta.env.DEV && (
   // section to only show in DEV
)
```

another consideration would be to block the route in the `index.jsx` similar to what we are doing with the CAN pages

fpigeonjr (changes requested):
wonder if using Vite's environment variables may be a better approach:

```
import.meta.env.DEV && (
   // section to only show in DEV
)
```

another consideration would be to block the route in the `index.jsx` similar to what we are doing with the CAN pages

jonnalley:
> wonder if using Vite's environment variables may be a better approach:
> 
> ```
> import.meta.env.DEV && (
>    // section to only show in DEV
> )
> ```
> 
> another consideration would be to block the route in the `index.jsx` similar to what we are doing with the CAN pages since hackerz gonna hack.

good point! plz re-review.

fpigeonjr:
@jonnalley pls add the same logix to the portfolio paths in the `index.jsx` to not allow users to bypass the dropdown and go directly to the `/portfolios` url

jonnalley:
@fpigeonjr I've done that but it seems that's making the tests angrier? or maybe GH is broken.   Personally, I think the risk profile is lessened since this is just a temporary cosmetic / UI thing we want to hide from plain sight that we'll revert later and if someone is crafting manual URLs, then more power to them. I'll look at the tests again Friday maybe

fpigeonjr (approved):
W

rajohnson90 (approved):

johndeange (approved):

jonnalley:
hold on merging, want to use for an experiment

PR #3449, opened 2025-02-10
fix: fixes logout bug

### LINKED ISSUE #3436: Idle logout not working in some scenarios
## Expected Behavior
As on other pages within OPS, users should be gracefully logged out after 28 minutes of idleness in their session, before their JWT expires.

## Current Behavior
Instead, what the user experiences is a page that says "Loading CAN..." and in the console and backend logs we see constant requests to the backend for /cans URLs and 401s being returned as the JWT is now invalid and can't be read.

## Possible Cause
unsure

## Steps to Reproduce
1. Navigate to a CAN's detail page
2. Stay on the details tab or click on either of the other two tabs, doesn't matter--have tested with all tabs
3. wait for the configured session and JWT expiration timing and observe console and backend logs
4.

## Context
* saw this in staging and in localdev
* Confirmed the issue isn't present when viewing a list of CANs or a list of agreements
* For the record, @npage123 actually discovered this issue and should get full credit for it. 

## Possible Implementation

around 12:30-12:32 ET on 2/7, 2/8 at 03:03-4 ET

fpigeonjr:
## What changed

- fixes #3436 

## How to test

1. goto /cans/504
2. wait until logout (can be hacked `backend/ops_api/ops/environment/default_settings.py`)
3. will redirect to login
4. 💰 profit


fpigeonjr:
#3451 is a more elegant solution

PR #3546, opened 2025-02-28
feat: improve authentication and login ux

### LINKED ISSUE #2743: Long delay to get to home page after AMS auth
When logging in via AMS (with PIV), after the authentication is complete, there is what appears to be a 1-5 second delay before the home page displays

## Expected Behavior
The delay between completing the authentication process and getting redirected to the correct OPS environment landing page should be more reasonable.

## Current Behavior
The homepage doesn't render for about 1-5 seconds after the AMS authentication is seemingly complete. Instead, the user is taken back to the same screen they saw before auth to AMS ever occurred and they think they need to repeat the login action again--causing further issues due to only active session being permitted.

## Possible Cause
Waiting on AMS for the last round trip call being made without any visual indication of the user to wait for the authn process to complete.

## Steps to Reproduce
1. From GFE, navigate to the Azure dev, staging, or prod environment URLs.
2. Choose the option to log in with HHS AMS presuming you have a valid user created and ready for you in the DB.
3. Observe the delay before the OPS home page is returned. 
4.

See the recording of the 02/20/2025 meeting starting at 28:33 for the experience captured in production

## Context
This is not the same as the ACA cold start delay, which is far longer and is seen on the first login or real application activity following a new deployment. 

## Possible Implementation
Display a spinner or separate splash page while we wait for the token to come back from AMS. We would also want some kind of safety condition to kick in though to avoid a scenario of the user getting stuck on the spinner forever if something did go wrong.

See 

Possible UI improvements to not disorient users:
* https://github.com/davidhu2000/react-spinners
* https://www.davidhu.io/react-spinners/storybook/?path=/docs/pacmanloader--docs

fpigeonjr:
## What changed

- fixes #2743 

## How to test

1. login in prod
2. see pacman spinner
3. 💰 profit

or locally:
1. update the refresh token logic and update the `backend/ops_api/ops/environment/default_settings.py`

```py
JWT_ACCESS_TOKEN_EXPIRES = timedelta(minutes=3)  
JWT_REFRESH_TOKEN_EXPIRES = timedelta(hours=12) # leave setting
USER_SESSION_EXPIRATION = timedelta(minutes=1)
```

1. also need to make sure that fakeAuth is excluded from idle check in the `/backend/ops_api/ops/__init__.py`

```py
# if not is_unit_test() and not is_fake_user(app, current_user):
if not is_unit_test():
```


frontend/src/components/Auth/MultiAuthSection.jsx
```diff
@@ -111,31 +139,34 @@ const MultiAuthSection = () => {
...
                     </p>
                 )}
                 <p>
                     <button
                         className="usa-button usa-button--outline width-full"
                         onClick={() => handleSSOLogin("hhsams")}
+                        disabled={isLoading}
                     >
-                        Sign in with HHS AMS
+                        {isLoading ? "Signing in..." : "Sign in with HHS AMS"}
```
jonnalley, line 159: @fpigeonjr 🏆 🎸

jonnalley:
I'll defer to others to more thoroughly review but there's some really great stuff in here.

frontend/src/api/opsAuthAPI.js
```diff
@@ -27,8 +28,48 @@ export const opsAuthApi = createApi({
...
+                try {
+                    await queryFulfilled;
+                    // Dispatch logout action to clear auth state
+                    dispatch(logout());
+                } catch (err) {
+                    console.error("Error during logout:", err);
+                }
+            }
+        }),
+        refreshToken: builder.mutation({
```
johndeange, line 55: Where is this being invoked from now?

rajohnson90 (approved):

frontend/src/api/opsAuthAPI.js
fpigeonjr: invocations are in the `MultiAuthSection` like before but now it has been refactored to use RTK Query.

weimiao67 (approved):

frontend/src/api/opsAuthAPI.js
johndeange: I'm not seeing it in there for some reason, i.e. where `useRefreshTokenMutation` is used?

fpigeonjr: added the refresh token in the `AuthSection` and tested locally and seems to work. Logged in and sat idle for a few minutes and logs out and redirects to login page.

frontend/src/components/Auth/AuthSection.jsx
```diff
@@ -1,26 +1,72 @@
...
-    const activeUser = useSelector((state) => state.auth.activeUser);
+    /** @type {boolean} */
+    const isLoggedIn = useSelector((state) => state.auth?.isLoggedIn);
+    /** @type {Object|null} */
+    const activeUser = useSelector((state) => state.auth?.activeUser);
     const dispatch = useDispatch();
     const navigate = useNavigate();
+    const [loginMutation] = useLoginMutation();
+    const [logoutMutation] = useLogoutMutation();
+    const [refreshTokenMutation] = useRefreshTokenMutation();
```
fpigeonjr, line 26: 👀 @johndeange

johndeange (changes requested):
I think it's close but needs a little more work.  The refresh functionality isn't working for me and the UI is no longer polling the `/notifications` API when I tested it.  What should happen is when you set `JWT_ACCESS_TOKEN_EXPIRES = timedelta(minutes=1)` (leaving the rest the same) - this will force the UI to attempt to refresh every minute:

1. every minute hit `/notifications` API
2. `/notifications` returns a 401 because the access token is expired
3. The UI handles the 401 by hitting `/refresh` API
4. If the `/refresh` API returns a 200 it returns a new access and refresh token
5. If the `/refresh` API returns a 401 the user should be logged out

johndeange (approved):

PR #3573, opened 2025-03-06
feat: adds helper function based on papaParse

fpigeonjr:
adds the ability to handle paginated data from table 🫠

### DIFF
--- a/frontend/src/components/CANs/CANTable/CANTable.jsx
+++ b/frontend/src/components/CANs/CANTable/CANTable.jsx
@@ -28,6 +28,16 @@ const CANTable = ({ cans, fiscalYear }) => {
 
     const handleExport = async () => {
         try {
+            // table row headers
+            const TABLE_HEADERS = [
+                "CAN",
+                "Portfolio",
+                "Active Period",
+                "Obligate By",
+                "FY Budget",
+                "Funding Received",
+                "Available Budget"
+            ];
             // Get funding data for each CAN individually
             const fundingPromises = cans.map((can) =>
                 trigger({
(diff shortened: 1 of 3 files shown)


frontend/src/components/CANs/CANTable/CANTable.jsx
```diff
@@ -1,72 +1,82 @@
...
+                    }) ?? "TBD",
+                    fundingMap[can.id]?.received_funding?.toLocaleString("en-US", {
+                        style: "currency",
+                        currency: "USD"
+                    }) ?? "TBD",
+                    fundingMap[can.id]?.available_funding?.toLocaleString("en-US", {
+                        style: "currency",
+                        currency: "USD"
+                    }) ?? "TBD"
+                ],
```
weimiao67, line 74: brilliant idea!

weimiao67 (approved):

frontend/src/utils/tableExport.js
```diff
@@ -0,0 +1,38 @@
...
+
+    // Get headers from th elements
+    const headers = Array.from(tableRef.current.querySelectorAll("thead th")).map((header) => {
+        const clone = header.cloneNode(true);
+        clone.querySelectorAll(".usa-tooltip__body").forEach(
+            /** @param {HTMLElement} tooltip */
+            (tooltip) => tooltip.remove()
+        );
+        return clone.textContent?.trim();
+    });
```
weimiao67, line 22: if we want to add extra columns, how to add column name to headers?

fpigeonjr: good point, we'll probably want to pass in the table headings as a parameter.

PR #3807, opened 2025-04-28
fix: update agreement type label to AA for non contract agreement

### LINKED ISSUE #3778: Agreements allowing edit capability from list where otherwise prohibited
## Expected Behavior
There should either be no edit link in the agreements list for users any longer ( #1859) or there should be improved logic on whether that button is enabled to match the logic used on whether the edit link is displayed on an agreement's details.  In other words, if I'm not supposed to be able to, I shouldn't be able to edit the agreement no matter which method I try.

Also expected is that if the Agreement type label on the agreement's details is going to say "Contract", then that should not contradict the banner displayed for ContractAgreements with a procurement shop chosen and thusly prevented from editing.  Either we should allow editing of ContractAgreements with a chosen procurement shop or we should call them something other than a "Contract" or otherwise let the user know that's why it's not editable. 

## Current Behavior
After recent changes, the edit link for an agreement that's present in the list of all agreements will allow a qualified user (COR/ACOR/team member) to click the pencil icon to edit an agreement even though if that user browses to that agreement's details (without trying to edit), there is no edit link visible in the UI.

## Possible Cause
not applying changes to extra places ?  

## Steps to Reproduce
1. Browse to the agreements list as a user that's a COR/ACOR/team member on a given contract but that's not editable due to either procurement shop being chosen, agreement type, awarded state of agreement, etc...
2. Notice the active edit link for that contract

See video  below

## Context
#3757 removed the same links in the BL list. maybe the same should be done for Agreements.  Or fix the logic on whether the edit button is enabled.  

Found this bug at the same time as #3779 

## Possible Implementation
See expected behavior section above. See also #1859

https://github.com/user-attachments/assets/044e83f9-b6b9-493e-b2ba-acd995020b88

weimiao67:
## What changed

> This PR updates the agreement type label logic so that non-GCS contract agreements are shown as "AA" rather than "Contract". The changes update the helper function used for label generation, adjust the details view to use the new helper, and update tests (unit, integration, and e2e) to verify the new behavior.
>
> Updated AgreementDetailsView to use getAgreementType from agreement.helpers.
Revised tests in AgreementDetails.test.js and agreementDetails.cy.js to reflect the "AA" label when applicable.
Updated AgreementTableRow and the agreement.helpers.js to centralize the agreement type logic.

Update the agreement type label to AA if a contract with procurement shop that is not GCS. fixed #3778 AC 2

## Issue

#3778 

## How to test

1. all e2e test pass
2. go to agreement list and check AA agreement type
3. click on an AA type agreement, the agreement type should AA on agreement details page.

### list view
image

###  detail view
image

### DIFF
--- a/frontend/cypress/e2e/agreementDetails.cy.js
+++ b/frontend/cypress/e2e/agreementDetails.cy.js
@@ -54,7 +54,7 @@ it("Non contract type agreement loads with details", () => {
     cy.get("h2").first().contains("Support Project #1");
     cy.get("h2").eq(1).contains("Agreement Details");
     cy.get('[data-cy="details-right-col"] > :nth-child(1) > :nth-child(1)').contains("Agreement Type");
-    cy.get('[data-cy="details-right-col"] > :nth-child(1) > :nth-child(2) > .font-12px').contains("Contract");
+    cy.get('[data-cy="details-right-col"] > :nth-child(1) > :nth-child(2) > .font-12px').contains("AA");
     cy.get('[data-cy="details-right-col"] > :nth-child(2) > :nth-child(1)').contains("COR");
     cy.get("span").contains("Amelia Popham");
     cy.get('[data-cy="details-right-col"] > :nth-child(2) > :nth-child(2)').contains("Alternate COR");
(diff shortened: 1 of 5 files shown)


fpigeonjr (approved):
🏆 great job team

kimschulke:
I think we should write out the full agreement type on the agreement details page: Assisted Acquisition (AA). The acronym by itself okay in the list pages (AA).

maiyerlee (approved):

Santi-3rd (approved):

PR #3893, opened 2025-05-09
Master spreadsheet enhancement 1

johndeange:
## What changed

- Move the `--first-run` check to the top of the function.
- If `SYS_BUDGET_ID` is set but no BLI is found - skip it
- Use more robust regex for currency parsing

## How to test

Automated tests pass.


maiyerlee (approved):

backend/data_tools/src/common/utils.py
```diff
@@ -93,3 +92,20 @@ def get_bli_class_from_type(
...
+) -> float | None:
+    """
+    Converts a string representation of a budget amount to a float.
+    """
+    if not budget_amount:
+        return None
+
+    try:
+        # Remove $, commas, spaces, and negative signs
+        cleaned_amount = re.sub(r"[$,\s-]", "", budget_amount)
```
weimiao67, line 108: removing the negative sings will change the amount value completely, is it a desired scenario?

fpigeonjr (approved):
🧪 tests pass

backend/data_tools/src/common/utils.py
johndeange: I believe this is OK - we currently have validation in the system that prevents negative amounts.

Issue #2167, opened 2024-04-08
Budget Change: Edit Procurement Shop (including prompt for send to approval)


jonnalley:
### User Story
As a COR, I need to edit the procurement shop on an agreement with no BLs in a status higher than Planned, and know that it requires an approval, so that I understand how my edit will impact the overall budget

### Preconditions
* A contract exists with at least one budget line in `Planned` status but no budget lines in any status higher than that.

### Acceptance Criteria
- [x] On a given existing contract, authorized users are able to select a different procurement shop from the one currently selected and the changes are successfully persisted. Authorized users would be limited to Project officer, named team members, team leader associated with any CANs associated with the BLs of the agreement, Division Director linked to associated CANs, budget team, and System Admin(s).
- [ ] Changing the procurement shop should not be allowed if the contract has any BLs in `Executing` or higher status unless the user is a System Admin
- [x] Any BLs on the agreement in `Planned` status should change to In Review and start an approval process to get approved due to this financial change. 
  - [x] In this scenario where the agreement has Planned BLs, the procurement shop cannot be altered again until the active request to alter the procurement shop has been reviewed by an approver.
- [x] The user should be aware that making the procurement shop change will require re-approval of any `Planned` BLs
- [x] Validation rules would still need to be enforced on contract metadata such that required attributes based on the associated BL status can't be emptied out or otherwise made invalid
- [x] Upon successful committal of the pending procurement shop change, if a given contract has one or more BLs in `Planned` status such that changing the Procurement shop requires approval, the stored fee amount should not change at this time
  - [x] A contract with only draft BLs should have all the fees updated based on the procurement shop change.
- [x] Upon successful committal of the pending procurement shop change, relevant entries should be recorded in the event/history table(s) of who made what changes when and this event should be properly rendered  in the history of changes to the agreement. 

-----------------------------------------

### Test
- [x] Upon submitting the new procurement shop for approval, the agreement should still have the original procurement shop in place/rendered until it's approved. 
- [x] The user can't submit Yet Another procurement shop change while a current/previous change in the procurement shop is awaiting review.
- [x] If all BLs are in draft state, no approval would be required anyway for the BLs or contract metadata change. 
- [x] If at least one BL is in planned state, approval for the proc shop change would be required.
- [x] Even if approval is required, original fees shown pre-change would still persist until approved.

#### UX
* designs are already done as part of #1964

#### Dev

* essentially encompassing what #1473 was intended to do.

##### UI
- [x] #3869
[...]

PR #4006, opened 2025-06-10
feat: trigger approval process when procurement shop changes

maiyerlee:
## What changed
Refactored the Change Request (CR) flow and helper functions, including notifications for reviewers and submitters. A CR is initiated when the procurement shop on an Agreement changes, contingent upon the Budget Line Item status, and relevant CR OPS events are generated. 

## Issue

[API: Approval process when Procurement Shop is updated #3934](https://github.com/HHS/OPRE-OPS/issues/3934)

## How to test

🚦All green — tests passed!🟢 

## Screenshots

N/A

## Links
N/A


maiyerlee:
Business rule: "Agreements are associated with Divisions and Portfolios through CANs utilized within each budget line"

Reminder to handle `managing_divison_id` on AgreementChangeRequest (if needed) since an Agreement can be apart of multiple Divisions (unlike BLIs).

fpigeonjr:
🍖 beefy PR
image

fpigeonjr:
issue was resolved with a Docker flush
image

fpigeonjr:
q: did agreement get a new property `in_review`?
image

backend/ops_api/ops/resources/change_requests.py
```diff
@@ -102,64 +23,51 @@ def build_change_request_response(change_request: ChangeRequest):
...
+                return make_response_with_headers({"error": "change_request_id is required"}, 400)
+
+            service = ChangeRequestService(current_app.db_session)
+
+            try:
+                change_request, _ = service.update(change_request_id, request_json)
+                meta.metadata.update({"Updated Change Request": change_request.to_dict()})
+                return make_response_with_headers(change_request.to_dict(), 200)
+            except AuthorizationError:
+                return make_response_with_headers({"error": "Forbidden"}, 403)
```
johndeange, line 71: should not need to `return make_response_with_headers` in a number of places here - you can just raise the exception here and `error_handlers.py` with catch and handle it

johndeange (approved):

backend/ops_api/ops/services/agreements.py
```diff
@@ -181,64 +158,67 @@ def get_list(self, data: dict | None) -> tuple[list[Agreement], dict | None]:
...
+            bli_statuses.index(bli.status) == bli_statuses.index(BudgetLineItemStatus.DRAFT)
+            for bli in agreement.budget_line_items
+        ):
+            agreement.awarding_entity_id = new_value
+            self._update_draft_blis_proc_shop_fees(agreement)
+            return None
+
+        # Create a change request if all BLIs are either DRAFT or PLANNED
+        if any(
+            bli_statuses.index(bli.status) <= bli_statuses.index(BudgetLineItemStatus.PLANNED)
```
weimiao67, line 190: only PLANNED bli status requires change request, so the condition here should be == instead of <=

maiyerlee: I think the idea is that if there’s at least one PLANNED BLI, a change request is needed — that’s why it uses <=. But I’ll check with the team to be sure.

maiyerlee: One of the test scenarios in the parent task (https://github.com/HHS/OPRE-OPS/issues/2167) states: "_If at least one BL is in planned state, approval for the proc shop change would be required._" That seems to align with the current logic using <=.

weimiao67: Yes, that's correct, we need to check if there's at least one PLANNED BLI. And in the code, the `if any` condition will return true if there is only a DRAFT BLI.

maiyerlee: Ohh I see — thanks for clarifying! I’ll take another look at the condition (and add unit tests for this).

PR #4112, opened 2025-07-16
feat: include fees in the budget line item response

maiyerlee:
## What changed

Exposed the fees attribute in the Budget Line Item response.

## Issue

Related to https://github.com/HHS/OPRE-OPS/issues/2167

## How to test

✅ Backend unit tests pass successfully.

## Screenshots

N.A.

## Links

N.A.


backend/ops_api/ops/schemas/budget_line_items.py
```diff
@@ -373,6 +373,7 @@ class Meta:
     procurement_shop_fee = fields.Nested(
         "ops_api.ops.schemas.procurement_shops.ProcurementShopFeeSchema", required=True, allow_none=True
     )
+    fees = fields.Decimal(as_string=True, required=True)
```
maiyerlee, line 376: I am open to any suggestions here.

I'm using `as_string=True` so that the `Decimal` field is serialized into a string — to avoid precision loss and ensure safe handling of financial numbers in formats like JSON, which don't support `Decimal` natively.

Interesting thread on StackOverflow on topic: [here](https://stackoverflow.com/questions/35709595/why-would-you-use-a-string-in-json-to-represent-a-decimal-number).

fpigeonjr (approved):
🧪 tests pass, will defer to ~~CoPilot~~ others

fpigeonjr:
💡 we can update the BudgetLineTypes TS file to add the new prop

backend/ops_api/tests/ops/budget_line_items/test_budget_line_item.py
```diff
@@ -1536,3 +1536,49 @@ def test_get_budget_line_items_filter_options(system_owner_auth_client):
...
+def test_budget_line_item_fees_is_zero_when_proc_fee_is_null(auth_client, test_bli_new):
+    test_bli_new.proc_shop_fee_percentage = None
+    test_bli_new.amount = 100.0
+    auth_client.application.db_session.commit()
+
+    assert test_bli_new.fees == Decimal("0.00")
+
+    response = auth_client.get(f"/api/v1/budget-line-items/{test_bli_new.id}")
+    assert response.status_code == 200
+    assert response.json["fees"] == "0"
```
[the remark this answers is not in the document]
maiyerlee: Copilot is wrong here.

If either `proc_shop_fee_percentage` or `amount` is falsy (e.g., None or 0), the method explicitly returns an int 0, not a Decimal or a formatted float like "0.00". [(source)](https://github.com/HHS/OPRE-OPS/blob/4fa92a02af5033f7c8c944dcdbc525592e56137f/backend/models/budget_line_items.py#L148)

rajohnson90 (approved):
Your choice to do decimal as string to avoid precision loss makes sense to me at least.

backend/ops_api/ops/schemas/budget_line_items.py
johndeange: I think we typically use `fields.Float` (that would be my preference so that we preserve the general type) but String would be OK.

maiyerlee:
Setting Fees type to number/float for consistency with other funding-related data.

Issue #5052, opened 2026-02-11
Fee percentages showing incorrectly on a newly-created budget line


jonnalley:
## Expected Behavior
Fee percentages should be displaying correctly consistently both in the backend database and in the frontend

## Current Behavior
When a user created a new budget line, the fee **amount** appears to be calculating and rendering correctly in various places in the UI. But when accessing the fees amount in the accordion of the budget lines listing on the spending tab of viewing a portfolio, the fee percentage is showing as 100 times the correct amount (i.e. 300% instead of 3%).  Other places that render a percent in the UI appear to be doing so correctly. 

## Possible Cause
I think in the accordion mentioned above, it may be looking at the BL's `proc_shop_fee_percentage` and I'm not sure if it should or not. But if so, there's some inconsistency amid various tuples. Some tuples show 0.03 and newly created BLs show 3.0

## Steps to Reproduce
1. Create a new BL on an agreement with a procurement shop
2. Note it's `proc_shop_fee_percentage` perhaps as compared to an older one
3. Note the fee amount, which should be correct and compare with fee and amount shown in an export, in the agreement's details, and in the budget lines list. 
4.

## Context
This was found in the production environment when working with an OPRE user. Specifically found on the ADR portfolio.  Recommend looking at BL 1330, and two newly-created BLs: 16331 and 16332.

PR #5053, opened 2026-02-12
fix: fix fee percentages display bug

weimiao67:
## What changed

Fix the bug that the procurement fee percentage is displaying incorrectly in the budget line table under the portfolio spending tab.
 - Nests procurement_shop (with current_fee) in the budget line item responses, eliminating a separate API call to get the current procurement shop fee percentage.
 - Update frontend fee percentage calculation logic according to the API updates. 

## Issue

#5052 

## How to test

1. Create a new BL on an agreement with a procurement shop
2. Check the BL's fee percentage is correct under Portfolio spending, Agreement SCs&Budget Lines, and the Budget Line Item table.

## Screenshots

Screenshot 2026-02-12 at 12 53 45 PM


weimiao67:
> Non-blocking concern [P3] Fee calculation logic is now duplicated and can drift File: BLIsFYSummaryCard.helpers.js (line 10) getProcShopFeePercentage re-implements fallback logic instead of using the shared calculateProcShopFeePercentage helper. This creates a maintenance risk: the shared helper now falls back to agreement.procurement_shop.current_fee.fee, but this local copy does not. Recommendation: reuse calculateProcShopFeePercentage here to keep behavior consistent across views.

The BLIsByFYSummaryCard.jsx is a special case and is used in multiple components. In the creating agreement wizard, before we click the final create button, the agreement and budget line items are only stored in React state, the budget line does not have any procurement shop properties. That's why I created a separate function to handle the fee percentage.

Issue #2728, opened 2024-08-27
Read-Only role


jonnalley:
### User Story
As an OPS User Admin, I want to assign a user to a read-only role so that proper least-user privilege practices can be maintained for users that have no business need to create objects in OPS

### Acceptance Criteria
- [ ] A read-only user role should exist
- [ ] This read-only user role should not have any backend nor frontend permissions to create new or update any existing data structures. A user with **only** the read-only role ought not even see options to create BLs or agreements, for example.
  - [ ] with the exception that users with this role can acknowledge and mark their own notifications as seen/read
- [ ] A user with this read-only role ought not be included in any list of potential team members
- [ ] A User Admin should be able to assign or remove a read-only role to a user
- [ ] Users with the read-only role and any other role would thusly have the permissions of the sum or more-permissive of the set of roles.

-----------------------------------------

#### Dev

* Add role to test data set
* Add to initial production data set
* Add checks to frontend interfaces for creating data.

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] New design/tech debt documented (if applicable)
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

### Additional Context & Resources
* There's already a `read-only-user` branch started for this
* We were told on 3/18/2025 of the requirement for this to exist before the project support matrix team could be onboarded and most recently reiterated on 3/20.

jonnalley:
@rajohnson90 Sheila gave us the directive Wed. afternoon to proceed ahead fully with this story as originally written and intended here. So feel free to run with the frontend as well.

kimschulke:
@rajohnson90 UXers discussed this a little today and I think the main UX/UI changes are:
- Remove create from the main nav
- Do not display edit, delete, request (status change) or send icons/buttons anywhere (this includes agreement details page, project details page, hover actions on agreement list and budget lines list)

I think thats it, but we can double check in staging when time allows 😄

jonnalley:
reviewed and verified most AC are met, awaiting remaining UX review items and to test final AC

jonnalley:
@kimschulke @andressacurvelo @npage123 not sure if you all have already reviewed this or not

npage123:
[...]

PR #5600, opened 2026-04-29
feat(users): add read-only role and filter read-only users from user list

rajohnson90:
## What changed

This PR adds a new READ_ONLY role to the system and implements filtering logic to exclude users with this role from the user list API response. Important note: This doesn't close the task OPS-2728 yet because we have to choose an implementation for the frontend still. 

**backend/data_tools/data/user_data.json5**
- Added READ_ONLY role (ID 8) with GET-only permissions for all major resources (agreements, CANs, projects, portfolios, users, history, workflows, change requests, documents)
- Created test user "Read-Only Randall" (ID 529) assigned to the READ_ONLY role

**backend/ops_api/ops/services/users.py**
- Updated `get_users()` service to filter out users with READ_ONLY role from the response
- Added business rule documentation explaining the filtering behavior

**backend/ops_api/tests/ops/users/test_users_get.py**
- Added test assertions to verify READ_ONLY users exist in the database but are excluded from the API response

## Issue

https://github.com/HHS/OPRE-OPS/issues/2728

## How to test

**Run automated tests:**
```bash
cd backend/ops_api
pipenv run pytest tests/ops/users/test_users_get.py::test_get_all_users -v
```

**Manual verification:**
1. Start the application with Docker: `docker compose up --build`
2. Verify the READ_ONLY role exists in the database:
   ```bash
   docker compose exec db psql -U ops -d postgres -c "SELECT id, name FROM ops.role WHERE name = 'READ_ONLY';"
   ```
3. Verify Read-Only Randall exists in the database:
   ```bash
   docker compose exec db psql -U ops -d postgres -c "SELECT id, first_name, last_name, email FROM ops.ops_user WHERE id = 529;"
   ```
4. Call the users API endpoint and verify user ID 529 is not in the response:
   ```bash
   curl http://localhost:8080/api/v1/users/ | jq '.[] | select(.id == 529)'
   ```
   (Should return empty result)

## A11y impact

## Screenshots

N/A — backend changes only

## Links

N/A


backend/ops_api/ops/services/users.py
```diff
@@ -89,7 +92,10 @@ def get_users(session: Session, **kwargs) -> list[User]:
 
     users = session.execute(stmt).scalars().all()
 
-    return list(users)
+    # Filter out users with READ_ONLY role
+    filtered_users = [user for user in users if not any(role.name == "READ_ONLY" for role in user.roles)]
+
+    return filtered_users
```
jonnalley: I like it, but I am curious if this change will also affect the list of users returned that a user admin can pick from to be editing users?? If so, this might mean once a user admin could make a user read-only and then never be able to access them again to change that.

rajohnson90: It probably would do that. I wasn't really planning to solve this sort of problem until we went with a front end solution. My plan was to have something like an 'include_read_only' flag on the request that will include all users. I could make that change to the frontend now at least.

josbell (changes requested):
I had two independent agents review this PR, followed by a deep investigation into the concerns raised. Here are the findings:

  🚨 BLOCKING ISSUE: Role ID Mismatch

  Critical data integrity bug - will cause FK constraint violations during data import.

  Problem:
  - SQL files insert READ_ONLY as role ID 7 (backend/data_tools/initial_data/003-role.sql line 7, 004-role_version.sql line 7)
  - JSON5 file assigns user 529 to role ID 8 (backend/data_tools/data/user_data.json5 line 695)

  Fix Required:
  Change both SQL files to use ID 8 instead of ID 7:
  INSERT INTO ops.role (id, name, permissions, ...) VALUES (8, 'READ_ONLY', ...

  The existing role sequence is 1-6 (SYSTEM_OWNER through PROCUREMENT_TEAM), so READ_ONLY should be ID 8, not 7.

  ---
  ⚠️  Non-Blocking Observations

  1. User Management Visibility (Can address in follow-up)

  READ_ONLY users are filtered from GET /users list endpoint, making them "invisible" in the UI. However:
  - They're still accessible via GET /users/{id} ✅
  - They can still be updated via PUT/PATCH /users/{id} ✅
  - This is a UX gap, not a data integrity issue

  Recommendation: Implement the include_read_only parameter mentioned in the PR description in a follow-up PR. Consider adding it to the QueryParameters schema to allow admins to toggle READ_ONLY user visibility.

  2. API Design (Working as intended)

  GET /users filters READ_ONLY users, but GET /users/{id} does not. This is intentional and correct REST API design:
  - List endpoints apply business logic filtering
  - Item endpoints return the resource if it exists and user has permission

  Recommendation: Consider adding a test (test_get_read_only_user_by_id) to document that direct access works even though list filtering is applied.

  ---
  ✅ Action Items

  Must fix before merge:
  - Change role ID from 7 to 8 in 003-role.sql and 004-role_version.sql
                                                                                                                                                                                                                                                                                                                     
  Recommended before merge:
  - Test database import after fix: docker compose down -v && docker compose up db data-import --build                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                     
  Follow-up PR (not blocking):                                                                                                                                                                                                                                                                                       
  - Implement include_read_only query parameter for better admin UX                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                     
  ---                                                                                                                                                                                                                                                                                                                
  Summary: One critical bug to fix (2-line change), then good to merge. The other concerns are either design-as-intended or nice-to-have enhancements for follow-up work.

backend/ops_api/ops/services/users.py
rajohnson90: This has been resolved. Visually verified that the user shows up in the admin panel listing

rajohnson90:
The JSON role's listing is correct. Read-Only should be id 8, not 7. I removed the SQL code because as I remember it, this code isn't used anymore. It was initially used to seed databases early on but has no effect anymore

rajohnson90:
The medium importance items have been addressed. The read-only users are no longer filtered by default and are instead filtered by query param. The front end functionality that shouldn't let a read-only user be added to as a team member or COR are filtering the read-only users. However, the admin console which can assign and unassign roles does not filter out the read only user. This is an intentional design choice.

johndeange (changes requested):
## Code Review

### Issues

#### 1. Potential `NameError` in `get_users` service (Bug)

In `backend/ops_api/ops/services/users.py`, the `exclude_read_only` variable is only assigned inside the `for key, value in kwargs.items()` loop, but referenced outside it. While the marshmallow schema sets `load_default=False`, any direct caller of `get_users()` that doesn't go through the schema will hit a `NameError`. Additionally, if `exclude_read_only` isn't popped from kwargs, it will fall into the `else` branch which tries `getattr(User, "exclude_read_only")` and will fail.

Fix: initialize and remove it before the loop:
```python
exclude_read_only = kwargs.pop("exclude_read_only", False)
```

#### 2. Missing Alembic migration

The PR body references `backend/data_tools/initial_data/003-role.sql` and `004-role_version.sql` but these files are not in the diff. More importantly, an Alembic migration file should be generated so that the new READ_ONLY role is added to the staging and production databases. Without a migration, deployed environments won't have this role. Please generate an Alembic migration that inserts the READ_ONLY role (and its permissions) into the `role` and `role_version` tables.

#### 3. Filtering in Python vs. SQL (Suggestion)

The read-only filter currently loads all users into memory then filters in Python:
```python
filtered_users = [user for user in users if not any(role.name == "READ_ONLY" for role in user.roles)]
```

Consider filtering at the SQL level instead:
```python
if exclude_read_only:
    stmt = stmt.where(~User.roles.any(Role.name == "READ_ONLY"))
```

#### 4. OpenAPI changes seem out of scope

The `openapi.yml` diff adds 234 lines documenting POST/PUT/PATCH user endpoints and renames the path param `user_id` → `id`. These aren't related to the read-only role feature and make the PR harder to review. Consider splitting into a separate PR.

### What looks good

- The opt-in query parameter approach is the right design — admin views can still see all users
- Frontend correctly targets only user-assignment comboboxes
- Fake auth provider addition enables easy testing
- Test assertions verify the user exists in DB but is excluded from filtered responses

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rajohnson90:
Implemented John's suggestions. If you want to test the new script you have to delete the read only user and role from user_data.json5 file and then run the script in `data_tools/scripts/src/add_read_only_role/add_read_only_role.py --env dev` and verify the read only role is generated. In this case I removed the read-only user to remove a reference to a nonexistent role value.

josbell (approved):

johndeange (approved):
## Code Review

### Overview

This PR adds a `READ_ONLY` role with GET-only permissions and implements an `exclude_read_only` query parameter so the frontend can filter out read-only users from selection dropdowns (team member, project officer, procurement tracker). The opt-in filtering via query parameter is a good design choice.

---

### Issues

#### 1. Missing Alembic Migration (Non-blocking — can address in follow-up)

No Alembic migration in the PR. The standalone script at `backend/data_tools/src/add_read_only_role/add_read_only_role.py` has issues:
- Final log message says "Procurement tracker backfill complete" (copy-paste error)
- Doesn't check if the role already exists before inserting (docstring says it does, code doesn't)
- Passes a Python list to raw SQL `text()` INSERT — this won't serialize correctly to a PostgreSQL array

#### 2. Return Type Change in `get_users` (Minor)

The diff removes `list()` wrapper:
```python
- return list(users)
+ return users
```
Function signature declares `-> list[User]` but `scalars().all()` returns `Sequence[User]`. Consider keeping `list(users)` for type consistency or updating the annotation.

#### 3. OpenAPI Spec Scope Creep

The `openapi.yml` changes add full POST/PUT/PATCH operations, rename path parameters, and document all query params — significantly broader than what this PR needs. Consider splitting into a separate documentation PR.

---

### Suggestions

1. Add a test for the `exclude_read_only=False` (default) case to verify read-only users ARE included when not filtering.
2. Add a frontend test for `excludeReadOnlyUsers: true` in `UsersComboBox.test.jsx`.
3. Consider a backend constant for the `"READ_ONLY"` role name string to reduce duplication.

---

### Positives

- Clean opt-in filtering design — doesn't break existing behavior
- Frontend correctly applies filtering only in selection dropdowns, not admin views
- SQL-level filtering (`~User.roles.any(...)`) is efficient
- Fake auth provider updated for local dev testing
- Good test coverage of the core filtering logic

LGTM — core logic is solid and well-designed. The script issues and OpenAPI scope can be addressed in follow-ups.

Issue #5241, opened 2026-03-10
Filter list of projects


jonnalley:
### User Story
As an OPS user, I want to filter the list of all projects so that I can view and find only the specific  projects I desire

### Acceptance Criteria
- [ ] User can access filters for the list of projects including fiscal year, portfolio, project title or nickname, type, or agreement title/nickname
  - [ ] The filtering of fiscal years should work in concert with the component selecting the fiscal year that is used apart from the filters. 
- [ ] User can easily remove and reset all filters
- [ ] Each filter should be multi-select if/where possible
-----------------------------------------

#### UX
- [x] Draft design options for view a list of projects
- [x] Share design with OPS stakeholders for feedback
- [x] Refine design based on feedback
- [x] Share designs with OPRE users for feedback
- [x] Refine design based on feedback
- [x] Annotate design and move to ready for dev

#### Dev
- 

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] New design/tech debt documented (if applicable)
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

### Additional Context & Resources
* Related to #4210 
* [Designs in Figma](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=27966-34645&t=mzTQQ2SiaLiBnnGI-1)

jonnalley:
@kimschulke @andressacurvelo @npage123 This is now in dev and staging and is awaiting your review.

jonnalley:
@rajohnson90 please do review DoD at your convenience

npage123:
Tested in staging and all looks good except when selecting multiple FYs, the page level FY should read 'Multi' in all combinations of multiple year selections except when all are selected which should then read 'All' (which is currently working correctly). 

Also, if filters are selected, and the user selects a new FY via the page level dropdown, filters should reset.

kimschulke:
Adding a video to show some of the issues relating to filters. Ideally, we can get all the filters working exactly the same on every page. We also starting work on this in #5332 

https://github.com/user-attachments/assets/6c0c19a3-1116-4c9f-af29-8122e0f779f5

PR #5614, opened 2026-05-01
feat: add project filtering with fix for multi-filter queries

rajohnson90:
## What changed

This PR adds comprehensive project filtering to the Projects list page and fixes a critical backend bug that prevented multiple filters from working together correctly.

**Backend query fix** (`backend/ops_api/ops/services/projects.py`)

Replaced LEFT OUTER JOIN-based filtering with EXISTS subqueries in `_get_research_projects_query` and `_get_administrative_and_support_projects_query`. The original approach required a single joined row to satisfy all filter conditions simultaneously, which failed when:
- One agreement had budget line items in the target fiscal year
- A different agreement on the same project matched the agreement search term

The new approach uses independent EXISTS subqueries for each filter (portfolio, fiscal year, agreement search), allowing each filter to match different related records while still applying AND logic at the project level.

**Frontend filtering UI** (`frontend/src/`)

- `ProjectFilterButton/`: New modal-based filter component with multi-select dropdowns for fiscal year, portfolio, project title, project type, and agreement name
- `ProjectFilterTags/`: Displays active filters as removable tags with clear visual feedback
- `ProjectTypeComboBox/`: Reusable project type selector component
- Updated `ProjectsList.jsx` to integrate filtering UI and handle filter state

**API enhancement**

Agreement names in the `/projects/filters` endpoint now return `{id, name}` objects instead of plain strings, aligning with other filter options and enabling better frontend handling.

**Tests**

- 260 lines of Cypress E2E tests covering all filter combinations
- New backend test for combined fiscal year + agreement filtering scenario

## Issue

https://github.com/HHS/OPRE-OPS/issues/5241

## How to test

**Backend tests:**
```bash
cd backend/ops_api
pipenv run pytest tests/ops/project/test_project.py::test_agreement_and_fiscal_year_filter -xvs
pipenv run pytest tests/ops/project/test_project.py::TestProjectFilterOptions -xvs
```

**Frontend E2E tests:**
```bash
cd frontend
npm run cypress:open
# Run projectsList.cy.js - new tests starting at line 360
```

**Manual verification:**
1. Navigate to `/projects`
2. Click the "Filter" button
3. Select "FY 2044" from fiscal year dropdown
4. Select "AA #1: Fathers and Continuous Learning (FCL)" from agreement dropdown
5. Click "Apply"
6. ✅ "Annual Performance Plans and Reports" project should appear (previously returned no results)
7. Verify filter tags display: "FY 2044" and "AA #1: Fathers and Continuous Learning (FCL)"
8. Click the X on a filter tag to remove it
9. Click "Filter" → "Reset" to clear all filters
10. Test other filter combinations (portfolio, project type, project title)

## A11y impact

- [x] Color contrast verified for filter tags and buttons
- [x] Keyboard navigation tested in filter modal
- [ ] Screen reader announces filter application and removal
- [ ] Focus management when opening/closing modal

## Screenshots
The image below shows the UI element
Screenshot 2026-05-01 at 11 57 56 AM

## Links

**Related context:**
- Backend uses EXISTS subqueries pattern similar to Agreement filtering
- Frontend filter patterns follow existing CAN and Agreement list filter implementations
- OpenAPI schema updated to match filter response structure


josbell (approved):

frontend/src/pages/projects/list/ProjectsList.jsx
weimiao67, line 1: Bug: Pagination not reset when filters change

ProjectsList.jsx:69-71 — the useEffect that resets pagination only depends on sort/fiscal year.

filters is missing from the dependency array. If a user is on page 3 and applies a filter that returns fewer results, they'll see an empty page. filters needs to be added here.

frontend/src/pages/projects/list/ProjectFilterTags/ProjectFilterTags.hooks.js
```diff
@@ -0,0 +1,135 @@
...
+ * Removes a filter tag
+ * @param {Tag} tag - The tag to remove
+ * @param {function(function(Filters): Filters): void} setFilters - Function to update filters
+ */
+export const removeFilter = (tag, setFilters) => {
+    switch (tag.filter) {
+        case "fiscalYear":
+            setFilters((prevState) => ({
+                ...prevState,
+                fiscalYear: prevState.fiscalYear.filter((fiscalYear) => "FY " + fiscalYear.title !== tag.tagText)
```
weimiao67: Bug: Double "FY" prefix in filter tags

The tagText for fiscal year is "FY " + item.title. If the FiscalYearComboBox options already have titles like "FY 2044", the resulting tagText is "FY FY 2044".

rajohnson90: Great catch. I've fixed this in my latest push. 
Screenshot 2026-05-04 at 9 08 51 AM

johndeange (changes requested):
## Overview

Adds project filtering UI (modal + tags) and fixes a backend bug where combining fiscal year + agreement filters returned empty results. The EXISTS-subquery approach is sound and matches patterns used elsewhere in the codebase. Two blocking issues before merge.

---

## 🔴 Blocking

### 1. Broken assertion in new backend test
`backend/ops_api/tests/ops/project/test_project.py:360`

```python
assert response_research == 200
```

`response_research` is a `TestResponse` object, not an int — this comparison is always `False`. Should be:

```python
assert response_research.status_code == 200
```

This test is the whole motivation for the backend fix, so it needs to actually execute. Consider also adding a negative case (agreement matches but FY doesn't → empty result).

### 2. Frontend/backend contract mismatch on \`agreement_names\`

Backend now returns \`agreement_names: [{id, name}]\`, but \`frontend/src/api/opsAPI.js:486\` reads \`.title\`:

```javascript
queryParams.push(\`agreement_search=\${encodeURIComponent(agreement.title)}\`);
```

If \`AgreementNameComboBox\` transforms \`{id, name}\` → items with \`title\` before they land in \`agreementSearch\` state, this works — but the transform isn't visible in this diff and is easy to break. Please verify end-to-end and ideally add a test that asserts the produced query string. The mocked combobox in \`ProjectFilterButton.test.jsx\` uses \`{title: \"Agreement 1\"}\`, so the unit tests wouldn't catch a shape mismatch.

---

## Important

### Inconsistent shape for \`projectType\` filter
\`opsAPI.js:500\` sends \`type.name\`, but \`ProjectFilterTags.hooks.js:111\` removes by \`type.title\`, and tests use both \`{title: \"RESEARCH\"}\` and \`{name: PROJECT_TYPE_RESEARCH}\`. Pick one key (recommend \`name\` since it's an enum value) and apply everywhere.

### \`Modal.setAppElement(\"#root\")\` inside render
\`ProjectFilterButton.jsx:109\` — runs on every render. Move to module level or a \`useEffect\` with empty deps.

### \`useTagsList\` is more complex than needed
\`ProjectFilterTags.hooks.js\` has 5 \`useEffect\`s, a \`useCallback\`, and three near-duplicate branches that differ only by \`item.name\` vs \`item.title\`. This can collapse to a single \`useMemo\` derived from \`filters\` — no state, no effects.

```javascript
const tagsList = useMemo(() => {
    const map = { portfolio: 'name', fiscalYear: 'title', projectSearch: 'title', agreementSearch: 'title', projectType: 'title' };
    return Object.entries(map).flatMap(([key, prop]) =>
        (filters[key] ?? []).map(item => ({ tagText: item[prop], filter: key }))
    );
}, [filters]);
```

### \`useProjectFilterButton\` sync pattern
The 5 \`useEffect\`s that copy \`filters.X\` → local state have a subtle UX quirk: if the user opens the modal, makes selections, closes without Apply, the internal state persists on re-open. Confirm this matches the design.

---

## Minor

- \`console.warn\` in \`removeFilter\` default case — unreachable with current types, remove it.
- \`ProjectFilterTypes.d.ts\` defines \`Filters\` but \`ProjectFilterTags.hooks.js\` redefines it via JSDoc — consolidate.
- \`ProjectTitleComboBox.jsx:13-14\` JSDoc typo: \"IS the data for\" → \"Is the data for\".
- \`.distinct(ResearchProject.id)\` is kept but no longer needed now that outer joins are gone — can be removed.
- PR description's A11y checklist is all unchecked, and \"Screenshots\" section appears twice.
- Cypress \`cy.wait(1000)\` calls — prefer asserting on a changed-state condition rather than a fixed wait.
- \`selectinload\` chains are duplicated verbatim between the two query builders — candidate for a shared helper if we want to DRY further.

---

## Strengths

- EXISTS-subquery fix is correct and addresses a real user-facing bug.
- Good Cypress coverage across filter combinations.
- Thorough unit tests on the filter button modal.
- No injection concerns — all values go through SQLAlchemy parameter binding.

Requesting changes for the two blocking items; the rest are polish.

johndeange (approved):
## Re-review after "respond to pr comments" commit

Both blocking issues from my previous review are resolved:
- ✅ `assert response_research.status_code == 200` fixed, negative case added
- ✅ `Modal.setAppElement` moved to `useEffect`
- ✅ `useTagsList` collapsed to a single `useMemo` (much cleaner)
- ✅ `console.warn` removed, type defs consolidated

### Minor items for follow-up

1. **Pagination reset on filter change** — Wei flagged that the pagination `useEffect` doesn't depend on `filters`. Please confirm this is handled (here or follow-up).
2. **`agreement_search` data shape** — pipeline works (E2E proves it), but the `{id, name}` → `{id, title}` transform in `AgreementNameComboBox` isn't visible in this diff. A brief comment would help the next dev.
3. **`projectType` naming asymmetry** — `opsAPI.js` sends `type.id`, `removeFilter` filters by `type.title`. Works because both exist on each item, but naming is inconsistent across the filter lifecycle.

None of these are blocking. Approving — nice work on the EXISTS-subquery fix and the filter UI. 👍

Issue #5566, opened 2026-04-22
Unsaved Changes modal for Edit CAN workflow


kimschulke:
### Goals
* If a user is editing a CAN and attempts to navigate away without saving, we should display a confirmation modal just like we do for the edit agreement workflow

### Current state
Currently, a budget team user can click edit on a CAN, start making changes, and then click on a new page in the main nav and their changes will be lost.

### Tasks
- [x] TBD

### Additional Context
* might have missed this during our previous work on the edit/save improvements and navigate-away modal

### Resources
* TBD

kimschulke:
This has been resolved via below --

See https://github.com/HHS/OPRE-OPS/issues/5565#issuecomment-4379619614

Issue #5565, opened 2026-04-22
Stuck in CAN edit mode


kimschulke:
If you click on edit from the CAN Details tab, but then click on the CAN Funding tab, you get stuck in edit mode on the CAN Details even after you click save changes on the CAN Funding tab (video attached)

## Expected Behavior
If the user clicks edit mode, that mode should end/close after they click Save Changes (even across tabs)

## Current Behavior
Edit mode stays active even after the user clicks save changes

## Possible Cause
We never intended for edit mode to persist across tabs, but thats how its currently implemented.
If we leave it like that, we need to make sure edit mode closes after they click Save Changes.
If we change it so you need to click Edit on each Tab, then we dont need to worry about this.

## Steps to Reproduce
1. Sign in to staging as a Budget Team Member
2. Click on CANs
3. Click on CAN G995679
4. Click edit on CAN Details
5. Click on CAN Funding
6. Make changes to the CAN Funding
7. Click save changes
8. Click CAN Details

## Context
Not sure if this is a bug or tech debt

## Detailed Description
TBD

## Possible Implementation
TBD

kimschulke:
https://github.com/user-attachments/assets/74393645-fefc-45dc-937b-699f3cbae462

kimschulke:
Collaborated with @weimiao67 

Heres where we ended up
- if the user clicks edit mode, and makes a change, and then tries to navigate between tabs, we will trigger the unsaved changes modal (this is the same behavior for agreements)
- If the user clicks on edit mode, and does **not** make a change, and then tries to navigate between tabs, will just exit edit mode

This update might also fix GH #5566 🎉

kimschulke:
here is what happens on agreement-edit mode so now CANs-edit mode will match 🎉 

https://github.com/user-attachments/assets/0f562a16-6274-420c-8e0f-15cae5fa9e56

weimiao67:
@kimschulke the fix is ready for review and test on Dev and Staging. It also fixes #5566

kimschulke:
Thank you @weimiao67 ! This looks great! 🎉 I can close this one out along with #5566 

For others awareness, this now fixes:
-  if user clicks edit, makes a change, and navigates away to another tab or page ➡️ then unsaved changes modal will display 
-  if user clicks edit, **does not** make any changes, and navigates away to another tab or page ➡️  then edit mode gets exited automatically 

video attached

[...]

PR #5616, opened 2026-05-01
OPS-5565: add unsaved changes navigation blocking on CAN forms

weimiao67:
## What changed

- Add unsaved changes navigation blocking to CAN Detail and CAN Funding edit forms, matching the existing agreement edit form behavior
- When a user is in edit mode with unsaved changes and tries to navigate away (e.g., clicking a different tab), a "You have unsaved changes" modal appears with options to Save Changes, Leave without saving, or close the modal to stay
- When a user is in edit mode with no changes and navigates away, edit mode exits silently without a modal
- Add e2e test coverage for all unsaved changes navigation blocking scenarios

## Issue

#5565 
#5566 

## How to test

- Enter CAN detail edit mode, change nickname, click another tab → modal appears
- Enter CAN detail edit mode, make no changes, click another tab → navigates directly
- Enter CAN funding edit mode, change budget, click another tab → modal appears
- Enter CAN funding edit mode, make no changes, click another tab → navigates directly
- Modal "Save Changes" → saves and navigates
- Modal "Leave without saving" → discards and navigates
- Modal close/Escape → stays on page with changes intact

## A11y impact


johndeange (changes requested):
## Overview

Extracts the unsaved-changes navigation-blocking pattern from `AgreementEditForm.hooks.js` into a reusable `useNavigationBlocker` hook, then wires it into the CAN Detail and CAN Funding edit forms. Adds `SaveChangesAndExitModal` usage on both forms, plus unit tests for the hook and e2e coverage for all modal paths (save / leave / cancel / escape / no-changes).

## Strengths

- **Reusable hook.** `useNavigationBlocker` cleanly encapsulates blocker state, modal prop generation, and the `isCancelling` bypass.
- **Ref-based closures** (`saveChangesRef`, `onExitRef`, `blockerRef`) avoid stale-closure bugs when the effect depends only on `blocker.state`.
- **Solid test coverage.** Hook unit tests cover happy path, no-changes bypass, cancelling bypass, same-path navigation, save success, save failure, secondary action, close, and the known RR blocker transition error.
- **E2E matches the described flows** — each scenario in the PR description has a corresponding test.
- **Matches existing conventions** (React Router `useBlocker`, `SaveChangesAndExitModal`, Vest suite patterns).

## Issues and Suggestions

### 1. `saveChanges` in `CANDetailForm.hooks.js` swallows API errors (blocking)

```javascript
// CANDetailForm.hooks.js:45-57
const saveChanges = React.useCallback(async () => {
    const payload = { ... };
    await updateCan({ id: canId, data: payload });   // no .unwrap()
    setAlert({ type: "success", heading: "CAN Updated", ... });
}, [...]);
```

Without `.unwrap()`, RTK Query resolves the promise with `{ error }` on failure instead of throwing. That means:
- The success alert fires even when the API call failed.
- `onSaveError` in the blocker will **never** fire → error redirect is unreachable.
- Navigation proceeds despite a failed save.

Compare `CanFunding.hooks.js`, which correctly uses `.unwrap()`. Fix:

```javascript
await updateCan({ id: canId, data: payload }).unwrap();
```

This matters more now because the blocker contract expects `saveChanges` to throw on failure.

### 2. Hook order in `Can.hooks.js` is confusing

```javascript
// Can.hooks.js:17-31
const urlPathParams = useParams();
const location = useLocation();

React.useEffect(() => {
    setIsEditMode((prev) => ...);   // setIsEditMode declared 15 lines below
}, [location.pathname]);

// ...
const [isEditMode, setIsEditMode] = React.useState({ ... });
```

The effect references `setIsEditMode` before the `useState` call. It works (setters are stable, closure resolves after render), but it's surprising to read. Move the effect below `useState`.

### 3. `blocker.reset()` inside the effect uses the non-ref'd blocker

```javascript
// useNavigationBlocker.hooks.js:414
} catch (error) {
    ...
    blocker.reset();   // closure over the render-time blocker
}
// eslint-disable-next-line react-hooks/exhaustive-deps
}, [blocker.state]);
```

Everywhere else in this hook we correctly go through `blockerRef.current`. For consistency and safety against the blocker identity changing, prefer `blockerRef.current.reset()` here and in `closeModal`.

### 4. `isCancelling` has no reset path in `CANDetailForm`

`CanFunding.hooks.js` adds a `wasEditModeRef` effect that clears `isCancelling` on each re-entry to edit mode. `CANDetailForm.hooks.js` does not. This only works because `CANDetailForm` unmounts when edit mode toggles off (so fresh state on re-entry) — but that's an implicit assumption. Worth either documenting with a comment or mirroring the `wasEditModeRef` pattern for symmetry.

### 5. `AgreementEditForm.hooks.js` is not migrated to the new hook

~120 lines of near-identical blocker/modal plumbing still live in `AgreementEditForm.hooks.js` (including the same `proceedIfBlocked` and "Invalid blocker state transition" workaround). Migrating it in this PR (or a follow-up) would reduce duplication and make the shared hook the single source of truth. If out of scope, consider leaving a TODO.

### 6. No unit test for the new `Can.hooks.js` location effect

`Can.hooks.test.js` only adds a `useLocation` mock, no test for the edit-mode-reset behavior. An assertion that `isEditMode` clears when `location.pathname` changes would be easy to add and document the intent.

### 7. Minor: fragile error message matching

The string-based matching in `proceedIfBlocked` (`message.startsWith("Invalid blocker state transition")`) is tied to internal RR wording. A comment referencing the upstream issue would help future readers understand why this workaround exists.

## Risk Assessment

- **Behavior impact:** Medium — changes navigation semantics on CAN detail/funding pages. E2E coverage is good.
- **Regression risk for agreements flow:** Low — `AgreementEditForm` untouched.
- **Error-handling regression:** Low-medium — see issue #1; a failed CAN detail save now gives misleading feedback.

## Recommendation

Requesting changes — issue #1 (missing `.unwrap()`) should be fixed before merge. The rest are minor polish / follow-up items.

josbell (changes requested):
## Code Review - Request Changes

After thorough investigation, two **critical bugs** must be fixed before merge. These findings were validated by examining the actual implementation, comparing patterns across the codebase, and surveying all RTK Query mutation usage.

---

## 🚨 Critical Issue #1: Missing `.unwrap()` in CANDetailForm

**File:** `frontend/src/components/CANs/CANDetailForm/CANDetailForm.hooks.js`  
**Lines:** 48-51

### Current Code (INCORRECT):
```javascript
await updateCan({
    id: canId,
    data: payload
});
```

### Problem:
RTK Query mutations without `.unwrap()` **never throw on error**. They return `{data, error}` objects instead. This causes:

- ✗ **Data integrity bug**: Success alert displays even when API call fails
- ✗ **Silent error swallowing**: No error propagates to `useNavigationBlocker`
- ✗ **Incorrect state**: Form exits edit mode despite save failure
- ✗ **Contract violation**: `useNavigationBlocker` JSDoc explicitly requires `saveChanges` "should throw on failure" (line 7)

### Evidence:
- **CanFunding.hooks.js** correctly uses `.unwrap()` at lines 168, 173
- **Codebase standard**: Surveyed ALL hook files - 100% of RTK Query mutations use `.unwrap()`
- **30+ examples** across components/pages confirm this is a hard requirement

### Required Fix:
```javascript
await updateCan({
    id: canId,
    data: payload
}).unwrap();
```

---

## 🚨 Critical Issue #2: Missing Error Handling in handleSubmit

**File:** `frontend/src/components/CANs/CANDetailForm/CANDetailForm.hooks.js`  
**Lines:** 88-92

### Current Code (INCORRECT):
```javascript
const handleSubmit = async (e) => {
    e.preventDefault();
    await saveChanges();
    cleanUp();
};
```

### Problem:
Once `.unwrap()` is added (fixing Issue #1), errors will throw but won't be caught:

- ✗ Uncaught exception bubbles up
- ✗ `cleanUp()` doesn't execute on error
- ✗ Form left in inconsistent state
- ✗ No user-facing error message or recovery path

### Required Fix:
Match the **CanFunding.hooks.js** pattern (lines 287-303):

```javascript
const handleSubmit = async (e) => {
    e.preventDefault();
    try {
        await saveChanges();
        cleanUp();
    } catch (error) {
        console.error("Error Updating CAN", error);
        setAlert({
            type: "error",
            heading: "Error",
            message: "An error occurred while updating the CAN.",
            redirectUrl: "/error"
        });
    }
};
```

---

## ⚠️ Code Quality Issue: Hook Ordering in Can.hooks.js

**File:** `frontend/src/pages/cans/detail/Can.hooks.js`  
**Lines:** 20-27 vs 44-47

### Problem:
The `useEffect` at line 20 references `setIsEditMode` before its `useState` declaration at line 44 (24 lines later).

**This is functionally safe** (JavaScript hoisting + React closures), but:
- ✗ Violates React conventions (state should be declared before use)
- ✗ Not used elsewhere in codebase (surveyed 41 hook files)
- ✗ Makes code harder to maintain and reason about

### Recommended Fix:
Move the `useState` declarations (lines 44-47) to before the `useEffect` (line 20) for consistency with project patterns.

**Severity: Low** - Non-blocking, but should be fixed for code quality

---

## Positive Aspects

- ✅ Excellent E2E test coverage with comprehensive scenarios
- ✅ Clean hook abstraction with `useNavigationBlocker`
- ✅ CanFunding implementation follows correct patterns
- ✅ Good documentation and JSDoc comments

---

## Summary

**Two critical bugs must be fixed:**
1. Add `.unwrap()` to `updateCan()` mutation (line 48-51)
2. Add try/catch error handling to `handleSubmit()` (lines 88-92)

Both fixes are straightforward and align with existing patterns in CanFunding.hooks.js. Once these are addressed, the PR will be production-ready.

johndeange (approved):
## Review: OPS-5565 — Unsaved changes navigation blocking for CAN forms

### Overview

This PR extracts the navigation-blocking logic (already present on Agreement forms) into a reusable `useNavigationBlocker` hook, then wires it up to both CAN Detail and CAN Funding edit forms. It also adds comprehensive E2E coverage for all blocking scenarios. The refactor is well-motivated and the new hook is a clear improvement in reuse and testability.

---

### Positives

- `useNavigationBlocker` is a clean, well-documented abstraction. The ref pattern for `saveChanges`/`onExit`/`onSaveError` avoids stale closures correctly.
- The "Known React Router bug" comment in `proceedIfBlocked` is exactly the kind of comment that belongs in code — explains a non-obvious workaround.
- `CANDetailForm`'s note "No wasEditModeRef needed: this component unmounts when edit mode toggles off" is a good self-contained explanation.
- `useNavigationBlocker.test.js` covers all paths: blocked/unblocked, save success/failure, secondary action, modal close, and the known React Router error. Thorough.
- E2E tests cover all six scenarios (detail + funding, unsaved/no-change/save/leave/escape). The save-and-revert pattern is good for avoiding data pollution.

---

### Issues to Address

#### Bug — `CreateBLIsAndSCs.hooks.js`: modal closes before save completes

```js
handleConfirm: async () => {
    setShowSaveChangesModal(false);  // ← closes modal first
    await handleSaveRef.current(true);
    blocker.reset();
    navigate(destination);
}
```

The modal closes before the save resolves. If `handleSaveRef.current(true)` throws, the modal is gone and the user gets no feedback. The original order (`await save()` → `setShow(false)`) was correct — this is a silent regression on the error path.

#### Risk — `CanFunding.hooks.js`: verify `budgetEnteredAmount` initialization

```js
const budgetChanged = +budgetEnteredAmount !== +totalFunding;
```

If `budgetEnteredAmount` initializes to `0` (not `totalFunding`) on mount, `hasChanged` will be `true` the moment the user enters edit mode — meaning the no-changes path would erroneously show the blocker. The E2E test implies it works correctly, but worth explicitly confirming how this value is seeded.

#### Design — `setIsCancelling` exposure creates fragile consumer contract

All three consumers must independently know to call `setIsCancelling(true)` on manual cancel. A future consumer could easily forget. Consider wrapping it in a `cancel()` function returned by the hook instead.

#### Test gap — `hasChanged` logic and `CreateBLIsAndSCs` changes are untested

- `CanFunding.hooks.test.js` mocks `useNavigationBlocker` entirely, so the `hasChanged` computation has no unit coverage.
- The `CreateBLIsAndSCs` reordering + switch from `proceedIfBlocked` to `blocker.reset() + navigate()` has no test coverage.

#### Minor
- `onExit` + `location.pathname` effect create a silent dual-reset of edit mode (harmless but adds coupling).
- `beforeEach` is used as a global in `useNavigationBlocker.test.js` while other test globals are explicitly imported — minor consistency nit.

josbell (approved):

Issue #4636, opened 2025-11-12
ui: add validation to ensure agreement title & nickname are unique


fpigeonjr:
### Goals
Need to add a validation on create/edit agreement workflows to ensure the user input for an agreement title or agreement nickname is a unique value.

### Tasks
- [ ] TBD

kimschulke:
We dont currently block users from using a duplicate title in the frontend/UI which can lead them to receive an unclear error message. Hoping to prevent this from happening to people and causing a negative user experience.

jonnalley:
Thank you @weimiao67 for picking this up!

npage123:
Tested in staging and looks good!

https://github.com/user-attachments/assets/9f528e33-76bc-4622-9a48-037a45e65f9c

PR #5670, opened 2026-05-14
feat: add agreement title and nickname frontend validation

weimiao67:
## What changed

Adds inline uniqueness validation for agreement title and nickname fields in the create/edit Agreement form. A new backend endpoint (`GET /agreements/check-unique`) checks whether a candidate value is already in use, respecting existing DB constraints (case-insensitive name scoped by agreement_type, globally unique nick_name). The frontend calls this endpoint on blur and shows an inline error if a duplicate is detected. The check is debounced when agreement_type changes to avoid rapid API calls.

Also adds `onBlur` support to the shared `Input` UI component.

## Issue

#4636

## How to test

1. Navigate to **Agreements > Create Agreement**
2. Select agreement type "Contract"
3. Enter a title that matches an existing contract agreement
4. Tab out of the title field — an inline error should appear: "This title already exists. Try a different one"
5. Change the agreement type — the error should clear/re-evaluate (same title may be valid for a different type)
6. Enter a nickname that matches an existing agreement's nickname, tab out — an inline error should appear: "This nickname already exists. Try a different one"
7. Verify the Continue button is disabled while any uniqueness error is shown
8. Edit an existing agreement — verify it does not flag itself as a duplicate

## A11y impact

## Screenshots
Screenshot 2026-05-14 at 10 43 02 AM


josbell (changes requested):
## Review: Request Changes

After deep investigation of the implementation, I found **2 critical issues** that need to be fixed before merge. I also validated several concerns raised by automated review and found them to be false positives.

---

## ✅ Issues That Need Fixing

### 1. **CRITICAL: Validation Bypass Due to Incomplete Dependency Array**

**Location**: `frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js:274-279`

**Issue**: The `useEffect` that re-checks title uniqueness when `agreementType` changes is missing `agreementTitle` in its dependency array:

```javascript
React.useEffect(() => {
    if (agreementType && agreementTitle) {
        checkUniqueOnBlur("name", agreementTitle);
    }
    // eslint-disable-next-line react-hooks/exhaustive-deps
}, [agreementType]);  // ❌ agreementTitle used but not in deps
```

**Impact**: Users can bypass uniqueness validation by editing the title without changing the agreement type:
1. User types "Agreement A" → uniqueness check runs ✓
2. User changes to "Duplicate Title" (without changing type) → **no check runs** ✗  
3. Form submits with duplicate title, backend rejects it

This defeats the core purpose of this PR (early duplicate detection).

**Fix**:
```javascript
React.useEffect(() => {
    if (agreementType && agreementTitle) {
        checkUniqueOnBlur.cancel();  // Also fix issue #2
        checkUniqueOnBlur("name", agreementTitle);
    }
}, [agreementType, agreementTitle, checkUniqueOnBlur]);  // ✅ Complete deps
```

---

### 2. **Race Condition: Missing Debounce Cancellation on Type Change**

**Location**: Same `useEffect` at lines 274-279

**Issue**: When `agreementType` changes, pending debounced uniqueness checks aren't cancelled before triggering a new check.

**Impact**: If a user rapidly changes agreement types, the old debounced call (300ms delay) may execute after the new type is selected, showing stale validation errors for the wrong type.

**Fix**: Add `checkUniqueOnBlur.cancel();` before the new check (included in Fix #1 above).

---

## ❌ False Positives (No Action Needed)

I also investigated several other concerns and found them to be **invalid**:

### ✅ Case-Sensitivity Is Consistent
- **Claim**: Frontend/backend case-sensitivity mismatch for title
- **Reality**: All three layers (DB, backend query, frontend) use case-**in**sensitive comparison for title via `lower()` / `.toLowerCase()`. No mismatch exists.

### ✅ Vest v6 Pattern Is Correct  
- **Claim**: Conditional `if (fieldName) { only(fieldName); }` violates Vest v6 patterns
- **Reality**: This is actually the **safer pattern**. The CLAUDE.md warning is about passing the data object to `only()`, not about conditional usage.

### ✅ Nickname Case-Sensitivity Is By Design
- All layers consistently treat nickname as case-**sensitive** (unlike title which is case-insensitive). This is working as designed.

---

## Summary

**Please fix the 2 issues in the `useEffect` at lines 274-279:**
1. Add complete dependency array: `[agreementType, agreementTitle, checkUniqueOnBlur]`
2. Add `checkUniqueOnBlur.cancel();` before calling the debounced function

These are critical to ensuring the uniqueness validation works correctly in all user workflows.

weimiao67:
@josbell Thanks for the review and comments! I looked into both. 
1.  The useEffect at 274–279 is a secondary trigger for the per-type re-check; the primary triggers are onBlur on the title input (AgreementEditForm.jsx:208) and verifyUniquenessBeforeSubmit on Save Draft / Continue (hooks.js:458). Adding agreementTitle to the deps would shift the check from on-blur to on-every-keystroke, which is a behavior change. 
2. The pending debounce is already cancelled when checkUniqueOnBlur is rebuilt — see the cleanup at hooks.js:270, which fires whenever agreementType changes.

Let me know if I missed anything.

johndeange (approved):
## Code Review: Agreement Title & Nickname Frontend Uniqueness Validation

**Approve with minor suggestions.**

### Strengths

- Clean separation of concerns — uniqueness logic lives in the hook, presentation in the JSX
- Edit-mode self-detection correctly avoids flagging the current agreement as its own duplicate
- Debounced re-check on type change since the name constraint is scoped per `agreement_type`
- Pre-submit verification cancels pending debounce and re-runs fresh checks before saving
- Good test coverage — hook unit tests, component wiring tests, and Input component tests (3 new test files)
- Graceful failure — API errors clear the error state rather than showing false positives

**1. Case-sensitivity in `nick_name` self-detection**

The name comparison is case-insensitive but the nickname comparison is strict:

```javascript
field === "name"
    ? (agreement?.name ?? "").toLowerCase() === trimmed.toLowerCase()
    : agreement?.nick_name === trimmed  // ← strict equality
```

If the backend enforces case-insensitive uniqueness on `nick_name`, editing an agreement with nickname "ABC" and changing to "abc" would incorrectly flag it as a conflict with itself. Worth verifying the DB constraint.

**2. `eslint-disable` on the `agreementType` effect**

Consider adding a brief comment explaining *why* `agreementTitle` and `checkUniqueOnBlur` are intentionally excluded from the dep array — prevents a future dev from "fixing" it and causing excessive API calls.

**3. PR description mentions a backend endpoint (`GET /agreements/check-unique`) that doesn't exist in the diff** — the implementation reuses the existing agreements list endpoint with filters. Minor description update would help future readers.

### Verdict

Solid implementation with good test coverage. No blocking issues.

rajohnson90 (approved):

weimiao67:
> ## Code Review: Agreement Title & Nickname Frontend Uniqueness Validation
> **Approve with minor suggestions.**
> 
> ### Strengths
> * Clean separation of concerns — uniqueness logic lives in the hook, presentation in the JSX
> * Edit-mode self-detection correctly avoids flagging the current agreement as its own duplicate
> * Debounced re-check on type change since the name constraint is scoped per `agreement_type`
> * Pre-submit verification cancels pending debounce and re-runs fresh checks before saving
> * Good test coverage — hook unit tests, component wiring tests, and Input component tests (3 new test files)
> * Graceful failure — API errors clear the error state rather than showing false positives
> 
> ### Suggestions (non-blocking)
> **1. Case-sensitivity in `nick_name` self-detection**
> 
> The name comparison is case-insensitive but the nickname comparison is strict:
> 
> ```js
> field === "name"
>     ? (agreement?.name ?? "").toLowerCase() === trimmed.toLowerCase()
>     : agreement?.nick_name === trimmed  // ← strict equality
> ```
> 
> If the backend enforces case-insensitive uniqueness on `nick_name`, editing an agreement with nickname "ABC" and changing to "abc" would incorrectly flag it as a conflict with itself. Worth verifying the DB constraint.
> 
> **2. `eslint-disable` on the `agreementType` effect**
> 
> Consider adding a brief comment explaining _why_ `agreementTitle` and `checkUniqueOnBlur` are intentionally excluded from the dep array — prevents a future dev from "fixing" it and causing excessive API calls.
> 
> **3. PR description mentions a backend endpoint (`GET /agreements/check-unique`) that doesn't exist in the diff** — the implementation reuses the existing agreements list endpoint with filters. Minor description update would help future readers.
> 
> ### Verdict
> Solid implementation with good test coverage. No blocking issues.

Database constraint (backend/models/agreements.py:186):
nick_name: Mapped[Optional[str]] = mapped_column(String, unique=True, nullable=True)
This is a plain unique=True column constraint — case-sensitive.

josbell (approved):
## Approve with Optional Improvements

After running your defense through two independent code review agents, I'm approving this PR. Your implementation is **functionally correct** and the validation cannot be bypassed in practice.

### Summary

Both agents confirmed:
- ✅ Pre-submit verification (`verifyUniquenessBeforeSubmit`) prevents all bypass scenarios
- ✅ onBlur validation is properly wired
- ✅ Cleanup mechanism at line 270 correctly cancels pending debounced calls
- ✅ No race conditions occur in practice

You were right that adding `agreementTitle` to deps would cause suboptimal UX (debounce resets), and that the cleanup mechanism works correctly.

---

### Optional Suggestions (Non-Blocking)

These would improve code maintainability but aren't required:

**1. Add clarifying comment to the useEffect (lines 274-279):**

```javascript
React.useEffect(() => {
    // NOTE: agreementTitle intentionally excluded from deps to avoid re-triggering
    // on every keystroke. Primary validation happens on blur (line 208) and 
    // pre-submit (lines 467, 505). This effect only re-checks when type changes
    // because the backend constraint is scoped per agreement_type.
    if (agreementType && agreementTitle) {
        checkUniqueOnBlur("name", agreementTitle);
    }
}, [agreementType, checkUniqueOnBlur]);
```

**Why:** Makes the intentional eslint-disable clear to future maintainers.

**2. Consider explicit cancellation (optional):**

```javascript
React.useEffect(() => {
    if (agreementType && agreementTitle) {
        checkUniqueOnBlur.cancel();  // Explicit - makes intent clear
        checkUniqueOnBlur("name", agreementTitle);
    }
}, [agreementType, checkUniqueOnBlur]);
```

**Why:** While the cleanup at line 270 works correctly, explicit cancellation is more defensive against future refactoring and makes the intent obvious to reviewers.

---

### Verdict

The original review overstated the severity by calling these "critical issues" when they're actually code quality/maintainability suggestions. Your defense successfully demonstrated that no validation bypasses are possible.

**Approving as-is.** Feel free to incorporate the suggestions if you'd like, but they're not blockers.

Issue #5674, opened 2026-05-14
Load missing agreement data provided by OPRE


johndeange:
### User Story

As an OPS user, I want to have accurate data associated with an agreement.

Over the course of development, certain fields that are now required for agreements when they are in certain states. The production DB has some missing data for these fields. 

We sent OPRE a spreadsheet of the missing data for them to fill in. 

### Acceptance Criteria
- [ ] Scripts created for loading the data.
- [ ] Data loaded into production.

-----------------------------------------

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] Usability validated (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] New design/tech debt documented (if applicable)
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

### Additional Context & Resources
* _Any additional information that would be useful to share about this work_
* _Dependencies, constraints?_
* _Info already known or found, previous work that's been done?_

johndeange:
# Plan: Load Agreement Missing Data Script

## Overview

Create a new data loading script at `backend/data_tools/src/load_agreement_missing_data/` that reads a TSV file, looks up agreements by ID, and updates `description`, `vendor`, and `project_officer` fields when they differ from current values.

## Input Data

TSV file with columns:
- `agreement_id` — integer, used for lookup
- `agreement_description` — string, update target
- `agreement_vendor` — vendor name (string), must resolve to `vendor_id` FK
- `agreement_project_officer` — person name (string), must resolve to `project_officer_id` FK
- Other columns (project_name, agreement_name, agreement_type, violation_type, agreement_title, agreement_agreement_type) — informational only, not used for updates

## Architecture

Follow the existing `load_*/` module pattern:

```
backend/data_tools/src/load_agreement_missing_data/
├── __init__.py
└── utils.py
```

## Implementation Details

### 1. Dataclass: `AgreementUpdateData`

Fields from TSV row:
- `agreement_id: int`
- `agreement_description: Optional[str]`
- `agreement_vendor: Optional[str]`
- `agreement_project_officer: Optional[str]`

`__post_init__`:
- Strip whitespace from all string fields
- Convert `agreement_id` to int
- Normalize empty strings / "TBD" to `None` for description (or keep "TBD" if that's intentional — the data shows "TBD" as a valid value to set)

### 2. Lookup Strategy

#### Agreement Lookup
[...]

PR #5678, opened 2026-05-15
feat: add load_agreement_missing_data ETL script

rajohnson90:
## What changed

Adds a new `load_agreement_missing_data` ETL script that reads a TSV file and backfills missing `description`, `vendor`, and `project_officer` fields on existing agreements.

**`backend/data_tools/src/load_agreement_missing_data/utils.py`**
- `AgreementUpdateData` dataclass parses TSV rows, strips whitespace, and normalizes empty strings to `None`
- `get_or_create_vendor()` does a case-insensitive name lookup; creates a new active `Vendor` record if no match is found
- `get_user_by_name()` does a case-insensitive full-name lookup; logs a warning and skips the field if the user is not found (or logs an error if multiple matches exist)
- `update_agreement()` compares each field against the current value and only writes what has changed, emitting an `OpsEvent` and calling `agreement_history_trigger_func` for audit history; supports `DRY_RUN` env var
- `transform()` is the standard entry point matching all other load scripts

**`backend/data_tools/src/load_data.py`**
- Registered `agreement_missing_data` in the `--type` CLI choices and match/case dispatch block

**`backend/data_tools/tests/load_agreement_missing_data/test_utils.py`**
- 14 unit tests covering: happy path (all three fields), description-only update, no-change short-circuit, vendor auto-create, user-not-found (skips officer but updates other fields), agreement-not-found (skips entire row), vendor whitespace normalization, and case-insensitive matching

**`backend/data_tools/test_csv/agreement_missing_data.tsv`**
- Manual test fixture with 6 rows: 3 complete entries, 1 with a missing agreement ID, 1 with a vendor that triggers auto-creation, and 1 with a user that cannot be found

## Issue

https://github.com/HHS/OPRE-OPS/issues/5674

## How to test

**Automated tests:**
```bash
cd backend/data_tools
pipenv run python -m pytest tests/load_agreement_missing_data/ -v
```

**Manual end-to-end test (requires running Docker stack):**
```bash
cd backend
python -m data_tools.src.load_data \
  --type agreement_missing_data \
  --env local \
  --input-csv data_tools/test_csv/agreement_missing_data.tsv
```

Expected behavior per row:
- Rows 1–3 (agreement IDs 6, 7, 8): description, vendor, and project officer updated
- Row 4 (agreement ID 9999): warning logged, row skipped — no DB changes
- Row 5 (agreement ID 9, unknown vendor): new Vendor record created and assigned
- Row 6 (agreement ID 10, unknown user): warning logged for project officer, description and vendor still updated

Run with `DRY_RUN=1` to verify changes are computed without committing them.

## A11y impact

## Screenshots

N/A — backend/data-tools changes only

## Links

N/A

### DIFF
--- a/backend/data_tools/src/load_agreement_missing_data/utils.py
+++ b/backend/data_tools/src/load_agreement_missing_data/utils.py
@@ -90,9 +90,7 @@ def get_or_create_vendor(vendor_name: str, sys_user: User, session: Session) -> 
     :param session: The database session.
     :return: The Vendor instance, or None if vendor_name is empty.
     """
-    vendor = session.execute(
-        select(Vendor).where(func.upper(Vendor.name) == vendor_name.upper())
-    ).scalar_one_or_none()
+    vendor = session.execute(select(Vendor).where(func.upper(Vendor.name) == vendor_name.upper())).scalar_one_or_none()
 
     if vendor:
         return vendor
@@ -117,11 +115,13 @@ def get_user_by_name(name: str, session: Session) -> Optional[User]:
     :param session: The database session.
     :return: The User instance, or None if not found.
     """
-    results = session.execute(
-        select(User).where(
-            func.upper(func.concat(User.first_name, " ", User.last_name)) == name.upper()
+    results = (
+        session.execute(
+            select(User).where(func.upper(func.concat(User.first_name, " ", User.last_name)) == name.upper())
         )
-    ).scalars().all()
+        .scalars()
+        .all()
+    )
 
     if len(results) > 1:
         logger.error(
--- a/backend/data_tools/src/load_agreement_missing_data/utils.py
+++ b/backend/data_tools/src/load_agreement_missing_data/utils.py
@@ -1,0 +1,264 @@
+import os
+from csv import DictReader
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+from loguru import logger
+from sqlalchemy import func, select
+from sqlalchemy.orm import Session
+
+from models import (
+    Agreement,
+    OpsEvent,
+    OpsEventStatus,
+    OpsEventType,
+    User,
+    Vendor,
+    agreement_history_trigger_func,
+)
+
+
+@dataclass
+class AgreementUpdateData:
+    """
+    Dataclass to represent a row from the missing agreement data TSV.
+    """
+
+    agreement_id: int
+    agreement_description: Optional[str] = field(default=None)
+    agreement_vendor: Optional[str] = field(default=None)
+    agreement_project_officer: Optional[str] = field(default=None)
+
+    def __post_init__(self):
+        if not self.agreement_id:
+            raise ValueError("agreement_id is required.")
+
+        self.agreement_id = int(self.agreement_id)
+        self.agreement_description = self.agreement_description.strip() if self.agreement_description else None
+        self.agreement_vendor = self.agreement_vendor.strip() if self.agreement_vendor else None
+        self.agreement_project_officer = (
+            self.agreement_project_officer.strip() if self.agreement_project_officer else None
+        )
+
+        if not self.agreement_vendor:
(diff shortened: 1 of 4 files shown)


backend/data_tools/src/load_agreement_missing_data/utils.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def validate_data(data: AgreementUpdateData) -> bool:
+    """
+    Validate an AgreementUpdateData instance.
+
+    :param data: The AgreementUpdateData instance to validate.
+    :return: True if valid, False otherwise.
+    """
+    return data.agreement_id is not None
```
weimiao67, line 71: validate_data is dead code, `__post_init__ `already raises ValueError if agreement_id is falsy (line 33). By the time you have an AgreementUpdateData instance, agreement_id is guaranteed non-None and an int. validate_data can never return False, and validate_all therefore can never fail. The whole validation block in transform() (lines 257-261) is dead.

weimiao67 (approved):

jonnalley (approved):
LGTM!  CC was also quite happy:
```
The PR follows existing patterns correctly and introduces no new bugs.
```

johndeange (approved):
## Code Review: `load_agreement_missing_data` ETL Script

### Strengths

- **Consistent with existing patterns**: Structure mirrors `load_award_date` closely — same commit-per-row approach, same `DRY_RUN` env var, same `agreement_history_trigger_func` usage.
- **Good defensive coding**: Case-insensitive lookups for vendors and users, whitespace stripping, graceful handling of missing agreements/users.
- **Solid test coverage**: Tests cover happy path, edge cases (whitespace, case-insensitivity, not-found), and the fixture cleanup is thorough.

### Issues

#### Bug: Whitespace-only description not normalized to `None`

**`utils.py:37-39`** — Vendor and project_officer get a second normalization pass (`if not self.agreement_vendor: self.agreement_vendor = None`), but `agreement_description` does not. A whitespace-only description like `"   "` would strip to `""`, which passes the `if data.agreement_description is not None` check in `update_agreement` and write an empty string to the database.

```python
# Missing after line 35:
if not self.agreement_description:
    self.agreement_description = None
```

#### Dead validation logic

**`utils.py:65-83`** — `validate_data()` checks `agreement_id is not None`, but `__post_init__` already raises `ValueError` when `agreement_id` is falsy. The `validate_all()` function can never return `False` because `create_all_agreement_update_data()` would have already raised during construction. This validation layer gives a false sense of safety — the `RuntimeError("Validation failed.")` in `transform()` is unreachable.

Consider either:
- Removing the dead validation, or
- Moving the `ValueError` out of `__post_init__` and into `validate_data` so validation actually catches invalid rows without raising during construction.

#### No test for `DRY_RUN` behavior

The `DRY_RUN` env var triggers a rollback path that has no test coverage. Consider adding a test using `monkeypatch.setenv("DRY_RUN", "1")` to verify changes are rolled back.

#### No test for `transform()` entry point

The top-level `transform()` function isn't tested. A single integration test verifying the full pipeline from `DictReader` → DB would catch wiring issues.

### Minor / Style Notes

- **`utils.py:30`**: `if not self.agreement_id` would also reject `agreement_id=0`. Since 0 isn't a valid DB ID this is fine, but `int(self.agreement_id)` on the next line would succeed for `"0"`, so the error message could be confusing. Consider checking after the `int()` conversion.

- **`utils.py:50`**: `create_agreement_update_data` passes `data.get("agreement_id", "")` — if the TSV has a typo in the header, this silently passes `""` which then raises a generic `ValueError`. A more explicit error would help operators debug TSV issues.

### Summary

Solid PR that follows project conventions well. The whitespace-only description bug is the only item I'd flag as worth fixing before merge — it's a one-liner. The dead validation logic is worth a brief discussion but not blocking. Approving — ship it after addressing the description normalization.

Issue #5679, opened 2026-05-15
Agreement draft saved not appearing in agreement list


npage123:
After saving a draft of an agreement on step 2 of creating an agreement, I'm not seeing it appear in the agreement list.

## Expected Behavior
Saving a draft of an agreement should create the agreement and it should appear in the agreement list.

## Current Behavior
After saving a draft of an agreement on step 2 of creating an agreement, I'm not seeing it appear in the agreement list.

## Steps to Reproduce (see attached recording)
1. Log into staging
2. Go to create agreement from Create in main nav
3. Select a project and click Continue
4. Enter agreement name and select service requirement type
5. Click Save Draft
6. On agreements list page, select All from FY dropdown
7. Search for agreement just created

https://github.com/user-attachments/assets/46d4804e-77d6-4eee-8eb9-0994a472f146

jonnalley:
I confirmed the agreement @npage123  tried in his video up above never actually got created. Some of CC's thoughts:

The only code path that actually creates (POSTs) a new agreement is in **Step 3** (`CreateBLIsAndSCs.hooks.js` line 873 via `addAgreement()`). The `saveAgreement()` function in `AgreementEditForm.hooks.js` only supports PATCH (updating existing agreements).

## Why This Is a Regression

The most likely cause is recent refactoring that changed the wizard flow. Candidates:

- **Commit 82060b4cc** (May 4, 2026): Navigation blocker refactoring — may have simplified `handleDraft` and removed a fallback that previously called `addAgreement` for new agreements
- The wizard structure may have previously POST'd the agreement on Step 1's "Continue" click, giving it an ID before Step 2

## Key Files

| File | Role |
|------|------|
| `frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js` | `handleDraft` (line 417) and `saveAgreement` (line 277) |
| `frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.jsx` | Save Draft button (line 518-528) |
| `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js` | `addAgreement` POST logic (line 830-873) |
| `frontend/src/api/opsAPI.js` | `addAgreement` mutation (POST) and `updateAgreement` mutation (PATCH) |
| `frontend/src/helpers/agreement.helpers.js` | `cleanAgreementForApi()` |

## Proposed Fix

[...]

PR #5691, opened 2026-05-18
fix(agreements): create new agreement on save draft from step 2

jonnalley:
## What changed

Fixed a bug where clicking "Save Draft" on Step 2 of the agreement creation wizard silently failed to create the agreement. The user was navigated to `/agreements` but nothing was persisted.

**frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js**
- Added `useAddAgreementMutation` import and hook initialization
- Modified `handleDraft()` to detect when `saveAgreement()` returns `false` (no existing agreement to PATCH) and the agreement has no `id`, then POSTs the new agreement via `addAgreement`
- Added proper error handling with user-facing alerts

**frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.test.js** (new)
- Unit tests verifying `handleDraft` creates new agreements, skips creation for existing agreements, and handles errors

**frontend/cypress/e2e/createAgreement.cy.js**
- New E2E test "can save a draft agreement from step 2" that verifies the POST, navigation, and cleanup

## Issue

https://github.com/HHS/OPRE-OPS/issues/5679

## How to test

1. **Unit tests:**
   ```bash
   cd frontend
   bun run test -- --watch=false src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.test.js
   ```

2. **E2E test:**
   ```bash
   cd frontend
   bun run test:e2e -- --spec cypress/e2e/createAgreement.cy.js
   ```

3. **Manual verification:**
   - Start the Docker/podman stack using whatever method you find most gratifying personally
   - Navigate to Create Agreement
   - Select a project, click Continue
   - On Step 2, fill in agreement type, name, contract type, and service requirement type
   - Click "Save Draft"
   - Verify the agreement appears in the agreements list

## A11y impact

## Screenshots

N/A — logic-only fix, no visual changes.

## Links

N/A


rajohnson90 (changes requested):
PR #5691 Review — fix(agreements): create new agreement on save draft from step 2                                                                    
                                                                                                                                                       
  Overview                                                                                                                                             
                                                                                                                                                       
  This PR fixes a silent failure bug where "Save Draft" on step 2 of the agreement creation wizard navigated the user to /agreements without persisting
   anything. The root cause: handleDraft() called saveAgreement(), which only PATCHes agreements that already have an id — new agreements fell through
  to a return false with no POST ever made.                                                                                                            
                             
  The fix intercepts the false return value, checks that the agreement has no id, and calls addAgreement to POST it. Unit and E2E tests are included.  
   
  ---                                                                                                                                                  
  Issues                     
                                                                                                                                                       
  Bug: Double error alert for UPDATE failures
                                                                                                                                                       
  This is the most significant correctness regression. saveAgreement() already handles update errors by calling setAlert(...) and then re-throwing. The
   old handleDraft catch block had a comment acknowledging this:                                                                                       
                                                                                                                                                       
  // Error already handled in saveAgreement with alert and redirect
  return;                                                                                                                                              
                                                                                                                                                       
  The new catch block replaces that with another setAlert(...) call. For any existing agreement that fails to update via handleDraft, the user now sees
   two error alerts — one from saveAgreement and one from the new catch. This affects the edit flow on step 2.                                         
                                                                                                                                                       
  Fix: Propagate only errors that aren't already handled, or check whether the error already triggered an alert. The simplest fix is to restore the    
  original catch intent:
                                                                                                                                                       
  } catch (error) {          
      // Only set alert if saveAgreement didn't already handle it                                                                                      
      // (saveAgreement sets an alert and re-throws for UPDATE failures)                                                                               
      if (!agreement.id) {                                                                                                                             
          setAlert({                                                                                                                                   
              type: "error",                                                                                                                           
              heading: "Error",                                                                                                                        
              message: "An error occurred while saving the agreement.",                                                                                
              redirectUrl: "/error"                                                                                                                    
          });                                                                                                                                          
      }                                                                                                                                                
      return;                                                                                                                                          
  }                                                                                                                                                    
                                                                                                                                                       
  Or better: pass a flag/sentinel through the thrown error to differentiate.                                                                           
                             
  progress/session-1-fix-save-draft-step2.md should not be committed                                                                                   
                             
  progress/session-1-fix-save-draft-step2.md is a development session log (review history, status notes, environment caveats). It does not belong in   
  the repository. This type of context belongs in the PR description or a ticket comment, not committed source files.
                                                                                                                                                       
  Data preparation duplicated from saveAgreement                                                                                                       
   
  handleDraft manually rebuilds the data object — including team_members, requesting_agency_id, servicing_agency_id — duplicating identical logic      
  already in saveAgreement. These diverge subtly: saveAgreement uses selectedTeamMembers (destructured from context), while handleDraft uses
  agreement.team_members. They should be the same data, but the inconsistency is fragile.                                                              
                             
  A cleaner approach: add a skipIdCheck parameter to saveAgreement (similar to the existing skipChangeCheck) that lets it call POST when there's no id,
   instead of returning false. That consolidates all save logic in one place.
                                                                                                                                                       
  result === false is fragile                                                                                                                          
   
  The condition if (result === false && !agreement.id) relies on saveAgreement returning exactly false (strict equality). Since the function contract  
  is currently true | false | throw, this works — but it's an implicit contract that's easy to break. A named constant, a return type comment, or an
  enum-style result object would make this more maintainable.                                                                                          
                             
  ---
  Code Quality
                                                                                                                                                       
  - E2E bearer_token timing: The const bearer_token = ...window.localStorage.getItem(...) is called synchronously outside any Cypress command. This
  works if the token is always set before the test runs, but it's fragile — if the auth state is async, this could return null. Consider               
  cy.window().then(win => win.localStorage.getItem('access_token')) inside the cleanup block.
  - Unit test coverage gap: There's no test for the "existing agreement with changes → UPDATE succeeds" path through handleDraft. Given the            
  double-alert bug above, a test for "existing agreement UPDATE fails" would have caught the regression.                                               
  - console.log at line 298 (UPDATE: agreement updated: ...) is an existing issue but worth noting — debug logs like this typically don't belong in
  production code.                                                                                                                                     
                             
  ---                                                                                                                                                  
  Summary                    
                                                                                                                                                       
  The fix is correct for the primary case (new agreement, no id, step 2 save draft). But two things need to be addressed before merging:
                                                                                                                                                       
  1. The double error alert regression for UPDATE failures — this is a real behavioral bug introduced by the new catch block.                          
  2. Remove progress/session-1-fix-save-draft-step2.md from the commit — it's personal session notes, not repo content.                                
                                                                                                                                                       
  The data duplication and result === false fragility are lower-priority but worth addressing if the code is being refactored anyway.

Note from Reed:
I tested this and the fix does absolutely work. That said, I think the AI PR review is correct that I'm not sure why the error logic was duplicated here. Also I don't think the readme file should be committed either.

jonnalley (commented):
Thanks for the review @rajohnson90. Addressed the two blocking items in `01d8c2f`:

**1. Double error alert — fixed.** The catch block now guards `setAlert` with `!agreement.id`, so it only fires for `addAgreement` failures. When `saveAgreement` re-throws an update error (which it already handled with its own alert), the catch just returns without a second alert. Added a unit test for this path.

**2. Progress file — removed.** Deleted from the commit and added `progress/` to `.gitignore`.

**Items not addressed (intentionally):**

- **Data preparation duplication / `skipIdCheck` refactor** — The duplication is 6 lines of straightforward data assembly. Introducing a new parameter to `saveAgreement` changes its contract and touches a `useCallback` with an existing dependency array — broader scope than the bug fix warrants. If we refactor the save flow later, we can consolidate then.

- **`result === false` fragility** — The function's return contract (`true | false | throw`) is stable and tested. Adding a named constant or return type for a two-value boolean would be over-engineering for this context.

- **E2E `bearer_token` timing** — This follows the identical pattern used in every other test in the file (lines 145, 312). Changing it here would be inconsistent; if it needs fixing, it's a separate cleanup across the file.

- **`console.log` on line 298** — Pre-existing, not introduced by this PR.

josbell (changes requested):
## Code Review - Issues Confirmed

Great work fixing the silent save failure! The core logic is sound, but investigation revealed two issues that need attention.

---

## 🔴 Required Fix: Agreement ID Not Captured

**File:** `frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js:535`

**Issue:** After `addAgreement().unwrap()` succeeds, the response containing the new agreement ID is discarded. The context still has `agreement.id === undefined` even though the server created the agreement.

**Current code:**
```javascript
await addAgreement(cleanData).unwrap();  // Response thrown away
```

**Evidence:**
- Backend returns `{ id: agreement.id, ... }` (backend/ops_api/ops/resources/agreements.py:327-335)
- The `AgreementEditorContext` maintains local state via reducer
- RTK Query cache invalidation doesn't auto-update the context
- After creation, context has `id: undefined` but server has the agreement with an ID

**Impact:**
- If navigation fails (network error, route guard, etc.), state is inconsistent
- Fragile pattern that relies entirely on successful navigation
- Any subsequent operations expecting an ID would fail

**Fix:**
```javascript
const response = await addAgreement(cleanData).unwrap();
dispatch({ type: "UPDATE_AGREEMENT", key: "id", value: response.id });
```

---

## 🟡 Recommended: Enhance E2E Test

**File:** `frontend/cypress/e2e/createAgreement.cy.js:406-445`

**Issue:** The E2E test verifies the API POST succeeded but doesn't verify what the user actually sees (success alert, agreement in list).

**What's missing:**
- ❌ Success alert display check
- ❌ Agreement appearing in list check
- ✅ API call verification (good!)
- ✅ Navigation verification (good!)

**Comparison:** Other E2E tests verify user-facing elements:
- `agreementDetailsEdit.cy.js:128` checks alert body
- `createResearchProject.cy.js:52` checks success message

**Suggested enhancement** (add after line 443):
```javascript
// Verify success alert shown to user
cy.get('[data-cy="alert"]').should("contain", "Agreement Draft Saved");
cy.get('[data-cy="alert"]').should("contain", "E2E Draft Save Test");

// Verify agreement appears in the list
cy.contains("E2E Draft Save Test").should("exist");
```

---

## ✅ False Alarms - No Action Needed

**Redundant error condition (line 546):** The `if (!agreement.id)` guard is NOT redundant - it prevents double error alerts when UPDATE operations fail. This was specifically fixed in commit `01d8c2fdd`. Code is correct as-is.

**Race condition:** No practical race condition - the `agreement` object is context-based and stable during async execution. RTK Query doesn't auto-update the context.

---

## Summary

**Must fix:** Capture and store agreement ID after creation (Issue #1)

**Should enhance:** Add user-facing assertions to E2E test (Issue #2)

Once the agreement ID is captured, this will be good to merge! The E2E enhancement can be included here or in a follow-up, your choice.

Full investigation report with detailed evidence available if needed.

rajohnson90 (approved):

jonnalley:
@josbell Thank you for looking at this! So my AI reviewer disagrees with your AI review. 😄 In all seriousness, one of my sub-agents suspected your issue #1 as well before I PR'ed this but other agents found it to be inaccurrate then.  Here's an assessment from Opus:

### Issue 1 (Red - "Required Fix"): Agreement ID Not Captured

**Reviewer's claim:** After `addAgreement().unwrap()` succeeds, the response containing the new agreement ID is discarded, leaving context with `agreement.id === undefined`. They suggest dispatching an action to update context.

**Assessment: NOT A BUG**

Verified facts:
- The `EditAgreementProvider` context is scoped to the wizard pages only (`CreateAgreement.jsx` wraps its content in the provider)
- After `addAgreement` succeeds, `handleDraft` calls `navigate("/agreements")` immediately
- Navigation unmounts the wizard component, **destroying the context entirely**
- The agreements list page at `/agreements` fetches its own data fresh via `useGetAgreementsQuery()`
- RTK Query's `invalidatesTags: ["Agreements"]` on the `addAgreement` mutation ensures the list refetches
- The ID is never needed again — there is no code path that reads `agreement.id` from context after navigation

The "impact" scenarios are theoretical:
- "If navigation fails" — `navigate()` is synchronous client-side routing, not a network call
- "Subsequent operations expecting an ID" — there are none; the context is destroyed

Additionally, `CreateBLIsAndSCs.hooks.js` (the Step 3 code) captures the response from `addAgreement` at line 873 but also **never uses it** beyond logging. The pattern of discarding the response is consistent across the codebase.

**Conclusion: No action.

### Issue 2 (Yellow - "Recommended"): Enhance E2E Test

That's a valid improvement, I like it. However, I was intentionally scoping my change to just the issue at hand and not trying to make more wider-impact testing improvements in this effort. The existing E2E tests in this file (lines 146-164, 311-336) follow the identical pattern: verify the API response, verify navigation, clean up. None of them check alert display or list presence. Adding these checks in this case, while probably needed and an improvement, would be inconsistent with the rest of the file unless all tests were updated.

The test already verifies: (1) POST returned 201, (2) response contains "Agreement created", (3) URL includes "/agreements". This confirms the feature works end-to-end. But, i like the idea still. If it would help, what I can do is add a backlog item to address this more globally with the e2e tests about what the user sees across other tests too and I can plan to address that in the next week or two if you want

weimiao67 (approved):

josbell (changes requested):
## Independent Code Review - PR #5691

I conducted a deep investigation of this PR using two independent review agents (security/correctness focus + architecture/maintainability focus). All findings have been verified through code inspection.

---

## 🔴 BLOCKING ISSUE: Agreement ID Not Captured After Creation

**Location:** `frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js:535`

**Problem:** After `addAgreement().unwrap()` succeeds, the response containing the new agreement ID is discarded. The context still has `agreement.id = undefined` even though the server created the agreement.

**Current code:**
```javascript
await addAgreement(cleanData).unwrap();  // Response thrown away
```

**Verified Evidence:**
- ✅ Backend returns `{ id: agreement.id, ... }` (confirmed in `agreements.py:327-335`)
- ✅ Context has `UPDATE_AGREEMENT` action (confirmed in `AgreementEditorContext.hooks.js:79-84`)
- ✅ E2E test extracts `body.id` successfully (line 428)
- ✅ Similar code in `CreateBLIsAndSCs.hooks.js:873` captures the response

**Impact:**
- **Data consistency hazard**: Frontend context claims no agreement exists (`id: undefined`) while backend has persisted it
- **Race condition window**: If navigation fails/delays, subsequent "Save Draft" click would attempt another POST → **duplicate agreement**
- **Violates architectural invariant**: `agreement.id` should reflect database reality

**Required Fix:**
```javascript
const response = await addAgreement(cleanData).unwrap();
dispatch({ type: "UPDATE_AGREEMENT", key: "id", value: response.id });
```

---

## 🟡 RECOMMENDED: Code Duplication - Data Preparation Logic

**Location:** `AgreementEditForm.hooks.js` - Lines 526-534 (handleDraft) vs Lines 357-365 (saveAgreement)

**Problem:** 8 lines of data preparation logic duplicated between two functions, already showing divergence:
- `handleDraft` uses `agreement.team_members ?? []`
- `saveAgreement` uses `selectedTeamMembers`

**Verified Evidence:**
- ✅ Exact same pattern exists in `CreateBLIsAndSCs.hooks.js:856-865` (codebase-wide issue)
- ✅ This is pre-existing technical debt, not introduced by this PR

**Impact:**
- Maintenance burden: changes must be made in multiple places
- Risk of bugs from code drift (already diverging)

**Suggested Fix:** Extract to shared helper function:
```javascript
const prepareAgreementData = () => {
    return {
        ...agreement,
        team_members: (agreement.team_members ?? []).map(formatTeamMember),
        requesting_agency_id: requestingAgency?.id ?? null,
        servicing_agency_id: servicingAgency?.id ?? null
    };
};
```

**Note:** This can be addressed in follow-up if timeline is tight, since it's pre-existing tech debt.

---

## 💡 ENHANCEMENT: E2E Test Coverage Gaps

**Location:** `frontend/cypress/e2e/createAgreement.cy.js:406-445`

**Issue:** Test verifies API call and navigation, but not user-facing experience:
- ✅ API POST verified
- ✅ Navigation verified
- ❌ Success alert not checked
- ❌ Agreement appearance in list not checked

**Verified Evidence:**
- ✅ Other E2E tests DO check alerts (e.g., `agreementDetailsEdit.cy.js:128`)
- ✅ Success alert IS shown in code (`AgreementEditForm.hooks.js:536-540`)

**Suggested Enhancement:** Add after line 431:
```javascript
// Verify success alert shown to user
cy.get(".usa-alert__heading").should("contain", "Agreement Draft Saved");
cy.get(".usa-alert__body").should("contain", "E2E Draft Save Test");

// Verify agreement appears in the list
cy.contains("E2E Draft Save Test").should("exist");
```

**Note:** Can be follow-up PR if preferred.

---

## 💡 TECH DEBT: Fragile Boolean Return Pattern

**Location:** `AgreementEditForm.hooks.js:524-525`

**Issue:** `saveAgreement()` returns `false` for TWO different reasons:
1. No changes detected (line 368)
2. No ID to update (line 421)

Code disambiguates with `!agreement.id` check, but this is implicit knowledge.

**Verified Evidence:**
- ✅ Boolean return has ambiguous semantics
- ✅ Works correctly now, but fragile for future changes

**Note:** This is architectural tech debt, can be addressed in future refactor.

---

## Summary

**Must fix before merge:**
- 🔴 Capture agreement ID after creation (Issue #1)

**Strongly recommended:**
- 🟡 Extract duplicate data preparation logic (Issue #2)

**Can be follow-up:**
- 💡 Enhance E2E test coverage (Issue #3)
- 💡 Refactor boolean return pattern (Issue #4)

The core fix works correctly and unit tests are solid. Issue #1 is the only blocker - it's a data consistency risk that's easy to fix. Once addressed, this will be good to merge!

**Verification method:** Two independent AI code review agents + manual code inspection of backend API, context reducers, and E2E tests.

jonnalley:
Re: Issue 1: Agreement ID Not Captured After Creation - `frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js`

So, in looking at the order of operations:

 - `navigate("/agreements")` is called immediately after the POST completes, which triggers a route transition
 - The `EditAgreementProvider` context (and all of its state) is destroyed when `CreateAgreement` unmounts on that route transition
 - There's not a code path that reads `agreement.id` from context after this point. Meaning nothing consumes or reads the agreement's ID at that point so if it's not set, it shouldn't be an issue anyway.
 - The identical pattern exists in the file CreateBLIsAndSCs.hooks.js:873-874 that you mentioned, where the payload is unwrapped, and then the response is captured **only for console.log** and never dispatched to state at any point if you follow the code path from there.  **However, I did add a new commit that mirrors the behavior of writing to console.log for some consistency with that pattern.** 
 - The "race condition window" for a duplicate requires clicking _Save Draft_ again withinin microseconds between the async response arriving and React unmounting the component. Candidly,ii feel that's a big stretch for a realistic scenario of a race condition and not in keeping with how we've handled other mounting/unmounting scenarios.


Re: Issue 2:  Code Duplication - Data Preparation Logic
`selectedTeamMembers` is destructured from `agreement` at line 122 of `AgreementEditForm.hooks.js`
```
team_members: selectedTeamMembers,
```
The only difference between the two code paths is the `?? []` null guard in `handleDraft`, which is appropriate defensive coding for the new-agreement path where `team_members` may be undefined in initial context state. There is no divergence and no data-loss risk.

josbell (approved):
Great fix for the core issue! Before merge, please consider adding scrollToTop() call in handleDraft() (line ~538, after the success alert) to match the pattern in saveAgreement() (line 407) and handleCancel() (line 606). This ensures consistent UX when navigating away.

jonnalley:
@josbell oh man. that's a great catch. I also added that convention to the `frontend/CLAUDE.md` in this PR. That's a great find.  See commit [333e6ba](https://github.com/HHS/OPRE-OPS/pull/5691/changes/333e6ba1f7649cd7dc4dc4e3aace3cf9fb134492)

Issue #5695, opened 2026-05-19
Edit & save changes bug


kimschulke:
When attempting to edit an agreement, by removing the agreement nickname, and clicking save changes, I receive an unexpected unssaved changes modal, followed by an error.

## Expected Behavior
When the user clicks save changes, the agreement should update. 
When the user clicks save changes, they should not receive an unsaved changes modal confirmation that is intended for when the user tries to navigate away without clicking save changes.

## Current Behavior
unexpected save changes/navigate away modal
unexpected error

## Possible Cause
Not sure

## Steps to Reproduce
1. sign in to staging as user demo
2. click on the agreement procurement tracker test 2
3. click edit on the agreement details page
4. edit the agreement nickname
5. click save changes
6. see unexpected modal and error

## Context
- We need edit agreement to work properly in prep for our user training session on May 26
- This agreement currently has 1 budget line in Executing Status and in step 1 of the procurement tracker

kimschulke:
https://github.com/user-attachments/assets/566530df-3b3b-40c4-b907-210bf5709be1

kimschulke:
Would be curious to understand --
- what caused this bug to occur?
- what tests can we add or modify to prevent this from happening again?

jonnalley:
@kimschulke this has been merged in and is  in dev and staging

PR #5697, opened 2026-05-19
fix: convert empty agreement nickname to null before API submission

rajohnson90:
## What changed

Fixes a bug where clearing the nickname field on an agreement and saving sends an empty string (`""`) to the backend instead of `null`. The backend uniqueness check treats `""` as a duplicate nickname value, causing a conflict error when multiple agreements have empty nicknames.

**`frontend/src/helpers/agreement.helpers.js`**
- In `cleanAgreementForApi`, converts `nick_name: ""` to `nick_name: null` before building the API payload. This ensures the backend receives `null` (no nickname) rather than an empty string when the field is cleared.

**`frontend/src/helpers/agreement.helpers.test.js`**
- Adds a `describe("cleanAgreementForApi")` block with three unit tests: empty string → null conversion, non-empty nickname preserved, and already-null nickname passes through unchanged.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5695

## How to test

**Unit tests:**
```bash
cd frontend
npx vitest run src/helpers/agreement.helpers.test.js
```

**Manual verification:**
1. Open an existing agreement that has a nickname set
2. Clear the nickname field
3. Save the agreement
4. Verify the save succeeds without a duplicate-nickname error
5. Confirm the nickname now displays as empty/blank on the agreement detail page
6. Repeat on a second agreement — both should save without conflict

## A11y impact

## Screenshots

N/A — backend payload fix, no visual changes.

## Links

N/A

### DIFF
--- a/frontend/src/helpers/agreement.helpers.js
+++ b/frontend/src/helpers/agreement.helpers.js
@@ -324,9 +324,14 @@ export const cleanAgreementForApi = (data) => {
         "updated_on"
     ];
 
+    const cleanData = omit(data, fieldsToRemove);
+    if (cleanData.nick_name === "") {
+        cleanData.nick_name = null;
+    }
+
     return {
         id: data.id,
-        cleanData: omit(data, fieldsToRemove)
+        cleanData
     };
 };
 
--- a/frontend/src/helpers/agreement.helpers.test.js
+++ b/frontend/src/helpers/agreement.helpers.test.js
@@ -1,6 +1,7 @@
 import {
     calculateAgreementTotal,
     calculateFeeTotal,
+    cleanAgreementForApi,
     getProcurementShopSubTotal,
     getProcurementShopFees,
     getAgreementType,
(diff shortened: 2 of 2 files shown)


johndeange (approved):
## Code Review

**Correctness: Good**
- The fix is correctly placed in `cleanAgreementForApi` — the single gateway for agreement payloads.
- The logic (`if (cleanData.nick_name === "") → null`) is straightforward and handles the exact case.
- `omit()` returns a new object, so the mutation is safe.

**Tests: Good**
- Covers all three cases: empty string → null, non-empty preserved, null passthrough.
- Tests are focused and readable.

### Optional Suggestion

Consider whether whitespace-only strings (`"   "`) should also coerce to null. A trim-then-check would be more robust:

```javascript
if (!cleanData.nick_name?.trim()) {
    cleanData.nick_name = null;
}
```

This also handles the empty string case. Not blocking — just a product decision for the team.

### Backend Follow-up (Non-blocking)

The backend could also coerce `""` to `None` in the Agreement model/schema validation as defense-in-depth, in case another client or future endpoint sends `""`. Could be a follow-up ticket.

LGTM — minimal, well-scoped, and well-tested fix. ✅

Santi-3rd (approved):

weimiao67 (approved):

Issue #3371, opened 2025-01-27
canHistory pagination should return error when going over CANHistoryItem count


fpigeonjr:
## Expected Behavior
canHistory pagination should return error when going over CANHistoryItem count when offset is greater than 0

## Current Behavior
canHistory pagination does not return error when going over CANHistoryItem count

## Steps to Reproduce

1.
2.
3.
4.

PR #5699, opened 2026-05-19
feat: standardize CAN history API response with pagination metadata

johndeange:
## What changed

Aligns the `GET /api/v1/cans/{id}/history/` endpoint with the established pagination pattern used by other list endpoints. The response changes from a bare JSON array to a wrapped object with `count`, `limit`, and `offset` metadata, enabling the frontend to reliably detect when it has exhausted paginated data.

**`backend/ops_api/ops/services/can_history.py`**
- Service now returns a `(results, metadata)` tuple. Adds a count query (using the same filters) before applying offset/limit.

**`backend/ops_api/ops/resources/can_history.py`**
- Unpacks the tuple and wraps the response as `{"data": [...], "count": N, "limit": X, "offset": Y}`.

**`backend/ops_api/tests/ops/can_history/test_can_history.py`**
- All service-level tests updated to expect the tuple return shape. API-level tests verify the wrapped response structure. New test for offset-beyond-total verifying `count > 0` with empty `data`.

**`frontend/src/api/opsAPI.js`**
- Adds `transformResponse` to `getCanHistory` endpoint extracting `{items, count}` from the wrapped response.

**`frontend/src/components/CANs/CANHistoryPanel/CANHistoryPanel.jsx`**
- Uses `count` metadata for scroll termination (`offset + limit >= count`) instead of relying on empty-array detection.

**`frontend/src/components/CANs/CANHistoryPanel/CANHistoryPanel.test.jsx`** (new)
- 6 unit tests covering: rendering items, empty state, scroll stop via count, loading state, error state, and correct query params.

## Issue

https://github.com/HHS/OPRE-OPS/issues/3371

## How to test

**Backend tests:**
```bash
cd backend/ops_api
pipenv run pytest tests/ops/can_history/ -v
```

**Frontend tests:**
```bash
cd frontend
bun run test --watch=false src/components/CANs/CANHistoryPanel/CANHistoryPanel.test.jsx
```

**Manual verification:**
1. Start the full stack: `docker compose up --build`
2. Navigate to a CAN detail page with history
3. Verify history items load and infinite scroll stops when all items are fetched
4. Switch fiscal years — verify history resets and loads correctly for the new year
5. Verify a CAN with no history shows "No History"

## A11y impact

## Screenshots

N/A — no visual changes (response shape change is transparent to the user).

## Links

N/A


jonnalley:
@johndeange I think this might be a BS finding from Opus??  but sharing just in case. LGTM otherwise

## Finding: `offset` Not Reset on Fiscal Year Change in `stopped` Check

**Location**: `frontend/src/components/CANs/CANHistoryPanel/CANHistoryPanel.jsx:54-56`
**Category**: Potential Bug

**Description**:
When the fiscal year changes, the first `useEffect` (line 41-45 in current code) resets `offset` to 0 and `stopped` to false. However, the second `useEffect` depends on `offset` in its dependency array. There's a potential timing issue: React may batch state updates from the fiscal year reset, but the `canHistoryResponse` from the *previous* fiscal year query could still be in RTK Query's cache when the second effect runs, causing `offset + limit >= canHistoryResponse.count` to evaluate against stale data and prematurely set `stopped = true`.

RTK Query should re-fetch when `fiscalYear` changes (since it's part of the query params), so in practice the `isLoading` state should prevent the stop logic from firing. But the dependency chain is subtle.

**Recommended Fix**:
Consider guarding the stop condition on `!isLoading && !isFetching` to ensure you're evaluating fresh data:

```javascript
if (!isLoading && !isFetching && canHistoryResponse) {
    if (offset + limit >= canHistoryResponse.count) {
        setStopped(true);
    }
}
```

jonnalley:
@johndeange Oh, also, do you need to update openapi.yml ? The `CANHistoryList` schema still says `type: array` but the response is now a wrapped object. ?

weimiao67 (approved):

josbell (approved):

Issue #5692, opened 2026-05-18
Fix misaligned status column


kimschulke:
The status column on all data/tables is misaligned.

## Expected Behavior
The status tag should always be center aligned with the row data.
Please make sure it is consistent on every single page in OPS.

## Current Behavior
The status tag is not center aligned with the row data, its touching the bottom of the row line.

## Possible Cause
:( 

## Steps to Reproduce
1. go to any page with a status column

## Detailed Description

<img width="1521" height="416" alt="Image" src="https://github.com/user-attachments/assets/890f6efb-e0a2-422c-ac43-c2e8f12ef8d7" />

weimiao67:
I think I know the cause of the issue, and I will fix it.

kimschulke:
Thank you so much for picking this up @weimiao67 ! Would be awesome to have this resolved before the demo on thursday ✨

weimiao67:
@kimschulke the fix has been deployed to Dev and Staging.

npage123:
It looks to be fixed but seeing some inconsistencies with some status tags. It looks like the Draft status tag is a little more centered than the Planned one on the last page only which would be preferred for all. But per our discussion earlier this week between Wei, Andressa and I, moving the statuses to be more centered than they are now would also make the rows text to be uncentered. Wei, just wondering if you have any ideas on why the last page Draft statuses are looking more centered?

<img width="300" height="710" alt="Image" src="https://github.com/user-attachments/assets/a29fee2b-e2ae-44ae-9c55-a2a7c628260b" />

weimiao67:
@npage123 It is because when the agreement name is a clickable Link, the cell is rendered with a block element(div) inside a text element (a). This causes the entire table row vertical alignment to be lifted up a little bit. The last page 3 Draft agreements have a TBD agreement name, and it is not a link, so the table row vertical alignment is normal. I have tried different approaches to fix it, and the current layout is the best I can do.

PR #5702, opened 2026-05-19
fix: fix BLI row alignment issue caused by tooltip

weimiao67:
## What changed

- Replaced the character-count threshold in `TextClip` with a `ResizeObserver`-driven check that measures actual overflow before showing the USWDS tooltip, so the tooltip only appears when the rendered text is truly clipped.
- Managed the USWDS tooltip imperatively (no `usa-tooltip` class or `title` in JSX). USWDS mutates both on init, so re-asserting them on every render previously caused stacked wrappers and empty tooltip bodies — and produced the BLI row misalignment.
- Removed the now-unused `tooltipThreshold` prop from all `TextClip` callers (`AgreementTableRow`, `AllBLIRow`, `CANBudgetLineTableRow`, `ProjectSpendingAgreementRow`, `ProcurementDetailsTableRow`).
- Split the `padding-bottom: 0 !important` rule out of `.verticalAlignTop` into a new `.noPaddingBottom` class. `AllBLIRow` and `CANBudgetLineTableRow` opt in only when the row has an agreement link, so non-link rows keep their natural padding.
- Added a `"SC"` column-width rule (9rem) to `Table.jsx` so the Service Component column matches the design.
- Added a global `ResizeObserver` mock to `frontend/src/tests/setupTests.jsx` (jsdom 29 does not provide one). Without this, every test that mounts a component using `TextClip` would throw `ReferenceError: ResizeObserver is not defined`.

## Issue

#5692 

## How to test

1. Open **Budget Lines** (All Budget Lines table). Confirm:
   - Rows align cleanly with no extra padding under the agreement-name link.
   - Hovering a long agreement name shows the USWDS tooltip; hovering a short name that fits in one line shows no tooltip.
   - No stacked / duplicate tooltip wrappers appear in the DOM after multiple hovers.
1. Open a **CAN detail page** and check the Budget Lines table behaves the same way (alignment + overflow-only tooltip).
1. Open the **Agreements list**, **Project Spending** table, and **Procurement Dashboard details** table; confirm long names truncate to the configured `maxLines` and the tooltip only appears when text actually overflows.
1. Resize the browser window — the tooltip enables/disables as the column width changes (ResizeObserver path).
1. Run `bun run test --watch=false` from `frontend/` and confirm the suite passes (in particular `AllBLIRow`, `CANBudgetLineTableRow`, `ProjectSpendingAgreementRow`, `ProcurementDetailsTableRow`, `AgreementTableRow`).

## A11y impact

The USWDS tooltip is still attached when text overflows, so screen-reader/keyboard users continue to get the full text via `aria-describedby` exactly as before.


johndeange (approved):
## PR #5702 Review: Fix BLI row alignment issue caused by tooltip

### Strengths

- **Root cause fix**: The imperative USWDS tooltip management correctly addresses the conflict between USWDS's DOM-mutating behavior and React's declarative rendering model.
- **Responsive**: Using `ResizeObserver` instead of a static character threshold means the tooltip adapts to column width changes — much more robust.
- **Clean effect lifecycle**: Both effects properly clean up on unmount/dependency change. The functional `setIsOverflowing` update avoids unnecessary re-renders.
- **Good separation**: Splitting `noPaddingBottom` from `verticalAlignTop` makes the CSS more composable.

### Minor Suggestions (non-blocking)

1. **Tooltip body element leak**: When `isOverflowing` flips `true` → `false`, verify that `tooltip.off(el)` removes the sibling `.usa-tooltip__body` element USWDS creates. If it doesn't, consider explicit removal in the cleanup function.

2. **Table.jsx column width**: The "Type" and "SC" blocks are identical — could consolidate (not a blocker).

3. **PropTypes removed**: The component drops PropTypes entirely. Minor DX regression but not blocking.

4. **`+ 1` overflow tolerance**: A one-line comment explaining the sub-pixel rounding tolerance would help future readers:
   ```javascript
   const overflowing = el.scrollWidth > el.clientWidth + 1 || ...
   ```

5. **`noPaddingBottom` conditional application**: Previously all rows had `padding-bottom: 0`; now only link rows do. Visually verify non-link rows look correct with restored padding.

### Verdict

Solid fix that addresses the root cause properly. Approved.

Santi-3rd (approved):

Issue #5367, opened 2026-03-25
backend: define and expose project history for project detail page


fpigeonjr:
### Goals
* Define the backend contract needed to power the Project Details History section.
* Expose project history data in a frontend-consumable format so the Project Details page can replace the current placeholder with API-backed history.

### Tasks
- [ ] Confirm the expected behavior for project history on the detail page, including which events should appear.
- [ ] Define the response contract for project history, including event shape, actor display, timestamps, and descriptions.
- [ ] Determine whether history should be included on `GET /projects/:id` or served from a dedicated endpoint.
- [ ] Implement the backend changes needed to expose project history.
- [ ] Add or update backend tests covering the final history contract.
- [ ] Add or update sample data if needed so the response is exercised in local/dev environments.
- [ ] Document the final response contract for the frontend.

### Additional Context
* This is a follow-up to #668.
* The current Project Details page in the frontend still renders `History coming soon.` in `frontend/src/pages/projects/detail/ProjectDetailsView.jsx`.
* The approved UI includes a History section, but the current project detail API work appears focused on project metadata rather than a defined audit/history contract.
* This should be tracked separately from missing metadata fields like portfolios or ownership contacts because it may require different backend architecture or endpoint design.

### Resources
* Parent story: #668
* Frontend detail view: `frontend/src/pages/projects/detail/ProjectDetailsView.jsx`
* Parent UX/story context: #4210, #668

weimiao67:
@kimschulke @andressacurvelo @npage123 The project history is ready for review/test in Staging.

npage123:
Reviewed in staging and all looks good.

https://github.com/user-attachments/assets/a2a31d30-0edc-4470-9c35-d08cb3de70fe

PR #5724, opened 2026-05-26
feat: add project history panel and API

weimiao67:
## What changed

Adds a project change-history feature end to end:

- **Backend data layer**: new `ProjectHistory` model + alembic migration with the `projecthistorytype` enum and the `project_id_record` index. `project_id` is FK `SET NULL` so history survives project deletion via `project_id_record`.
- **Backend API**: new `GET /projects/{id}/history/` endpoint (paginated, sortable). The PATCH path now diffs the project via `generate_project_events_update` (skipping derived/computed fields and surfacing team-leader add/remove). `CREATE_PROJECT` and `UPDATE_PROJECT` are subscribed on the message bus to drive the trigger that builds history rows. Reads require `Permission.HISTORY` to match the agreement-history pattern.
- **Frontend**: new `ProjectHistoryPanel` rendered on the project detail page in place of the History placeholder. Uses the existing `LogItem` and `InfiniteScroll` components, deduplicates items across pages, and stops paging once `offset + limit >= count`.

## Issue

#5367

## How to test

1. `docker compose up --build` and `cd backend && alembic upgrade head` to apply the new migration.
2. Open a project at `http://localhost:3000/projects/{id}` — the History section under Description renders the new panel ("No History" if empty).
3. Edit the project (title, nickname, description, project type) and reload the detail page — corresponding history entries should appear, newest first.
4. Hit the API directly: `GET /api/v1/projects/{id}/history/?limit=10&offset=0` returns `{ data, count, limit, offset }`. Try `?sort_asc=true`, `?limit=0` (400), and an unauthenticated request (401).
5. Confirm a user without `Permission.HISTORY` gets 403 from the endpoint.
6. All tests pass

## A11y impact

### DIFF
--- a/backend/alembic/versions/2026_05_26_1000-d4f5a6b7c8e9_add_project_history_table.py
+++ b/backend/alembic/versions/2026_05_26_1000-d4f5a6b7c8e9_add_project_history_table.py
@@ -1,0 +1,71 @@
+"""add project_history table
+
+Revision ID: d4f5a6b7c8e9
+Revises: e9f8a7b6c5d4
+Create Date: 2026-05-26 10:00:00.000000+00:00
+
+"""
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = 'd4f5a6b7c8e9'
+down_revision: Union[str, None] = 'e9f8a7b6c5d4'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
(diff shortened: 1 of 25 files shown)


Santi-3rd (approved):

josbell (changes requested):
On manual testing im not able to edit project description.

josbell (approved):
Upon further testing I'm pretty sure the issue was that the current project data exceeds the current description textarea char limit. This means there's no code issues but it does suggest that if the current data is realistic we might be overly restricting for that field. I will approve and let you follow up with UX / business as you think fit.

rajohnson90 (approved):
Looks good to me

Issue #5775, opened 2026-06-04
Review react-slider status and replace if needed


jonnalley:
### Goals
* Have a react-19+ compatible solution in place for the frontend dependency `react-slider`.

### Tasks
- [ ] Assess viability of using something else instead of `react-slider` (i.e. rc-slider) if the landscape of the dependencies and/or react compatibility hasn't changed. 
- [ ] Test out other options and once something compliant with react 19+ is in place, remove the `--legacy-peer-deps`

### Additional Context
* `react-slider` hasn't been updated in 3 years at present. 
* This came up in our attempt to remove the `--legacy-peer-deps ` from `.github/workflows/security_snyk.yml`

PR #5776, opened 2026-06-04
Replace react-currency-format with formatCurrency helper

jonnalley:
## What changed

This PR replaces the abandoned [`react-currency-format`](https://www.npmjs.com/package/react-currency-format) package across the frontend with two alternatives, depending on the use case.

### Why this matters

`react-currency-format@1.1.0` declares a peer dependency on `react@^15 || ^16 || ^17` — it does **not** support React 18 or 19. The OPS frontend has been on React 19 for a while, and this peer-dep mismatch has had real consequences:

1. The Snyk SCA workflow (`.github/workflows/security_snyk.yml`) had to add `--legacy-peer-deps` to its `npm install` step to even generate a lockfile for scanning. That flag suppresses the resolver and accepts a "potentially broken" tree — the same workaround we'd otherwise want to avoid in CI.
2. The library is unmaintained (last release 2020). It will not gain React 18+ support. As long as it's a dependency, the project carries the noise.
3. Inside `CurrencyInput.jsx` we already had a hand-rolled workaround (`displayValue` state + a skip-sync ref) to compensate for the controlled-component echo race during decimal typing. We initially attempted to drop that workaround; CI revealed the race is intrinsic to the controlled-component pattern (parent stores a parsed float, library reformats on every prop echo), so the workaround had to be reinstated and refined. See "Post-CI fixes" below.

### The approach

`react-currency-input-field` is a maintained replacement that supports React 19 (as of `3.10.0`) — but it's an **input** component only. It has no `displayType="text"` mode for read-only formatting, which is how 32 of the 33 usages in this codebase used the old library.

So we split the migration in two:

- **32 display-only usages** → a new `formatCurrency()` helper using `Intl.NumberFormat`. This is a built-in browser API, no dependency required.
- **1 form-input usage** (`CurrencyInput.jsx`) → rewritten on `react-currency-input-field@3.10.0`.

### Breakdown by area

**`frontend/src/helpers/currencyFormat.helpers.js`** — Added `formatCurrency(value)`. Returns `"$0"` for zero / null / undefined / NaN / Infinity / unparseable strings, and `"$1,234.50"` for finite non-zero numeric input (numeric strings are coerced via `Number()`). Matches the previous `<CurrencyFormat displayType=\"text\" prefix=\"$\" thousandSeparator decimalScale={getDecimalScale(amount)} fixedDecimalScale />` output for the inputs that were valid before; pathological inputs that the old library masked silently now fall back to `$0`. The unused `getDecimalScale` helper has been removed (no remaining call sites in `src/`).

**31 component/page files** — Mechanical swap from `<CurrencyFormat>` to `{formatCurrency(value)}`. Net **−530 lines**. Examples: `BLIRow.jsx`, `BudgetCard.jsx`, `AgreementTableRow.jsx`, `ProjectFundingCANsTable.jsx`, etc.

**5 files with `prefix=\"$ \"` (dollar + space)** — These callers want a literal space between `$` and the number (`"$ 100"` vs `"$100"`). Since `formatCurrency` doesn't emit the space, these files use an inline template-literal pattern that preserves the original output exactly:
```jsx
{`$ ${new Intl.NumberFormat("en-US", { minimumFractionDigits: N, maximumFractionDigits: N }).format(value)}`}
```
Affected files: `BLIStatusSummaryCard.jsx`, `CANFYBudgetRangeSlider.jsx`, `BudgetRangeSlider.jsx`, `PortfolioFYBudgetRangeSlider.jsx`, `pages/cans/detail/CanFunding.jsx`.

**`CurrencyWithSmallCents.jsx`** — Rewritten with plain JSX + `Intl.NumberFormat`. The component renders dollars and cents in two separate spans (so the cents can be styled smaller); the structure and class names are unchanged. **Negative amounts** are rendered as `-$ 1,234` (sign before the dollar prefix), matching the previous `react-currency-format` output that callers like `BigBudgetCard` and Cypress assertions in `canDetail.cy.js` rely on. (The naive `Intl.NumberFormat` output would be `$ -1,234`; we render the sign explicitly and format `Math.abs(...)`.) The sign is derived from the raw `amount` rather than the truncated dollar string so sub-dollar negatives like `-0.50` render correctly as `-$ 0.50` (where `parseInt(-0.5)` would have collapsed to `0` and lost the sign). Non-finite inputs coerce to `0` at the boundary.

**`CurrencyInput.jsx`** — Rewritten on `react-currency-input-field@3.10.0`. Keeps a local `displayValue` string and a last-emitted-value ref so the parent's float echo (`5.` → `5` → `"5"`) doesn't overwrite the in-progress display string mid-typing, while genuine parent-driven changes (e.g. `resetForm` clearing `enteredAmount` between BLI submissions) still flow through. The external props interface (`name`, `label`, `onChange`, `pending`, `messages`, `value`, `className`, `setEnteredAmount`, `placeholder`) is unchanged, so all callers (`BudgetLinesForm`, `CANBudgetForm`, `CanFundingReceivedForm`) continue to work without modification.

**`CanCard.jsx` (scope expansion, but necessary)** — Fixed a latent string-concatenation bug in `spendingAmount`. The CAN funding API returns the funding fields as numeric strings (`"1000000.00"`); the old `+` operator was concatenating instead of adding them, producing `"1000000.002000000.00"` → `NaN`. The old `<CurrencyFormat>` rendered the NaN silently/invisibly, masking the bug. `formatCurrency(NaN)` renders `"$NaN"`, which surfaced it. Wrapping each addend in `Number(... || 0)` produces the actual sum. The corresponding test was updated from `$1,000,000.00` to `$3,000,000.00` (the correct sum of 1M + 2M + 0).

**Test updates** — 8 test files were updated to match the new rendering. Two patterns:
- `CurrencyWithSmallCents` now renders dollar and cent portions in separate `<span>` nodes, so `getByText(\"$1,234.56\")` no longer matches; tests use regex matchers or `textContent` checks.
- `formatCurrency(0)` returns `\"$0\"` (no decimals) where some tests previously asserted `\"$0.00\"`. This matches the original `decimalScale={getDecimalScale(amount)}` behavior — assertions were corrected.

Also adds two new direct test files: `currencyFormat.helpers.test.js` (8 cases covering zero, null/undefined, decimals, negatives, rounding, numeric strings, and non-finite inputs) and `CurrencyWithSmallCents.test.jsx` (7 cases including the sub-dollar-negative regression — none existed before this PR).

**`package.json` / `bun.lock`** — `react-currency-format@1.1.0` removed; `react-currency-input-field@3.10.0` added.

### Post-CI fixes

The initial commit (`13f1869`) shipped clean unit tests but failed several CI checks. Four follow-up commits address what CI surfaced:

- **`d2807e6` — `fix(frontend): restore CurrencyInput display-value sync for decimal entry`**
  Cypress test `createAgreement.cy.js > "allows entering a decimal budget line amount"` typed `500.75` and asserted the input retained `500.75`; actual was `50,075`. Root cause: `CurrencyInput` was fully controlled with the parent storing a parsed float; the library re-formatted the float on each prop echo, dropping the decimal mid-typing. Reintroduced a local `displayValue` string so the typed text isn't clobbered.

- **`f1a1d75` — `style(frontend): apply prettier formatting to currency-format migration`**
  23 files in the original commit were not run through Prettier; the Linting CI job blocked. Mechanical reformat, no behavior change.

- **`61fe405` — `fix(frontend): allow parent reset to flow through CurrencyInput`**
  After `d2807e6`, `createAgreement.cy.js > "can create an NON-SEVERABLE agreement"` started failing: the test types two BLIs in sequence (`1000000`, then `2000000`) and expects both totals on the FY summary card. The first commit's skip-flag was suppressing the parent's reset of `enteredAmount` between submissions, so the second BLI's typed amount appended to the first's stale display value. Replaced the skip-flag with a `lastEmittedRef` that stores both the raw-string and float forms of the most recent emit; an incoming `value` is treated as an "echo" only when it matches one of those, so a parent reset to `""` (or any other independent change) still flows through. Includes a regression test.

- **`f3cec66` — `fix(frontend): place minus sign before $ in CurrencyWithSmallCents`**
  `canDetail.cy.js` asserts the over-budget summary card contains `-$ 120,797,640.00`. The plain-JSX rewrite produced `$ -120,797,640.00` (sign after the prefix) because `Intl.NumberFormat` placed the minus inside the formatted number. The old `react-currency-format` render put the sign first; preserved that ordering by detecting negatives explicitly and prefixing `-$ ` to the absolute-value-formatted number.

Three additional commits land review feedback from a multi-agent code review:

- **`c38536e6` — `fix(frontend): preserve sign for sub-dollar negatives in CurrencyWithSmallCents`**
  `f3cec66` derived `isNegative` from `parseInt(dollarValue)`, which loses the sign for amounts in `(-1, 0)` (e.g. `parseInt(-0.5) === 0`). `-$0.50` rendered as `$ 0.50` — a regression vs. the old library's behavior. Derived the sign from the raw `amount` instead, and coerced non-finite inputs to `0` at the boundary. Added a new `CurrencyWithSmallCents.test.jsx` (none existed) covering positive, negative, zero, sub-dollar-negative, and NaN cases.

- **`55618f47` — `fix(frontend): coerce non-finite inputs to 0 in formatCurrency`**
  `formatCurrency` previously only guarded `null`/`undefined` with `?? 0`, so `NaN`, `Infinity`, and unparseable strings flowed through `Intl.NumberFormat` and rendered as `"$NaN"` / `"$∞"`. Now coerces with `Number()` and falls back to `0` when the result isn't finite. Removed the now-unused `getDecimalScale` export and a stale comment in `PortfolioTableRow.test.jsx` that referenced it. Added `currencyFormat.helpers.test.js` for direct coverage of the contract.

- **`ff3edd6a` — `docs(frontend): explain Number coercion in CanCard spendingAmount`**
  One-liner explaining why the explicit `Number(... || 0)` matters (the funding API returns numeric strings; `+` would concatenate them) so future maintainers don't silently revert it.

CI is now green on the resulting `ff3edd6a` HEAD.

### Plan deviations worth flagging to reviewers

 **Did NOT remove `--legacy-peer-deps` from `.github/workflows/security_snyk.yml`** even though that was an  original goal. Reason: while testing, the simulated `npm install --package-lock-only --ignore-scripts` (i.e. what Snyk does) revealed that `react-slider@2.0.6` *also* declares `react@^16 || ^17 || ^18` and has no React-19 release. So even with `react-currency-format` gone, the resolver still fails. The Snyk workflow needs `--legacy-peer-deps` until `react-slider` is replaced. [That's a separate follow-up](https://github.com/HHS/OPRE-OPS/issues/5775)...

## Issue

N/A, found while doing other work.

## How to test

```bash
cd frontend
bun install --frozen-lockfile
bun run lint        # should pass with the same 1 pre-existing warning as main
bun run test --watch=false   # 3332 tests should pass
bun run build       # production build should succeed
```

Manual UI verification (any of these pages, while running the full Docker stack):
1. **CAN detail page** (`/cans/<id>`) — currency cells in the funding cards, funding-received table, and budget-line table should look identical to main. Spot-check the "Previous FYs Carry Forward" line — it should display as `$ 1,234.56` (with leading space). On an over-budget CAN, the `BigBudgetCard` should display the negative remaining budget as `-$ 1,234.56` (sign before `$`); a sub-dollar over-budget condition (e.g. `-$ 0.50`) should also keep its sign rather than collapsing to `$ 0.50`.
2. **Budget Lines table** anywhere it's rendered (agreement detail, budget line items list) — totals/amounts should format identically.
3. **Sliders** (`/portfolios` filter, `/cans` FY filter) — values shown above the slider should still have a leading space (`$ 10,000,000`).
4. **`CurrencyInput`** form fields — visible in agreement creation and budget-line forms. Type `5.50` and verify (a) the input accepts and displays it, (b) it accepts a trailing decimal mid-entry (`5.`), (c) clearing the field works, (d) submitting a BLI and entering a second one starts from an empty input (the `61fe405` fix).
5. **`CurrencyWithSmallCents`** displays — visible on the BudgetCard / BigBudgetCard. Cents should render in a smaller superscript-style span; total visible text should match main.
6. Verify the **CanCard "Spending" total** on a CAN with non-zero planned/in-execution/obligated funding now shows the correct sum (this is the bug fix; on main it was rendering `NaN` invisibly).

CI should run the same suite plus the Cypress E2E pack.

## A11y impact

(All changes are display/rendering; no semantic or interactive changes were introduced. The form input keeps the same label/role/error-message structure. Reviewer should confirm the unchecked option above is correct.)

## Screenshots

N/A — visible output is unchanged from main with one exception: `CanCard` "Spending" totals now display the actual sum where they previously rendered `NaN`. Happy to add before/after screenshots if requested.

## Links

- Replaced package: https://www.npmjs.com/package/react-currency-format
- New input package: https://www.npmjs.com/package/react-currency-input-field
- React 19 support added in `react-currency-input-field@3.10.0` (verified via `npm view react-currency-input-field@3.10.0 peerDependencies`)


rajohnson90 (approved):
All seems to look good to me. AI review also said it's okay to approve

Santi-3rd (approved):

weimiao67 (approved):
Tested manually, no issue found.

Issue #5725, opened 2026-05-26
Change Request list is unpaginated


rajohnson90:
The change requests API endpoint implements pagination with a default page size of 10 and an offset of 0. The front end (both in the UI and the RTK query) do not deal with the pagination parameters at all and as a result the only thing a user will ever be able to retrieve is the first 10 change requests. 

## Expected Behavior
A DD should be able to view all the change requests under their purview, whether it's the 10 oldest or anything newer.

## Current Behavior
In the case where there are 11 or more 

## Steps to Reproduce
1. Create 11 change requests for a given agreement
2. Check the change request tab and see there are only 10

## Context
Discovered while determining why Hillary is unable to see a specific change request.

## Possible Implementation
Implement pagination in front end for change requests

jonnalley:
@kimschulke @andressacurvelo @npage123 this has been deployed out to all environments. not sure if we have any scenarios to test it.

npage123:
Tested in staging and verified I'm able to see 11 change requests. However, I'm just seeing the total number of requests, not paginated if that was the want?

https://github.com/user-attachments/assets/bb3030ea-df08-4edb-88bc-ee051c5dbc13

npage123:
@rajohnson90 I tested again in staging and the pagination is showing up. However, I did notice that the first page is showing 11 change requests and the second is showing 3, which add up to 14 in total. The number in the For Review red circle shows 13. I wonder if the Pre-Award card is both attributing to the innaccurate number and the pagination no recognizing it still?

https://github.com/user-attachments/assets/68e938b3-1d85-4e11-8273-70e6f645f640

kimschulke:
ready to test again - thanks Reed!

npage123:
Tested in staging and looks good! Thank you!

https://github.com/user-attachments/assets/da82c2db-f447-4391-aadf-afd047813b16

PR #5798, opened 2026-06-10
feat(change-requests): unify approval types into single paginated list

rajohnson90:
## What changed

Unified the Change Requests approval list so that all card types — budget requisitions, pre-award approvals, and change-request cards — are merged into a single client-side paginated list sorted by date descending.

**`frontend/src/components/ChangeRequests/ChangeRequestsList/ChangeRequestsList.jsx`**
- Changed the change-requests API call to fetch all results at once (limit 1000, offset 0) rather than fetching one page at a time from the backend, avoiding the previous issue where pre-award/requisition cards were excluded from the page count.
- Each `ChangeRequest` object with multiple change types (proc shop, budget, status) is now expanded into one tagged entry per card type so pagination counts are accurate (1 entry = 1 rendered card).
- Pre-award approvals and budget requisitions are similarly tagged and merged into a single `allItems` array.
- The merged array is sorted by `approval_requested_date` / `created_on` descending; client-side slicing by `PAGE_SIZE` (10) drives the existing `PaginationNav`.

**`frontend/cypress/e2e/changeRequestsPagination.cy.js`**
- Extended the E2E test suite with setup/teardown for procurement-tracker pre-award steps on agreements 19 and 20.
- Added a `setupPreAwardStep` helper that drives an agreement's tracker steps to COMPLETED through step 4, then calls a callback so individual tests can set `approval_requested` or `approval_status`.
- New test: verifies that page 1 shows exactly 10 cards and that a pre-award review card appears somewhere in the paginated list (as division-director).
- New test: verifies that at least one budget-team requisition card appears when logged in as budget-team.
- After-hook cleanup now declines the pending pre-award approval (agreement 19) and submits the requisition (agreement 20) to leave shared data in a neutral state.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5725

## How to test

1. Start the full stack: `docker compose up --build`
2. Log in as **division-director** and navigate to `/agreements?filter=change-requests`. Confirm:
   - Page 1 shows 10 cards.
   - Pagination nav is visible.
   - Pre-award approval cards appear somewhere in the paginated list (not just at the end, sorted by date).
3. Log in as **budget-team** and navigate to `/agreements?filter=change-requests`. Confirm:
   - Budget requisition cards are visible in the list.
4. Run the E2E test directly:
   ```bash
   cd frontend
   npx cypress run --spec "cypress/e2e/changeRequestsPagination.cy.js"
   ```

## A11y impact

## Screenshots

N/A — no visual layout changes; only the sorting/pagination logic was modified.

## Links

N/A


frontend/src/components/ChangeRequests/ChangeRequestsList/ChangeRequestsList.jsx
```diff
@@ -89,106 +89,127 @@ function ChangeRequestsList({ handleReviewChangeRequest }) {
...
+        return cards;
+    });
+
+    const allItems = [...taggedBudgetRequisitions, ...taggedPreAwardApprovals, ...taggedChangeRequests].sort((a, b) =>
+        b._sortDate > a._sortDate ? 1 : b._sortDate < a._sortDate ? -1 : b.item.id - a.item.id
+    );
 
-    return hasChangeRequests || hasPreAwardApprovals || hasBudgetRequisitions ? (
+    const totalPages = Math.ceil(allItems.length / PAGE_SIZE);
+    const pageItems = allItems.slice((currentPage - 1) * PAGE_SIZE, currentPage * PAGE_SIZE);
```
weimiao67, line 115: `currentPage` is not bounded against `totalPages`. If `allItems` shrinks (e.g., refetch after an approval), and the user was on page 3 when totalPages drops to 2, pageItems.slice(...) returns [] — the user sees an empty page with no nav indication. Add a useEffect to clamp currentPage to `Math.max(1, totalPages)`.

frontend/cypress/e2e/changeRequestsPagination.cy.js
```diff
@@ -113,40 +267,55 @@ describe("Change Requests List - Pagination", () => {
...
 
-    it("shows 10 change requests on page 1 and 1 change request on page 2", () => {
-        // Log in as division director (Dave Director, division 4)
+    it("shows 10 items on page 1, shows pagination, and a pre-award approval card appears somewhere in the list", () => {
         testLogin("division-director");
         cy.visit("/agreements?filter=change-requests");
 
-        // Intercept the first page request
-        cy.intercept({ method: "GET", url: "**/change-requests/*", query: { limit: "10", offset: "0" } }).as("page1");
+        cy.intercept({ method: "GET", url: "**/change-requests/**" }).as("changeRequests");
```
weimiao67, line 276: `cy.intercept` is registered after `cy.visit`. Cypress fires the network calls during page load, often before the intercept handlers register; the `cy.wait("@changeRequests")` may time out or catch only re-fetches. Standard pattern is intercept → visit → wait.

frontend/src/components/ChangeRequests/ChangeRequestsList/ChangeRequestList.test.jsx
weimiao67, line 1: Coverage of new logic in ChangeRequestsList.test.jsx is unchanged — only the constant 10 → 1000 was updated. The unit test still mocks empty arrays for pre-award/requisitions, so the new tagging/sorting/expansion code is not unit-tested at all. Worth adding cases for: a CR with multiple has_*_change flags rendering multiple cards; mixed pre-award + CR sort order; clamp behavior when currentPage > totalPages.

weimiao67 (approved):

josbell (changes requested):
Found several issues that should be addressed before merging:

**Critical Issues**

1. **Missing optional chaining for `created_by_user` (Line 184)**
   - `ChangeRequestsList.jsx:184` lacks optional chaining: `cr.created_by_user.display_name`
   - Lines 199 and 213 correctly use `cr.created_by_user?.display_name`
   - Backend schema allows null (`allow_none=True` in `change_requests.py:17`)
   - Will cause TypeError crash if `created_by_user` is null (deleted users, missing records)
   - Also affects `ReviewChangeRequestAccordion.jsx:45-46`

2. **Sorting tiebreaker compares IDs across different entity types (Line 103)**
   - Tiebreaker `b.item.id - a.item.id` compares IDs from different database tables
   - ProcurementTrackerStep IDs vs ChangeRequest IDs use independent sequences
   - Causes non-deterministic sort order when timestamps are identical
   - Should add type-aware secondary sort before comparing IDs

3. **Missing PropTypes**
   - Component lacks PropTypes validation despite project requirement
   - All child card components correctly define PropTypes

**Concerns**

4. **E2E test idempotency**
   - Test uses fixed agreement IDs (19, 20) with `before()`/`after()` hooks
   - Guards are present but pattern is fragile for interrupted local runs
   - Works in CI due to fresh database per run, but diverges from other E2E tests that use `beforeEach()` with unique data
   - Consider migrating to the more robust pattern used in other E2E tests

5. **Test coverage gaps**
   - No test for `created_by_user: null` case
   - No test for identical timestamps to exercise tiebreaker logic

rajohnson90:
To Josbell's comments:
1. Fixed
2. Fixed
3. I actually think Frank was moving away from using PropTypes in new work which is why I didn't retroactively include it here.
4. I know that the idempotency can be a bit off given how the tests are set up, but the tests as I have them structured now are testing important functionality. Namely, I need to be sure that ALL types of change requests will get paginated when mixed together. If you think this is critical, I can update the tests to create & delete agreements similar to 19/20. The only reason I haven't so far is that it would balloon the before & after functions.
5. I don't think created_by_user being null is actually a realistic scenario given how that value is populated. As for the same timestamp issue, I did create a new e2e test to cover that.

Santi-3rd (approved):

josbell (approved):

PR #5786, opened 2026-06-06
Remove unused mypy dependency from backend

jonnalley:
## What changed

Removes the `mypy` package from the backend `ops_api` Pipfile. mypy was added in the initial project scaffold (Oct 2022) but was never configured or executed — there is no CI step, no pre-commit hook, no nox session, and no mypy configuration file (mypy.ini or pyproject.toml section). Removing it eliminates unnecessary install time and Renovate maintenance churn.

**backend/ops_api/Pipfile** — removed `mypy = "==1.20.2"` from `[packages]`

**backend/ops_api/Pipfile.lock** — regenerated without mypy and its direct dependencies (librt, pathspec). Note: `mypy-extensions` remains as a legitimate transitive dependency of `marshmallow-dataclass` → `typing-inspect`.

**backend/ops_api/.gitignore** — removed vestigial `.mypy_cache/` entry

## Issue

N/A

## How to test

1. Verify the lockfile is consistent:
   ```bash
   cd backend/ops_api && pipenv lock --verify
   ```
2. Verify dependencies install cleanly:
   ```bash
   cd backend/ops_api && pipenv install
   ```
3. Verify the backend builds and starts:
   ```bash
   docker compose up backend --build
   ```
4. Confirm linting still passes (mypy was never part of the lint pipeline):
   ```bash
   cd backend/ops_api && pipenv run nox -s lint
   ```

## A11y impact

- [x] No accessibility-impacting changes in this PR

## Screenshots

N/A — backend dependency change only

## Definition of Done Checklist
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests updated and passed
- [x] Automated integration tests updated and passed
- [x] Automated quality tests updated and passed
- [x] Automated load tests updated and passed
- [x] Automated a11y tests updated and passed
- [x] Automated security tests updated and passed
- [ ] 90%+ Code coverage achieved
- [ ] Form validations updated

## Links

N/A

PR #5637, opened 2026-05-08
Procurement Tracker Step 6 Approval Request Flow

josbell:
## What changed

This PR implements the requester (COR) portion of Procurement Tracker Step 6 (Award) up to the point where approval is requested from the Budget Team. Budget Team approval functionality will be addressed in a subsequent PR.

**Step 6 Award Approval Workflow** enables the COR to:
1. View budget lines grouped by services component with status and CLIN information
2. Assign CLINs (Contract Line Item Numbers) 1-10 to budget line items
3. Enter vendor information (vendor selection, displays Unique Entity ID and vendor type)
4. Enter award details (Contract #, Award Amount, Award Date)
5. Add optional notes
6. Request approval from the Budget Team

### Backend Changes

**Database & Models**
- Added Step 6 database fields: `vendor_id`, `contract_number`, `award_amount`, `award_date`, `requester_notes`, `reviewer_notes`
- Added `ProcurementTrackerStepSix` model with field mappings
- Added CLIN serialization schema (`clins.py`)
- Updated budget line items schema to include CLIN relationships

**Validation**
- Required fields validation for Step 6 completion (vendor, contract #, award amount, award date)
- Approval status validation (prevents duplicate approval requests)
- CLIN and vendor validation
- Step 5 prerequisite validation (Pre-Award Approval must be completed)
- Validation suite with comprehensive test coverage

**API Endpoints**
- Step 6 field mapping in procurement tracker service
- CLIN assignment batch update endpoint

### Frontend Changes

**New Components & Pages**
- `RequestAwardApproval.jsx` - Award approval request page with form and budget line review
- `RequestAwardApproval.hooks.js` - Custom hook managing form state, validation, and submission
- `CLINSelector.jsx` - Inline CLIN assignment dropdown with fixed options 1-10
- `AgreementBLIAccordion` - Budget line items accordion with summary cards and grouped tables
- `ProcurementTrackerStepSix.jsx` - Step 6 component with disabled/completed/incomplete states

**Budget Line Display Enhancements**
- Added CLIN column to `BLIReviewTable` (Award Approval page only)
- Hover interaction to add/edit CLINs via inline selector
- Local state management with batch update on form submission
- Draft status budget lines show "N/A" and cannot be assigned CLINs
- Success toast notification when CLIN is added
- Persistent error message when non-Draft BLIs are missing CLINs
- Status-based error styling (red "TBD" for missing CLINs in non-Draft BLIs)

**Vendor Information Display**
[...]

PR #5801, opened 2026-06-11
stabilize BudgetLinesForm DatePicker against unmount + DOM accumulation

jonnalley:
## Summary

`editBudgetLineByPowerUser.cy.js` has been failing on `main` for several days, taking down CI on a long list of unrelated open PRs. Two interacting bugs were responsible:

1. **`BudgetLinesForm` was unmounting the DatePicker on every parent render.** A `const MemoizedDatePicker = React.memo(DatePicker)` was declared inside the component body, so eacfh render produced a new memoized component reference. React treated it as a different component type and unmounted the existing DatePicker — including on every keystroke into the sibling `#enteredAmount` field. That made `cy.clear()` on `#need-by-date` flake: the input was unmounted between Cypress's visibility check and the action.

2. **`DatePicker.jsx` re-initialized USWDS on every render.** The init/teardown `useEffect` had `[onChange, value, name]` as deps, and `onChange` from `BudgetLinesForm` is an inline arrow whose identity changes every render. So once the DatePicker stayed mounted (fix # 1), USWDS's `enhanceDatePicker` ran on every render. USWDS's `datePicker.off()` is a no-op (no `teardown`/`remove` defined in `datePickerEvents`), and `enhanceDatePicker` calls `datePickerEl.appendChild(calendarWrapper)` — so each render appended another `.usa-date-picker__wrapper` to the fieldset. The fieldset grew taller across renders and pushed the **Amount** field out of horizontal alignment with **CAN**.

## Changes

**`frontend/src/components/BudgetLineItems/BudgetLinesForm/BudgetLinesForm.jsx`**
- Remove the inline `React.memo(DatePicker)`; render the default-exported `<DatePicker>` directly. The default export is already wrapped in `React.memo` (`DatePicker.jsx:143`), so memoization is preserved without the per-render recreation.

**`frontend/src/components/UI/USWDS/DatePicker/DatePicker.jsx`**
- Hold `onChange` and `name` in refs and run init/teardown **once** on mount (empty dep array) so USWDS is initialized exactly once.
- Use a separate effect to sync the `value` prop into the uncontrolled USWDS input without reinitializing.

**`frontend/src/components/BudgetLineItems/BudgetLinesForm/BudgetLinesForm.test.jsx`**
- Vitest regression test asserting the DatePicker stays mounted (1 mount / 0 unmounts) across parent re-renders driven by sibling state changes.

## Why this regression appeared on `main`

The `MemoizedDatePicker` recreation has been latent for a while, but only manifested when the parent re-renders mid-Cypress-interaction. PR #5776 (`cc75a59bc`, \"refactor: replace react-currency-format with formatCurrency helper\") changed how `CurrencyInput` propagates state — `setEnteredAmount` now fires on every keystroke, so `BudgetLinesForm` re-renders on every keystroke into `#enteredAmount`. That re-render cadence intersected the unmount race in #1 and the DOM-accumulation bug in #2.

Earlier hotfix `d58789f9d` added a `setNeedByDate()` Cypress helper that waits for `should(\"be.visible\").and(\"not.be.disabled\")` before `.clear()`. The helper papered over the symptom but did not address the underlying race. We're leaving that helper in as defense-in-depth.

## Cross-PR impact

The same Cypress spec is currently red on these unrelated PRs with the same `cy.type() failed because it targeted a disabled element` symptom on `#need-by-date`:

- #5786, #5790, #5791, #5796, #5797, #5798, #5637

Re-running CI on each of those after this lands should unblock them — no rebase required.

## Verification

- [x] Wrote the unmount regression test first; verified it fails on the buggy code (3 mounts observed, expected 1) before applying the BudgetLinesForm fix — TDD red-green confirmed.
- [x] Manual visual check in local Podman env confirms Amount/CAN are realigned after the DatePicker fix.
- [x] `bun run lint` / `bun run format` clean.
- [x] **Unit tests:** all DatePicker consumers (BudgetLinesForm, ProcurementTracker steps 1–5, ServicesComponentForm, ReviewBudgetTeamRequisition, AgreementProcurementTracker) — 910/910 pass.
- [x] **Cypress:** 21/21 pass across `editBudgetLineByPowerUser.cy.js`, `createAgreement.cy.js`, `createAgreementWithValidations.cy.js`.
- [x] Pre-commit hooks (eslint, prettier, trufflehog, commitlint) all pass.

### DIFF
--- a/.claude/skills/ci-log-triage/SKILL.md
+++ b/.claude/skills/ci-log-triage/SKILL.md
@@ -1,0 +1,219 @@
+---
+name: ci-log-triage
+description: Use when CI is red, a PR is failing, you need to know if a failure is flaky or real, you want to verify whether a stated cause actually appears in CI logs before planning a fix, multiple PRs are failing for the same reason, or you need to find the commit that first broke a workflow. Covers all OPRE-OPS GitHub Actions workflows: Continuous Integration, E2E Tests, Storybook Build, Release, Deploy.
+argument-hint: "[verify <claim> | bisect | cross-pr | flake-or-real | <run-id>]"
+allowed-tools: Read, Grep, Glob, Bash, WebFetch
+disable-model-invocation: true
+---
+
+# CI Log Triage
+
+You diagnose GitHub Actions failures for OPRE-OPS. You do not fix code, run tests locally, or write files.
+
+**Companion skills** (for after diagnosis):
+- `/e2e-tests fix <spec>` — fix a Cypress failure once diagnosed
+- `/backend-tests` — fix a pytest failure
+- `/create-pr` — open a PR once a fix is ready
+
+## Iron Law
+
+**Evidence before claims. Always.**
+
+Never assert a cause without a verbatim log line that supports it. Every triage report must include the actual log snippet — not a paraphrase, not an inference from a code diff, not a prior diagnosis document. If you cannot find a matching log line, the verdict is `insufficient evidence` and the recommended next action is to re-run CI or reproduce locally.
+
+## Pre-flight Checks
+
+Run before any mode:
+
+```bash
+command -v gh > /dev/null && gh auth status 2>&1 | head -3 || echo "gh CLI: NOT AVAILABLE"
+command -v jq > /dev/null && echo "jq: OK" || echo "jq: NOT AVAILABLE"
+```
+
+If either is missing, stop and tell the user. All log-fetching requires both tools.
+
+## Choosing a Mode
+
+```dot
+digraph mode {
+    "Arguments?" [shape=diamond];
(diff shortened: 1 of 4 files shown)


weimiao67:
Opening the DatePicker breaks the alignment, it pushes the Amount field down.

Screenshot 2026-06-11 at 9 56 02 AM

jonnalley:
@weimiao67 fixed in the latest commit and tested manually. See the updated summary and changes section of the PR for what was going on. Thanks again!

> Opening the DatePicker breaks the alignment, it pushes the Amount field down.
>

weimiao67 (approved):

josbell (approved):

Issue #2098, opened 2024-03-20
Improve Edit/Resolve Errors on Send to Approval


kimschulke:
### Goals
- Improve the user workflow to edit/resolve errors during send to approval
- Currently, in dev, when a user clicks Send to Approval and sees errors to resolve, they click Edit and are taken to the Create Agreement workflow. However, this does not match Figma or the intended user path.
- When a user clicks Edit to resolve errors during Send to Approval, they should instead be taken to a single page where they can address the errors without having to go through the Create Agreement wizard.

### Tasks
- [ ] Update dev to match user workflow designed in Figma

### Additional Context
* 

### Resources
*

fpigeonjr:
the single page edit has not been built

fpigeonjr:
- Resolve page renders the agreement form and SC and Budget Lines forms
- Validation is working
- TODO: save all button logix

kimschulke:
@npage123 @andressacurvelo @kimschulke need to review edit workflows in general to see if we can find a solution that works for all edit experiences. Right now we have a few experiences:

Edit Agreement details tab and only edit agreement details
Edit SCs & BLs tab and only edit SCs & BLs
Edit during status change and edit everything on one page
OR
Edit during status change and edit in 3 step wizard


There are some dev issues/higher level of effort in regards to editing the whole agreement on one page

fpigeonjr:
@kimschulke should we move this back to needs refinement?

npage123:
I think that's a good idea @fpigeonjr. Based on how you explained that this is more difficult because of how things are structured in the backend, this may need some more thought put to it.

kimschulke:
@weimiao67 let us know if you want to collab on this at all or if you have any questions

jonnalley:
This has been deployed to prod and client already told about it

npage123:
Even though it has been released in prod I did a test in dev and all looks great!

https://github.com/user-attachments/assets/1cf19833-222c-4d3e-a6c2-b92885aaf9ae

PR #5808, opened 2026-06-14
feat: edit agreement and budget lines under review

weimiao67:
## What changed

Adds an `EditAgreementAndBudgetLines` page (and routing) so users can edit an agreement and its budget lines while the agreement is in the review state. Saving from the new page now uses a single atomic edit-bundle endpoint and returns the user to the Review Agreement page with a single success alert.

### Atomic edit-bundle endpoint (backend)
- New `POST` edit-bundle resource (`ops/resources/agreement_edit_bundle.py`) with Marshmallow schemas (`ops/schemas/agreement_edit_bundle.py`) and a service (`ops/services/agreement_edit_bundle.py`) that applies agreement, budget line, services component, and change-request updates in a single transaction
- Refactored `agreements`, `budget_line_items`, `services_component`, and `change_requests` services so the edit-bundle service can compose them without duplicating logic
- New tests in `tests/ops/agreement/test_agreement_edit_bundle.py` covering the bundle flow

### Frontend integration with edit-bundle
- New `editAgreementBundle` RTK Query mutation in `opsAPI.js`
- `EditAgreementAndBudgetLines`, `AgreementEditForm`, and `CreateBLIsAndSCs` updated to call the bundle mutation instead of orchestrating multiple sequential saves
- Updated component tests for the new flow

### New page & routing
- New `EditAgreementAndBudgetLines` page composing `AgreementEditForm` and `CreateBLIsAndSCs`, with accompanying tests
- New route wired up in `index.jsx`
- `ReviewAgreement` now navigates into the new edit flow

### Review mode support
- `CreateBLIsAndSCs`, `ServicesComponentForm`, and `ServicesComponents` accept a review-mode flag so the edit page can render the existing forms with review-appropriate behavior
- `BudgetLinesForm` layout restructured for usability in review mode
- Team member management in review mode includes validation
- `DatePicker` toggles an error class based on validation messages

### Coordinated save flow
- New `saveTrigger` mechanism so the parent page can trigger agreement and budget line saves in sequence and react via `onSaved` callbacks
- New `continueOverRide` flow surfaces a single success alert and redirects back to the review page after both saves succeed
- New `suppressSuccessAlert` parameter on `saveAgreement` / `handleSave` so the child forms don't fire their own success alerts when the parent owns the messaging

### Procurement shop change handling
- Detects procurement shop changes during edit and prompts the user with a confirmation modal before proceeding
- On save failure, services components are re-seeded so the form returns to a consistent state instead of leaving partial edits in place

### Error handling
- Save errors in agreement and budget line flows are now surfaced rather than swallowed
- Data fetch failures on the edit page navigate the user to the error page

### Misc
- Replaced default `React` imports with named `useEffect` / `useRef` imports in touched files
- Updated `createAgreementWithValidations.cy.js` E2E test to cover the new flow
- New unit tests for `EditAgreementAndBudgetLines`, `AgreementEditForm` (project section), and `CreateBLIAndSCs`

## Issue

#2098

## How to test

1. Start the app: `podman compose up --build`
2. Log in and open an agreement that is in the review state.
3. From the Review Agreement page, enter the new edit flow.
4. Edit agreement metadata and budget lines, then save.
   - Confirm a single success alert appears and you are redirected back to the review page.
   - Confirm changes are persisted on reload.
   - Confirm the request goes through the new atomic edit-bundle endpoint (single network call, all-or-nothing).
5. Edit the procurement shop and save.
   - Confirm the confirmation modal appears before the save proceeds.
6. Force a save failure (e.g. invalid data) and confirm services components are re-seeded, no partial state is persisted, and the form is usable again.
7. Trigger a data fetch failure and confirm the user is routed to the error page.

## A11y impact

- [ ] Any temporary suppression includes \`A11Y-SUPPRESSION\` metadata (owner, expires, rationale)


rajohnson90 (changes requested):
AI PR Review identified these things listed below. Ultimately I think 1-3 ar ethe most important. 
Significant

  1. Services call rollback() even when commit=False (services_component.py:988–990, services_component.py:1007–1009)

  except IntegrityError as e:
      self.db_session.rollback()  # ← fires whether commit=True or False
      raise ValidationError(...)

  1. When the orchestrator calls _create_services_components with commit=False, a uniqueness violation on the SC causes
  the SC service to roll back the entire session — including any agreement-level update and previously-flushed SCs from
  the same bundle. The orchestrator's own except Exception block then calls rollback() again on the already-clean
  session. End result is still a full rollback, but it's inconsistent with the commit=False contract: callers expect to
  own the rollback. Fix: guard the internal rollback behind if commit:, and let the orchestrator's except handle it in
  the non-commit case.
  2. No flush() in _apply_direct_edits and _handle_change_requests when commit=False (budget_line_items.py:860,
  budget_line_items.py:877)

  2. BLIService.create() correctly calls flush() when commit=False so the new BLI gets an id. But _apply_direct_edits
  and _handle_change_requests only call commit() — no else: flush(). If any subsequent step in the bundle needs to read
  the BLI's updated state (e.g. another BLI referencing the same SC), it sees stale data. All other service methods in
  this PR add the else: flush() branch — these two should too.

  Minor

  3. AgreementEditBundleRequestSchema bypasses nested schema for missing top-level keys
  (ops/schemas/agreement_edit_bundle.py:360–365)

  services_components = fields.Nested(
      _ServicesComponentMutationsSchema,
      load_default=lambda: {"create": [], "update": [], "delete": []},
  )

  3. When services_components is absent from the payload, Marshmallow returns the load_default dict directly — it never
  routes through _ServicesComponentMutationsSchema. The orchestrator accesses this via .get("create", []), so it works,
  but the schema-layer defaulting and coercion are silently bypassed. A cleaner approach is load_default=None and
  letting the service normalize None to {}.
  4. fake_request = SimpleNamespace(...) leaks HTTP concerns into business logic
  (ops/services/agreement_edit_bundle.py:652–657)

  fake_request = SimpleNamespace(json=data_for_load)
  updated_fields = loaded | {
      "method": "PATCH",
      "schema": patch_schema,
      "request": fake_request,
  }

  4. This is a pre-existing smell in BudgetLineItemService.update(), but the bundle exposes it. Worth a follow-up ticket
  to push request.json reading up to the resource layer so the service takes a plain dict.
  5. Rollback test assertion is underspecified (test_agreement_edit_bundle.py:1321)

  assert response.status_code >= 400

  5. The invalid CAN triggers a DB FK constraint → IntegrityError → re-raised by the orchestrator → Flask returns 500.
  Asserting >= 400 hides whether the bundle is correctly classifying this as a 400 or letting it fall through to 500. If
  it's expected to be a 500 (not a user error), the test should say so; if it should be 400, the orchestrator needs to
  catch FK violations more broadly.
  6. EditAgreementAndBudgetLines doesn't pass servicesComponentsReseedKey (EditAgreementAndBudgetLines.jsx:3148–3153)

  6. The PR description says "on save failure, services components are re-seeded." AgreementEditorContext.jsx adds the
  servicesComponentsReseedKey prop for exactly this. But the new page never bumps it after a failure. This appears to be
  an incomplete implementation of the stated behavior, or the description is inaccurate. Either wire up the reseed on
  failure or update the description.
  7. Route ordering risk (index.jsx:2917–2930)

  7. /agreements/review/:id/edit is registered before /agreements/review/:id/*. React Router v6 matches by specificity,
  not order, so this is fine — but worth a comment since it's easy to accidentally swap them.

  ---
  Test Coverage
  7. /agreements/review/:id/edit is registered before /agreements/review/:id/*. React Router v6 matches by specificity,not order, so this is fine — but worth a comment since it's easy to accidentally
  swap them.

  ---
  Test Coverage

  - Backend: happy paths (empty bundle, agreement-only, SC+BLI create with ref, updates, deletes), rollback semantics, and auth — good coverage of the stated rollback guarantee

  assert response.status_code >= 400

  5. The invalid CAN triggers a DB FK constraint → IntegrityError → re-raised by the orchestrator → Flask returns 500.
  Asserting >= 400 hides whether the bundle is correctly classifying this as a 400 or letting it fall through to 500. If
  it's expected to be a 500 (not a user error), the test should say so; if it should be 400, the orchestrator needs to
  catch FK violations more broadly.
  6. EditAgreementAndBudgetLines doesn't pass servicesComponentsReseedKey (EditAgreementAndBudgetLines.jsx:3148–3153)

  6. The PR description says "on save failure, services components are re-seeded." AgreementEditorContext.jsx adds the
  servicesComponentsReseedKey prop for exactly this. But the new page never bumps it after a failure. This appears to be
  an incomplete implementation of the stated behavior, or the description is inaccurate. Either wire up the reseed on
  failure or update the description.
  7. Route ordering risk (index.jsx:2917–2930)

  7. /agreements/review/:id/edit is registered before /agreements/review/:id/*. React Router v6 matches by specificity,
  not order, so this is fine — but worth a comment since it's easy to accidentally swap them.

Don't worry as much about 4-7 but they are probably nice to fix if you've got the bandwidth.

weimiao67:
@rajohnson90 Thanks for the review! Addressed the items with significant impacts in ecbb99084a10b0849e51fd3c23ea1bfebcbcb024. Please take a look when you have a moment.

1. Inner rollback now respects commit=False
Wrapped the self.db_session.rollback() calls in ServicesComponentService.create() and update() with if commit: so the orchestrator owns rollback when called with commit=False. The outer bundle try/except is now the single source of rollback for the bundle path.
2. else: flush() added to _apply_direct_edits and _handle_change_requests
Both BLI helpers now match the pattern in BLIService.create() and the SC service: if commit: commit() else: flush(). Subsequent steps in the bundle that read BLI state will see the flushed updates.
3. Nested schema no longer bypassed
services_components and budget_line_items switched to load_default=None. Orchestrator already normalizes via payload.get(...) or {}, so absent keys remain a no-op while present payloads always route through _ServicesComponentMutationsSchema / _BudgetLineItemMutationsSchema.
6. servicesComponentsReseedKey wired up on save failure
Added a servicesComponentsReseedKey state to EditAgreementAndBudgetLines.jsx that bumps in the fireBundleSave catch block and is passed through to EditAgreementProvider. Since the bundle save is atomic, server state is unchanged on failure — reseeding from the cached servicesComponents reverts any optimistic edits and leaves the form consistent for retry.
7. Route ordering comment
Added a short comment above /agreements/review/:id/edit noting that React Router v6 matches by specificity, so the order is documentation, not behavior.

rajohnson90 (approved):

Santi-3rd (approved):

Issue #5772, opened 2026-06-03
Enforcing SC and BL obligate by date logic


jonnalley:
### User Story
As an OPS user, I don't want to have budget lines with obligate by dates outside the bounds of the overall agreement PoP dates. 

### Acceptance Criteria

BLs:
- [x]  An authorized (based on existing business rules) user can save or request a status-transition for a non-DRAFT BL (assigned to an SC) whose obligate-by date is on or after the earliest SC period_start AND on or before the latest SC period_end across the agreement.
- [x]  An obligate-by date that exactly equals the earliest SC period_start or the latest SC period_end is valid (bounds are inclusive).
- [x]  A superuser (aka power-user) can save or status-transition a non-DRAFT BL with an obligate-by date **_outside_** the SC PoP window without error.

- [x] A non-superuser cannot save or status-transition a non-DRAFT BL (assigned to an SC) with an obligate-by date before the earliest SC period_start across the agreement. The user receives an actionable error message identifying what constraint was violated.
- [x] A non-superuser cannot save or status-transition a non-DRAFT BL (assigned to an SC) with an obligate-by date after the latest SC period_end across the agreement. The user receives an actionable error message identifying what constraint was violated.

SCs:
- [x] An SC date edit that narrows or shifts that SC's PoP but does not change the agreement-wide lower or upper bound (because another SC already covers the outer edge) is allowed without error.
- [x] An SC date edit that moves the agreement-wide bounds but all existing non-DRAFT BLs (assigned to any SC in the agreement) still fall within the new window is allowed.
- [x] A new SC can be added to an agreement with dates, provided those dates do not narrow the agreement-wide window enough to make any existing non-DRAFT BL out-of-bounds.

- [x] A user cannot edit an SC's period_start to a date that would move the agreement-wide lower bound forward to a point where any existing non-DRAFT BL's obligate-by date falls before the new lower
   bound. The edit is blocked with an error that identifies which BLs are affected.
- [x] A user cannot edit an SC's period_end to a date that would move the agreement-wide upper bound backward to a point where any existing non-DRAFT BL's obligate-by date falls after the new upper
  bound. The edit is blocked with an error that identifies which BLs are affected.
- [x] A user cannot set dates on an SC (on create or first-time date assignment via edit) if those dates would narrow the agreement-wide window such that any existing non-DRAFT BL's obligate-by date
  falls outside the resulting window. The operation is blocked with an error identifying the affected BLs.
[...]

PR #5828, opened 2026-06-18
feat(OPS-5772): enforce BLI obligate-by date within SC Period of Performance

rajohnson90:
## What changed

Enforces that a Budget Line Item's Need By Date (obligate-by date) falls within the Period of Performance (PoP) window defined by the agreement's Service Components, and prevents SC edits that would push non-draft BLIs outside that window.

**`backend/ops_api/ops/services/budget_line_items.py`**
- Refactored the monolithic `_validation_change_status_higher_than_draft` into three focused helpers: `_validate_agreement_for_status_change`, `_validate_amount_and_date_for_status_change`, and `_validate_date_within_sc_window`.
- `_validate_date_within_sc_window`: for non-superusers, rejects any non-DRAFT BLI whose `date_needed` falls outside `[sc_start_date, sc_end_date]` (inclusive). Both bounds are optional — a missing bound places no constraint.

**`backend/ops_api/ops/services/services_component.py`**
- `_validate_bli_pop_window`: called on SC `period_start`/`period_end` updates; rejects the change if any non-draft BLI would fall outside the resulting aggregate window across all SCs.
- `_validate_bli_pop_window_for_create`: called when a new SC is created with dates; same check against the combined window of new + existing SCs.

**`backend/models/agreements.py`**
- `sc_start_date` property: `min(period_start)` across all SCs (None if no SC has a start date).
- `sc_end_date` property: `max(period_end)` across all SCs (None if no SC has an end date).

**`frontend/src/components/BudgetLineItems/BudgetLinesForm/datePickerSuite.js`**
- Adds a third `needByDate` Vest test: "Date must fall within the agreement's period of performance." Guards on `scStartDate`/`scEndDate` being present and uses string comparison on `YYYY-MM-DD` to avoid timezone issues. Each bound is independently optional (one-sided window supported).

**`frontend/src/components/BudgetLineItems/BudgetLinesForm/BudgetLinesForm.jsx`**
- Accepts `scStartDate` and `scEndDate` props, passes them through to the date picker suite on every run. A `useEffect` re-runs the suite whenever the SC window changes so the PoP error appears without the user touching the date field.

**`frontend/src/components/ServicesComponents/ServicesComponentForm/suite.js`**
- New Vest suite. Validates `servicesComponentSelect` is non-zero. In `add`/`edit` mode, computes the aggregate SC window from `allServicesComponents` (pre-merged with live form values by the hook) and rejects changes that would leave any non-draft BLI outside the window. Each bound guards independently — absent bound = no constraint.

**`frontend/src/components/ServicesComponents/ServicesComponentForm/ServicesComponentForm.jsx`**
- Wires up `scFormSuite`, `allServicesComponentsForSuite`, and `nonDraftBudgetLines` props. Runs the suite on every form-data change and surfaces errors on `servicesComponentSelect`, `popStartDate`, and `popEndDate` fields. SC number field errors only shown after the field has been touched.

**`frontend/src/components/ServicesComponents/ServicesComponents.hooks.js`**
- Accepts `scFormSuite` param. Computes `allServicesComponentsForSuite` (memos the SC array with live form dates merged in for the SC being edited). Blocks `handleSubmit` if the suite has errors. Resets the suite on cancel.

**`frontend/src/components/CreateBLIsAndSCs/CreateBLIsAndSCs.jsx` / `.hooks.js`**
- Threads `scStartDate`/`scEndDate` from the agreement's SC window down to `BudgetLinesForm`.

**`backend/openapi.yml`**
- Adds response body schemas for SC PUT/PATCH/DELETE endpoints; adds `403` responses; corrects `ServicesComponent` schema (`nullable` fields, proper `required` placement).

**`frontend/src/types/AgreementTypes.d.ts`**
- Adds `sc_start_date` and `sc_end_date` to the `Agreement` type.

**Tests**
- `test_budget_line_item.py`: SC window boundary tests (before/on/after start and end), one-sided window (start-only, end-only), undefined window (multi-SC all-null), DRAFT→PLANNED transitions under each scenario, superuser bypass.
- `test_budget_line_item_power_user.py`: Superusers can set `date_needed` outside the SC PoP window.
- `test_services_component.py`: SC update/create validations that protect existing non-draft BLIs.
- `datePickerSuite.test.js` / `BudgetLinesForm.test.jsx`: PoP boundary validation in the BLI date picker.
- `ServicesComponentForm/suite.test.js`: Full suite — required-field validation, add/edit mode PoP checks, one-sided windows, undefined windows, two-SC scenarios, draft BLI exemption.
- Cypress E2E: updated selectors/assertions for the revised BLI and agreement workflows.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5772

## How to test

**Backend** (from `backend/ops_api/`):
```bash
# Full BLI validation suite
pipenv run pytest tests/ops/budget_line_items/test_budget_line_item.py -v

# SC window protection
pipenv run pytest tests/ops/services_components/test_services_component.py -v

# Superuser bypass
pipenv run pytest tests/ops/budget_line_items/test_budget_line_item_power_user.py -v
```

**Frontend** (from `frontend/`):
```bash
bun run test --watch=false src/components/BudgetLineItems/BudgetLinesForm/datePickerSuite.test.js
bun run test --watch=false src/components/ServicesComponents/ServicesComponentForm/suite.test.js
bun run test --watch=false src/components/BudgetLineItems/BudgetLinesForm/BudgetLinesForm.test.jsx
```

**Manual (UI)**
1. Create a Contract agreement with a Service Component (SC) with `period_start = 2025-01-01` and `period_end = 2025-12-31`.
2. Add a non-draft BLI. Verify that setting `date_needed` outside `[2025-01-01, 2025-12-31]` shows an inline error and blocks save.
3. Verify that a date inside the window saves successfully.
4. Edit the SC and try to shrink `period_end` to before the BLI's `date_needed`. Verify the form shows an error and blocks the save.
5. As a superuser, verify that the PoP-bounds check is bypassed and any `date_needed` is accepted.

## A11y impact

## Storybook

## Screenshots

_Please add screenshots showing the inline validation error on the BLI date picker and the SC form PoP error._

## Links

N/A


rajohnson90:
Hold on sorry I messed up. The description as is is NOT correct.

backend/ops_api/ops/services/budget_line_items.py
```diff
@@ -849,77 +849,96 @@ def _validation(self, budget_line_item, updated_fields):
...
-            # Check if the date_needed is set and in the future
-            today = date.today()
-            current_date_needed = budget_line_item.date_needed
-            requested_date_needed = updated_fields.get("date_needed")
-            final_date_needed = requested_date_needed if requested_date_needed is not None else current_date_needed
+    @staticmethod
+    def _validate_amount_and_date_for_status_change(budget_line_item, updated_fields):
+        current_amount = budget_line_item.amount
+        final_amount = updated_fields.get("amount") if updated_fields.get("amount") is not None else current_amount
+        if final_amount is None or not isinstance(final_amount, (Decimal, float, int)) or final_amount <= 0:
```
weimiao67: **Accidental regression: `0` is no longer a valid amount.**

This is `final_amount <= 0` here, but on `main` it's `final_amount < 0`. Commit `7c4483b5b (fix(OPS-2279): allow 0 as valid BLI amount)` *intentionally* made `0` valid; commit `ef09be8bd ("fix: fixing mistaken merge issue")` on this branch flipped it back to `<= 0`. This looks like an unintended revert during merge-conflict resolution rather than part of the obligate-by-date feature.

Also note the error message on the next line still reads `"Amount must be 0 or greater."`, which now contradicts the behavior (it rejects `0`).

⚠️ The existing tests do **not** catch this:
- `test_budget_line_item_validation_patch_to_zero_or_negative_amount` creates the BLI with no `services_component_id`, so it 400s on the missing-required-fields check *before* the amount check is ever reached — it only asserts `status == 400`, not which field failed, so it stays green on both `< 0` and `<= 0`.
- The BDD `test_valid_amount_greater_than_zero` (titled "0 is allowed, negative is not") only ever submits a negative amount, never `0`.

**Suggested fix:** restore `final_amount < 0`, and add a test that PATCHes a non-draft BLI to `amount=0` with all required fields present asserting a 2xx, so this can't silently regress again.

weimiao67 (commented):
Two inline findings from review (verified by running the test suites). See comments below.

frontend/src/types/AgreementTypes.d.ts
```diff
@@ -60,6 +60,8 @@ export type Agreement = {
     lifetime_obligated?: number | string | null;
     fy_obligated?: number | string | null;
     award_type?: keyof typeof AWARD_TYPE_LABELS | null;
+    earliest_bli_obligate_by_date?: string | null;
+    latest_bli_obligate_by_date?: string | null;
```
weimiao67: **Dead / mismatched type fields.** `earliest_bli_obligate_by_date` and `latest_bli_obligate_by_date` are added here but referenced nowhere in `frontend/src` or `backend/`. The feature actually uses `sc_start_date` / `sc_end_date` (serialized by the backend in `ops/schemas/agreements.py` and threaded through as `scStartDate` / `scEndDate`).

Recommend dropping these two fields. If the goal was to type the SC window on `Agreement`, add `sc_start_date` / `sc_end_date` instead.

josbell (changes requested):
## Review: Request Changes

Three confirmed bugs need to be fixed before merge, plus some design questions worth clarifying.

---

### 🔴 Bug 1 — `amount=0` silently rejected (regression)

**File:** `backend/ops_api/ops/services/budget_line_items.py` ~line 907

The refactor changed the amount guard from `final_amount < 0` to `final_amount <= 0`:

```python
# before
if final_amount is None or not isinstance(final_amount, (Decimal, float, int)) or final_amount < 0:

# after (current PR)
if final_amount is None or not isinstance(final_amount, (Decimal, float, int)) or final_amount <= 0:
```

The error message still says **"Amount must be 0 or greater."** — but the code now rejects zero. Any non-DRAFT BLI with `amount=0` will get a 400 on any PATCH (even an unrelated field update like changing `date_needed`), because `_validate_amount_and_date_for_status_change` runs for all non-DRAFT BLIs. The user sees an error message that directly contradicts the rejection. Either revert to `< 0` (keeping zero valid, as the message promises) or update the message to "Amount must be greater than 0."

---

### 🔴 Bug 2 — `AttributeError` (500) when `agreement_id` is in the PATCH body but the BLI has no current agreement

**File:** `backend/ops_api/ops/services/budget_line_items.py` ~line 884–889

```python
def _validate_agreement_for_status_change(budget_line_item, updated_fields):
    if not budget_line_item.agreement and (
        "agreement_id" not in updated_fields or updated_fields.get("agreement_id") is None
    ):
        raise ValidationError(...)  # only raises if agreement_id is also missing

    # execution continues here when agreement is None but agreement_id is present:
    agreement_required_fields = budget_line_item.agreement.__class__.get_required_fields_for_status_change()
    #                           ^^^^^^^^^^^^^^^^^^^^^^^^^ AttributeError: 'NoneType' object has no attribute...
```

When a PATCH simultaneously sets `agreement_id` and transitions status (or hits this path on a non-DRAFT BLI), the guard falls through because `agreement_id` is present — but the ORM hasn't flushed the new association yet, so `budget_line_item.agreement` is still `None`. The subsequent dereference of `None.__class__.get_required_fields_for_status_change()` raises an `AttributeError`, which surfaces as an unhandled 500 instead of a clean validation error. The fix is to either eagerly load the agreement from the session using the `agreement_id` in `updated_fields`, or to guard the subsequent lines with a `None` check.

---

### 🔴 Bug 3 — `scFormSuite` never reset on navigation between agreements (stale ghost error)

**File:** `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js` ~line 141

The mount/unmount cleanup effect resets every other suite but omits `scFormSuite`:

```javascript
// existing cleanup — resets suite, budgetFormSuite, datePickerSuite
React.useEffect(() => {
    suite.reset();
    budgetFormSuite.reset();
    datePickerSuite.reset();
    // scFormSuite is NOT reset here
    return () => { ... };
}, []);
```

`scFormSuite` (the SC select required-field suite) is a module-level singleton. If a user leaves an agreement with an SC validation error showing, navigates to a different agreement, and opens the SC form, the "This is required information" error appears immediately before they've touched anything. The fix is to add `scFormSuite.reset()` to the same mount/unmount effect — matching the pattern already used in `ServicesComponents.hooks.js` for `popValidationSuite`.

---

### 🟡 Design question — Union SC window vs. per-SC assignment

**File:** `backend/ops_api/ops/services/budget_line_items.py` ~line 933, `backend/models/agreements.py` ~line 269

`_validate_date_within_sc_window` checks `date_needed` against `agreement.sc_start_date` / `agreement.sc_end_date`, which are `min(period_start)` / `max(period_end)` across **all** SCs on the agreement — the union window. This means a BLI assigned to SC1 (Jan–Jun) with `date_needed` in September can pass validation if SC2 ends in December, because the union max is December.

If the intended rule is "a BLI's obligate-by date must fall within its own assigned SC's window" (not the aggregate), this is a bug. If the rule is "within any SC on the agreement," it's correct as written. Worth a quick clarification against the ticket acceptance criteria before merge.

---

### 🟡 Misleading error message

**File:** `backend/ops_api/ops/services/budget_line_items.py` ~line 939

The validation error reads: *"BLI must have a Need By Date within the **agreement's start and end dates**"* — but the constraint is actually derived from the Service Component `period_start`/`period_end` fields, not any top-level agreement dates. Users and support staff triaging the error will look at the wrong fields. Suggest: *"BLI must have a Need By Date within the agreement's Service Component Period of Performance"*.

---

### 🟡 Frontend PoP check silently skipped when adding a new SC

**File:** `frontend/src/components/ServicesComponents/ServicesComponents.hooks.js` ~line 54

In `add` mode, `allServicesComponentsForSuite` returns only the already-saved SCs — the new SC being added is never merged in. If a user adds the **first** SC on an agreement (or an SC whose dates are narrower than existing non-draft BLIs), `allSCs.length === 0` and the `popValidationSuite` early-returns, so the confirmation modal never appears. The backend `_validate_bli_pop_window_for_create` will then hard-reject the save. The user gets an API error with no prior warning. If this gap is intentional, a comment explaining why add-mode is excluded would prevent a future "fix" from breaking the backend contract.

backend/ops_api/ops/services/budget_line_items.py
rajohnson90: Good catch. Fixed in commit

rajohnson90:
In response to Josbell's design questions:
**Design question — Union SC window vs. per-SC assignment**
This is intended. I also identified this while working and explicitly confirmed it with Kim that this is acceptable and intended behavior.

**Misleading error message**
This is also the error message designed after discussion with UX so this will not be changing. 

** Frontend PoP check silently skipped when adding a new SC**
I think this one is valid and worked on a fix.

weimiao67 (approved):

josbell (dismissed):
## Review

The implementation is solid — all three red bugs from the previous review are fixed, and the `useSaveNotes` dirty-flag pattern and shared mutation instance preventing concurrent PATCHes are well thought out.

### Required before merge

**Missing server-side length validation on `notes` field**

`backend/ops_api/ops/schemas/procurement_tracker_steps.py`

```python
notes = fields.String(required=False, allow_none=True)
```

Every other user-typed string field in the same schema has a `validate.Length` constraint (`requestor_notes`, `reviewer_notes`, `requisition_number`). The frontend enforces 750 chars via `maxLength`, but direct API calls bypass that. Since the column is unbounded `Text`, an oversized payload would persist silently.

Fix: add `validate=validate.Length(max=750)` to match the frontend constraint.

---

### Nice to have (non-blocking)

**E2E coverage for the Save/Edit/Cancel Notes flow**

The new `data-cy` attributes (`save-notes-button`, `edit-notes-button`, `cancel-edit-notes-button`, `notes-save-announcement`) are never exercised in any E2E spec. The notes-only PATCH path is only covered by unit tests with mocked I/O. At minimum, a test that types a note → saves → asserts read mode → reloads → asserts the note survived would close the gap. Fine to track as a follow-up.

**`React.memo(DatePicker)` called inside hook body (Steps 1–5)**

Pre-existing issue, not introduced by this PR. `React.memo(DatePicker)` is called inside the hook body in Steps 1–5, creating a new component type on every render — this defeats memoization and causes DatePicker remounts. Step 6 already uses the correct pattern. Now that the PR adds notes state updates that trigger re-renders in all steps, this is slightly more likely to surface. Worth a follow-up cleanup ticket.

**Two unaddressed yellow items from previous review**

Both were non-blocking and weimiao67 approved without requiring them, but flagging for completeness:
- Union SC window check vs. per-SC — no code comment explaining the intentional design
- Error message still reads "within the agreement's start and end dates" rather than "within the agreement's Service Component Period of Performance"

josbell (changes requested):
## Adversarial Review

### Request Changes

---

#### Critical — SC deletion bypasses PoP window validation

**`frontend/src/components/ServicesComponents/ServicesComponents.hooks.js`**

`handleDelete` never calls `popValidationSuite.run()`. If the deleted SC is the only one whose `period_end` covers a non-draft BLI's `date_needed`, after deletion that BLI becomes permanently uneditable — the backend rejects every subsequent edit because the date is now outside any SC window, but the date can't be changed because the BLI is non-DRAFT.

Fix: Before confirming deletion, run `popValidationSuite` against the remaining SCs (list minus the one being deleted). If non-draft BLIs would fall outside the resulting window, show the same PoP-impact warning modal used in `handleSubmit` instead of the generic delete confirmation.

---

#### High — Superuser bypass exits entire `datePickerSuite`, including format validation

**`frontend/src/components/BudgetLineItems/BudgetLinesForm/datePickerSuite.js`**

The new `isSuperUser` early-return at the top of the suite skips format and future-date checks in addition to the SC window check. A superuser typing `"abc"` into the Need By Date field gets no inline error and must wait for the API round-trip to discover the problem. This is a UX regression — the intent was only to bypass the SC window constraint.

Fix: Move the early-return inside the SC window block only; keep format and future-date tests unconditional:

```js
// format check — always run
test("needByDate", "Date must be MM/DD/YYYY", () => { ... });

// future-date check — non-superusers only
if (!isSuperUser) {
    test("needByDate", "Date must be in the future", () => { ... });
}

// SC window check — non-superusers only
if (!isSuperUser && !data.isDraft && ...) { ... }
```

---

### Nice to Have (non-blocking)

**Misleading error message**

Both the backend (`budget_line_items.py`) and the frontend suite (`datePickerSuite.js`) say "within the agreement's start and end dates" — but the actual constraint is the SC Period of Performance window. Users will look for non-existent agreement-level date fields.

Suggested wording: `"BLI Need By Date must fall within the agreement's Services Component Period of Performance."`

**`popValidationSuite` not reset after `handleDelete`**

`handleCancel` calls `popValidationSuite.reset()` but `handleDelete`'s confirm handler does not. Stale suite results are left in state until the next `handleSubmit`. Add `popValidationSuite.reset()` alongside the existing `setFormData(initialFormData)` call in the delete confirm handler.

**No test coverage for `handleDelete` + PoP impact**

`ServicesComponents.hooks.test.js` covers add, edit, conflict modal, confirm, and cancel — but has no test for deleting an SC that would leave a non-draft BLI outside the remaining window. Given that the delete path has a real bug (above), a test here would have caught it.

rajohnson90:
Handled critical and high issues found by Josbell's adversarial review. Fixed popSuite not being reset on delete and added tests around delete functionality. The concern about the text wording directly conflicts with UX's desire on how to word the messages so it wont' be changing.

josbell (approved):
All previously requested changes addressed. Approving.

Issue #5819, opened 2026-06-16
Submitting changes on BLs in Executing status


jonnalley:
### User Story
As a COR or team member on an agreement, I need to be able to make edits on Budget Liness in Executing status in order to update the BLs as it progresses through the procurement process.

### Acceptance Criteria
- [ ] An already valid and authorized user who can request status or other changes on an agreement (`associated_with_agreement()` ) can go into edit mode and edit Budget Lines (BudgetLineItem)  that are in executing (IN_EXECUTION) status. Similar to how they can make edits to BLs in planned and draft status.
  - [ ] An exception to this is if there's an active ProcurementTracker on the agreement that's in step 5 or 6.  A tracker in Steps 1-4 or no step are ok.
- [ ] Budget-related changes to the BL (proc shop, amount, CAN, obligate by, deletion) would need to invoke a ChangeRequest. A procurement shop change specifically will route through as an agreement-level change request, which would not be a change to existing behavior today. 
- [ ] Non-budget related changes to a budget line (i.e. description, services component association, and comments) would not trigger a ChangeRequest. 

Existing code/business logic within the ProcumenTracker steps already enforces that a user cannot request pre-award approval (step 5) until all pending changes (ChangeRequest) on an agreement and their budget lines are approved or declined. This work should not counteract or introduce a regression with that. No-op. 

-----------------------------------------

#### Potential Backend Unit Tests
- [ ] `is_bli_editable()` returns `True` for an IN_EXECUTION BLI when the agreement has no active ProcurementTracker
- [ ] `is_bli_editable()` returns `True` for an IN_EXECUTION BLI when the active tracker is at steps 1–4
- [ ] `is_bli_editable()` returns `False` for an IN_EXECUTION BLI when the active tracker is at step 5 (PRE_AWARD)
- [ ] `is_bli_editable()` returns `False` for an IN_EXECUTION BLI when the active tracker is at step 6 (AWARD)
- [ ] `is_bli_editable()` returns `False` for an IN_EXECUTION BLI that already has a pending ChangeRequest (`in_review = True`)

#### Potential Backend Integration Tests
- [ ] `PUT /budget-line-items/{id}` — changing `amount` on an IN_EXECUTION BLI (tracker at step 1–4) creates a BudgetLineItemChangeRequest and returns 200; no direct mutation on the BLI
- [ ] `PUT /budget-line-items/{id}` — changing `can_id` on an IN_EXECUTION BLI creates a ChangeRequest
[...]

PR #5832, opened 2026-06-19
Allow editing executing BLs and deletion CRs

jonnalley:
## What changed

Makes Budget Line Items in **Executing** (`IN_EXECUTION`) status editable, the same way Planned BLIs already work (issue #5819). ~This is **PR 1 of 2**; PR 2 will add deletion-of-PLANNED/EXECUTING-BLIs as an approval-routed change request, and the `isDeletable` seam introduced here is what it builds on.~

For reviewers unfamiliar with the area: BLI editability is computed server-side and surfaced to the frontend via a `_meta` object on each BLI (`isEditable`, and now `isDeletable` / `lockedMessage`). The frontend edit/delete/duplicate icons key off those flags. Budget-relevant edits don't apply directly — they create a `BudgetLineItemChangeRequest` that a reviewer approves (HTTP 202); non-budget edits apply immediately (HTTP 200).

- **`backend/ops_api/ops/utils/budget_line_items_helpers.py`** — New single source of truth for editability. Adds `IN_EXECUTION` to the editable statuses; adds `is_agreement_in_pre_award_or_later` (True when the agreement has an ACTIVE procurement tracker at `active_step_number >= 5`, i.e. Pre-Award/Award); adds `compute_bli_editable`, `compute_bli_is_deletable`, and `get_bli_locked_message`. `is_bli_editable` now delegates to the shared core. Removed the dead `_bli_has_editable_status` (old DRAFT/PLANNED-only logic that would have drifted).
- **`backend/ops_api/ops/resources/budget_line_items.py`** — Removed the duplicate `_is_bli_editable_optimized`; the list-meta builder now delegates to the shared core and also sets `isDeletable` / `lockedMessage`.
- **`backend/ops_api/ops/services/budget_line_items.py`** — Appends `comments` to `ALWAYS_DIRECT_EDIT_FIELDS` (PR #5816's two fields preserved); eager-loads `procurement_trackers` in `get_list` and in the agreement-detail meta builder so the step check doesn't lazy-load; computes the three meta flags consistently across all three paths (single GET, list, agreement-detail).
- **`backend/ops_api/ops/schemas/{agreements,budget_line_items}.py`** — `isDeletable` + `lockedMessage` added to both `MetaSchema` classes.
- **Frontend** — `BLIRow.jsx` reads `_meta.isDeletable` (falling back to `isEditable` for legacy payloads) and passes `isItemDeletable` to `ChangeIcons`. `getTooltipLabel` prefers the backend-computed `lockedMessage` and drops the now-inaccurate "Executing status — contact the budget team" copy. `BudgetLineTypes.d.ts` gains the two `_meta` fields.

### Key decisions / assumptions
- **Editability and deletability are deliberately decoupled.** Making Executing BLIs editable must NOT auto-enable their delete button. PR1 keeps today's delete eligibility (DRAFT/PLANNED, or super user); the new `isDeletable` flag is the seam PR2 will repurpose.
- **The Pre-Award/Award block** (procurement tracker step >= 5) is an explicit AC of #5819, not just "edit like Planned." Super users bypass it, consistent with the existing super-user editability bypass.
- **`lockedMessage` is computed on the backend** because the BLI payload carries no procurement-tracker-step data, so the frontend cannot derive "blocked at Pre-Award" on its own.
- **The agreement-level procurement-shop block for IN_EXECUTION BLIs is intentionally untouched** (a regression test asserts it still raises). No new `ChangeRequestType` and no DB migration — this PR is purely editability + meta.
- **The `date_needed` "must be in the future" rule is kept as-is** for non-DRAFT edits (confirmed with product): an Executing BLI with a past need-by date must also be given a future date as part of the edit.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5819

## How to test

_Backend tests (suite self-manages its Postgres via pytest-docker; run with xdist):_
```bash
cd backend/ops_api
pytest tests/ops/budget_line_items/test_edit_executing_budget_line.py \
       tests/ops/features/test_edit_executing_budget_line.py \
       tests/ops/utils/test_budget_line_items_helpers.py \
       tests/ops/agreement/test_agreement_change_requests.py::test_proc_shop_change_still_blocked_for_in_execution_bli -n 2
```
_Frontend tests:_
```bash
cd frontend && bun run test --watch=false src/components/BudgetLineItems/BudgetLinesTable/BLIRow.test.jsx src/helpers/budgetLines.helpers.test.js
```
_Manual (against the acceptance criteria):_
1. As a team member on an agreement whose procurement tracker has NOT reached Pre-Award, open an Executing BLI.
2. Edit the description or comments → applies directly (success). Edit amount / CAN / obligate-by date → "Changes Sent to Approval" (a change request is created).
3. Advance the agreement's procurement tracker to Pre-Award (step 5) → the Executing BLI is no longer editable and the tooltip explains it's because the agreement reached Pre-Award.
4. Confirm the **delete** button stays disabled for the Executing BLI (still limited to DRAFT/PLANNED) while the **edit** button is enabled.
5. Confirm DRAFT and PLANNED BLIs behave exactly as before.

## A11y impact

## Storybook

## Screenshots

N/A — no new visual components; behavior change to existing edit/delete controls.

## Links

~PR 2 of 2 (deletion-as-change-request) to follow, branched from this one.~
PR 2 of this branch was merged into this one #5833


josbell (changes requested):
Found 4 issues that should be addressed:

**Critical:**
1. N+1 query in single-item GET endpoint - causes lazy-loading of procurement_trackers
2. TypeScript type mismatch for isDeletable field

**Recommended:**
3. Non-actionable delete button tooltip
4. Hardcoded procurement step threshold

See inline comments for details.

josbell (commented):
**Issue 1: N+1 Query in Single-Item GET**

`backend/ops_api/ops/services/budget_line_items.py` (line 160)

The `get()` method uses plain `db_session.get()` without eager-loading relationships. When `get_is_editable_meta_data()` checks `is_agreement_in_pre_award_or_later()`, it accesses `agreement.procurement_trackers` which triggers a lazy-load.

The list endpoint already eager-loads this relationship (line 185). Single-item GET should do the same for consistency and performance.

**Suggested fix:**
```python
def get(self, id: int) -> BudgetLineItem:
    budget_line_item = (
        self.db_session.query(BudgetLineItem)
        .filter(BudgetLineItem.id == id)
        .options(
            selectinload(BudgetLineItem.agreement).selectinload(Agreement.procurement_trackers)
        )
        .one_or_none()
    )
    if not budget_line_item:
        raise ResourceNotFoundError("BudgetLineItem", id)
    return budget_line_item
```

josbell (commented):
**Issue 2: TypeScript Type Mismatch**

`frontend/src/types/BudgetLineTypes.d.ts` (line 56)

The TypeScript definition declares `isDeletable: boolean` as required, but the backend schema has `required=False`. The fallback pattern in `BLIRow.jsx` (`isDeletable ?? isEditable`) only works if the field can be undefined.

**Suggested fix:**
```typescript
isDeletable?: boolean;  // Optional for backward compatibility
```

This matches the backend contract and makes the fallback logic type-safe.

josbell (commented):
**Issue 3: Non-Actionable Delete Tooltip**

`frontend/src/components/BudgetLineItems/ChangeIcons/ChangeIcons.jsx` (line 76)

The disabled delete button tooltip just says "Disabled" without explaining why. For IN_EXECUTION BLIs (editable but not deletable), users need to understand the constraint.

**Suggested fix:**
```javascript
label={isItemDeletable ? "Delete" : 
    item?.status === "IN_EXECUTION" 
        ? "Executing budget lines require approval to delete" 
        : "Delete disabled"}
```

This provides context-specific guidance instead of a generic "Disabled" message.

josbell (commented):
**Issue 4: Hardcoded Procurement Step Threshold**

`backend/ops_api/ops/utils/budget_line_items_helpers.py` (line 95)

The magic number `5` is hardcoded for the Pre-Award step threshold. If procurement step numbering changes, this will silently break.

**Suggested fix:**
Define a constant at the top of the file:
```python
PRE_AWARD_STEP_NUMBER = 5  # ProcurementTrackerStepType.PRE_AWARD

def is_agreement_in_pre_award_or_later(agreement) -> bool:
    # ...
    return bool(tracker and tracker.active_step_number and 
                tracker.active_step_number >= PRE_AWARD_STEP_NUMBER)
```

This makes the business logic self-documenting and easier to maintain.

Santi-3rd:
I'm not sure if it's caused by the issues already identified, but I found that you could still press the edit button on an executing budget line while the procurement tracker is on step 5 and a budget line is "In Review".  The backend catches it and causes an error. Which is good but might not be the ideal user experience. This is on agreement 15

https://github.com/user-attachments/assets/c3ffb3ed-401b-40e4-967d-3702556e1e74

jonnalley:
Hey @josbell and @Santi-3rd  this branch has been brought up to date with `main` and your review feedback addressed. There's some cypress test errors I'm still fixing at the moment. Additionally, 5833 (the stacked PR on this one) was merged into this branch on Wednesday so that's new. Since `main` had diverged substantially (it picked up a *different* editable-executing-BLI model via #5838, plus the edit-bundle flow (very much impacting the logic here), `PLANNED_MOD`, and `clin_id`), this update is a merge **plus** the review fixes. Summary of what changed and why below.

  ## Merge with `main` (4 conflicts resolved)
  `main` advanced ~317 commits and adopted its own editability approach, so there were real conflicts in 4 files. Resolutions:

  - **`utils/budget_line_items_helpers.py`** — kept our single-source `compute_bli_editable` / `compute_bli_is_deletable` / `get_bli_locked_message` family; dropped `main`'s now-redundant 
  `_bli_has_editable_status` (verified no remaining references). Preserved `main`'s `PLANNED_MOD` branch in the status→pretty-string converter.
  - **`services/budget_line_items.py`** — kept **both** `main`'s `commit=` param + `clin_id` lazy-CLIN direct-edit path **and** our DRAFT/super immediate-delete vs. PLANNED/IN_EXECUTION delete-as-change-request 
  routing. `ALWAYS_DIRECT_EDIT_FIELDS` is the union of both sides: `{services_component_id, line_description, comments, clin_id}`.
  - **`tests/.../test_budget_line_item.py`** — union of both sides' assertions.
  - **`ChangeRequestsList.jsx`** — re-applied our "render the budget-change card for delete-only CRs" change at `main`'s new card-tagging site (`has_budget_change || has_delete_change`), since `main` refactored 
  the render into a paginated `pageItems.map`.

  ## Addressing the review

**@Santi-3rd — pencil clickable then PATCH errors when pre-award is in review (the main find):** confirmed and fixed. `_meta.isEditable` came from `compute_bli_editable`, which only blocked at tracker step ≥ 5, 
  but the write-path also rejects on `is_pre_award_in_review` (approval requested + pending) — which can fire *before* step 5. So the meta and the write-path had diverged. Fix makes editability a single source 
  of truth: `compute_bli_editable` now also blocks on `is_pre_award_in_review` (matching the write-path guard exactly), and `get_bli_locked_message` returns an explanatory reason, so the pencil disables with a tooltip instead of erroring on click.

  **@josbell #1 — N+1 / lazy-load in the single-item GET meta:** fixed, and it turned out slightly broader than the original note. The real lazy-load was the re-fetch inside `get_is_editable_meta_data` (not 
  `get()`). On top of that, the fix above means editability now walks `tracker.steps`, so **all three** meta surfaces that feed editability now eager-load `procurement_trackers → steps` (single GET, list, and 
  the agreement-embedded builder). No remaining N+1 on those paths.

  **@josbell #2 — TS `isDeletable` / `lockedMessage` typed as required vs. `required=False` in the schema:** fixed. Both are now optional in `BudgetLineTypes.d.ts`, matching the Marshmallow schema
  (`dump_default`, not required) and the defensive `?? isEditable` fallback in `BLIRow`.

  **@josbell — hardcoded step threshold `5`:** addressed, with one correction worth noting. `ProcurementTrackerStepType` is a **string** enum (`PRE_AWARD = "PRE_AWARD"`), so
  `ProcurementTrackerStepType.PRE_AWARD.value` is `"PRE_AWARD"` — comparing `active_step_number >= "PRE_AWARD"` would raise a `TypeError`. The `5` is the PRE_AWARD **step number** (from `ProcurementTracker`'s 
  `step_definitions` mapping), so I introduced a named `PRE_AWARD_STEP_NUMBER = 5` constant, documented against that mapping, and used it in the comparison. Same "no magic number" intent, correct for an int step
  number.

  **@josbell — delete tooltip showing a bare "Disabled":** now obsolete post-#5833 (deletability mirrors editability and disabled cases show explanatory tooltips), so there's no separate change needed.

  ## Additional fixes found while integrating + testing
  - **Edit-bundle atomicity guard:** `main`'s new `PATCH /agreements/<id>/edit-bundle` calls `delete(commit=False)` inside one transaction, but our PLANNED/IN_EXECUTION delete routes to a change request that
  commits + notifies immediately — which would break the bundle's all-or-nothing contract if a non-DRAFT id ever reached it. The bundle only deletes DRAFT today, but the endpoint accepts arbitrary ids, so
  `delete()` now raises rather than silently committing mid-bundle when `commit=False` and the BLI isn't DRAFT/super.
  - **Write-path error message ordering:** because `is_pre_award_in_review` now also feeds `is_bli_editable`, the generic "not in an editable state" message was masking the specific "Pre-Award Approval is in
  review" message at step 5. Reordered `_validation` so the specific, actionable message wins.
  - **Test fixtures vs. `main`'s abstract tracker:** `main` made `ProcurementTracker` abstract (`use DefaultProcurementTracker`). Our #5819 fixtures (written pre-refactor) instantiated it directly, so the
  edit/delete integration + BDD step-def files now use `DefaultProcurementTracker`.

  ## `PLANNED_MOD` (new on `main`, didn't exist when this branch forked)
  Editability for `PLANNED_MOD` is intentionally left matching `main` (not in `EDITABLE_STATUSES`); its delete control becomes gated the same way OBLIGATED is. Issue #5819 predates `PLANNED_MOD` and doesn't
  mention it, so any `PLANNED_MOD`-specific edit/delete semantics are deferred to a follow-up. Note this is orthogonal to #5976 (which only folds `PLANNED_MOD` into the procurement-overview card) — no overlap.

  ## Verification
  - **Backend:** all four #5819 test files (edit + delete, integration + BDD) pass 30/30; the broader BLI suite is green. Added an R1 regression test plus full terminal-state coverage for
  `is_pre_award_in_review` (approved-awaiting-requisition, fully-approved, declined, not-requested), which previously had none. Updated the `ALWAYS_DIRECT_EDIT_FIELDS` assertion for the merged union.
  - **Frontend:** lint clean; the ChangeRequestsList + change-request-helper + BLIRow tests pass.
  - black / flake8 / isort and prettier / eslint all clean.

Great finds you guys, thanks again.   Ready for re-review

jonnalley:
## E2E teardown fixes for the delete-as-change-request behavior ( @josbell @Santi-3rd @rajohnson90  especially)

Heads up on a set of Cypress changes I pushed while getting CI green. By virtue of requiring a change request for trying to delete BLs, that's breaking a lot of cypress tests that create and advance BLs in various statuses and then try to delete them to tear them down. 

Deleting a PLANNED/IN_EXECUTION budget line can no longer end in an immediate hard delete with the changes in this branch. Many of these cypress tests create a BLI, move it to PLANNED (or approve a status change), then delete it as cleanup and asserted `200`. That contract changed, so those assertions had to be re-evaluated.

### Method
I swept every `DELETE /budget-line-items/${id}` site across all `cypress/e2e/*.cy.js` (40+ call sites, 13 files) and traced, per site, the BLI's **actual status at delete time** and the **acting user**. The backend now returns the following HTTP response codes:
- **200** — DRAFT, or a super user (`power-user`) — immediate hard delete.
- **202** — PLANNED/IN_EXECUTION, non-super, and *not* in review — creates a deletion change request.
- **400** — PLANNED/IN_EXECUTION but still **in review** (a pending CR) — blocked.

The key subtleties that drove per-site decisions: a `PATCH` to PLANNED returns 202 but does **not** change the BLI's status — only an *approved* status-change CR does; a DRAFT-but-in-review BLI still hard-deletes with 200 (the DRAFT branch is checked before the in-review gate); and the delete's bearer token is captured at `beforeEach` login, so a later `testLogin(...)` doesn't change the acting user.

### Changes made (three commits)
1. **16 assertions flipped `200 → 202`** across 6 specs (`approveChangeRequestsAtAgreementLevel`, `reviewChangeRequestsAtCardLevel`, `procurementShopChangeRequest`, `declineChangeRequestsAtAgreementLevel`, `budgetChangeRequest`, `saveChangesToEdits`) — sites where the BLI is provably PLANNED/IN_EXECUTION and its CR was already resolved (approved or declined), so the delete now routes to a deletion CR.
2. **2 direct-fix specs** committed earlier (`reviewChangeRequestResponse`, `approveCrossDivisionCRs`) — same root cause.
3. **5 teardown deletes removed entirely** (`budgetChangeRequest` ×4, `declineChangeRequestsAtAgreementLevel` ×1) — these delete a PLANNED BLI that is *still in review* (the CR is intentionally never resolved, or the approval is cancelled), which now returns **400**. Rather than assert 400 (which would bake in an incidental behavior that isn't what those tests are about — they verify the *In Review* state), I dropped the BLI-delete and rely on the subsequent **agreement delete**, which cascade-removes the BLI and its change request. Cleanup is preserved; the now-unused `bliId` destructures were cleaned up.

### Decisions worth being aware of
- **Left unchanged (still 200):** DRAFT deletes, `power-user` (super) deletes in `editBudgetLineByPowerUser`, and DRAFT-but-in-review deletes — all still correctly hard-delete.
- **The "drop the delete" choice** for the 5 in-review sites is the one most worth your eyes. Alternative was to resolve the CR first then delete, but that adds steps to tests whose purpose is the in-review state. Open to changing these to an explicit `400` assertion if you'd prefer the teardown to document the block.

Every site was cross-checked by a second independent pass; eslint/prettier clean on all touched specs.

josbell (changes requested):
Overall this is a strong, well-tested PR — the editability unification into `compute_bli_editable` / `compute_bli_is_deletable` / `get_bli_locked_message` removes two drift-prone duplicates, and the backend deletion-as-change-request flow (snapshot/expunge on approval, edit-bundle `commit=False` guard, `has_delete_change` on non-null JSONB) all checks out. One issue I'd like addressed before merge, plus one design question.

## Requesting one change

**`frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js` — `handleDeletions` doesn't handle the new 202 (routed-to-approval) response.**

In the agreement edit wizard, `BLIRow` now enables the delete control for PLANNED/IN_EXECUTION BLIs (`isDeletable: true`). Clicking delete removes the row from `tempBudgetLines` and fires a "successfully deleted" toast immediately (`handleDeleteBudgetLine`, ~line 674). On Save, `handleDeletions` (~line 235, called ~line 958) does `deleteBudgetLineItem(id).unwrap()` without checking `statusCode`, then `showSuccessMessage` reports the agreement was updated.

But a PLANNED/IN_EXECUTION delete now returns **202** and creates an approval change request — the BLI is **not** deleted. So the user sees the line vanish with a "successfully deleted" message, but it reappears in review state after a cache-invalidated refresh.

The sibling path `AllBudgetLinesTable.hooks.js` (~line 32) was explicitly updated with a `statusCode === 202` branch ("Changes Sent to Approval"); `handleDeletions` was missed. Please mirror that 202 handling here so the wizard's messaging matches reality.

## Design question (non-blocking)

**`backend/ops_api/ops/services/change_requests.py` `update()` — submitter's approval notification is cascade-removed on a delete-CR approval.**

The ordering creates + commits the "Budget Change Request APPROVED" `ChangeRequestNotification` (line 108) and then deletes the BLI (lines 125-126). That commit cascades `budget_line_item → change_request → change_request_notification` (both FKs are `ON DELETE CASCADE`), destroying the notification that was just written. The requestor gets no in-app notice that their deletion was approved.

I see the approval test docstring calls this out as intentional ("the ChangeRequest row and its notification row are intentionally cascade-removed with the BLI"), with `AgreementHistory` as the durable record — so this is a deliberate call, not a bug. Just flagging to confirm that's the intended UX for the submitter, since the approve path silently swallows the notification while reject keeps it. If it's intentional, no change needed.

jonnalley:
@josbell — thanks for the thorough review. Confirmed the `handleDeletions` issue and fixed it. Two notes on the specifics, since they changed the shape of the fix.

## `handleDeletions` — confirmed, fixed

You're right: in the edit wizard, a PLANNED/IN_EXECUTION delete now returns 202 and creates an approval change request (the BLI isn't deleted), but `handleDeleteBudgetLine` fired a "successfully deleted" toast and the save reported "Agreement Updated" — so the row vanishes and then reappears in review state after a cache-invalidated refresh. Real bug.

Two premises in the suggested fix didn't hold up when I went to mirror the pattern, so flagging them:

1. **`AllBudgetLinesTable.hooks.js` doesn't actually have a `statusCode === 202` branch to mirror.** Its delete handler is a single `.then((fulfilled) => …)` with one "Budget Line Deleted" success alert and an error `.catch` — no 200-vs-202 distinction and no "Changes Sent to Approval".
2. **`deleteBudgetLineItem` doesn't surface an HTTP status code.** The mutation has no `transformResponse`, so `.unwrap()` resolves to the raw body — `statusCode` isn't on it. A `statusCode === 202` check would read a field that doesn't exist. (Nothing in `frontend/src` reads `meta.response.status` today.)

So rather than a status-code check, I matched the pattern this codebase already uses for the edit-via-change-request messaging: **infer approval-routing from client signals.** The delete of a PLANNED/IN_EXECUTION line by a non-super user routes to approval — same complement as the backend contract (DRAFT/super → 200 hard delete; PLANNED/IN_EXECUTION non-super → 202 CR). Concretely:

- A small `isDeletionRoutedToApproval(budgetLine, isSuperUser)` helper.
- The optimistic delete toast now says the deletion was queued and will be sent for approval on save (instead of "successfully deleted") for approval-routed lines; DRAFT/super still say "deleted".
- `showSuccessMessage` shows "Changes Sent to Approval" when any deleted line is approval-routed. This also closes a related gap: deletions are removed from `tempBudgetLines`, so the existing `isThereAnyBLIsFinancialSnapshotChanged` flag never covered a delete-only save — those were silently reporting "Agreement Updated" even for edits.

Added unit coverage for the predicate and the toast/save messaging (PLANNED → approval; DRAFT and super-user → immediate delete). Frontend-only change.

## Design question (notification cascade)

Intentional — `AgreementHistory` is the durable record for the approved deletion, so the CR/notification cascading away with the BLI is by design. No change.

josbell (approved):
All three critical issues are resolved:
- N+1 query: fixed via eager-loading in `get_is_editable_meta_data()`
- TypeScript `isDeletable` optional: fixed
- `handleDeletions` 202 handling: client-side `isDeletionRoutedToApproval()` inference is the correct pattern for the batch-save model

One remaining non-critical item: `ChangeIcons.jsx:76` still shows the generic `"Disabled"` tooltip for IN_EXECUTION BLIs that are editable-but-not-deletable. The `lockedMessage` prop is already passed in from `BLIRow` but only used in the `!isItemEditable` branch. Suggest addressing in PR 2 when `ChangeIcons.jsx` is likely touched anyway.

jonnalley:
@josbell thanks, that's a good catch.  The delete tooltip now shows the descriptive `tooltipLabel`, falling back to "This budget line can't be deleted" (your chosen wording) instead of "Disabled". One production line;
  +5 tests (descriptive message, fallback, "Delete", and both loading-state branches).

jonnalley:
Notes on merge conflict resolution in `8b19160ea`:
- budget_line_items.py: ALWAYS_DIRECT_EDIT_FIELDS union — {services_component_id, grant_number_id, line_description, comments, clin_id}; verified no routing ambiguity vs
  budget_field_names.
- `CreateBLIsAndSCs.hooks.test.js`: lossless union of both test suites (correct describe nesting; independently verified 21 tests, 0 lost/duplicated).
- Caught and fixed a merge-exposed test interaction: main's restructured save flow (SC/grant creation before deletions) made 2 delete-flow tests throw; fixed with localized mock
  setup, assertions unweakened. Root-caused a secondary `clearAllMocks`-doesn't-clear-implementations pollution issue too.

jonnalley:
Several non-trivial modifications to this branch were required following the merge of @josbell's  #6012.  The gates for editability of executing BLs were made to match what the intent of main / #6012 shows.  More detail:


1. Super users are now subject to the locks (main's OPS-2280 deliberately removed the super-bypass) — I chose to adopt main.
2. Fold is_post_pre_award_locked into the editability meta (compute_bli_editable + get_bli_locked_message), preserving this PR's meta↔write-path lockstep
3. Main's completion-state lock model supersedes this branch's step-5 position gate. I chose to go with main; so I removed `is_agreement_in_pre_award_or_later` + `PRE_AWARD_STEP_NUMBER` entirely.

  What I did:
- Resolved` _validation` to main's lock structure (in-review + post-pre-award locks, budget-team bypass, clin-only-edit exception), with the generic editability check last.
- Reworked `compute_bli_editable/get_bli_locked_message`: dropped the step gate, added the post-pre-award lock, removed the super-bypass on both locks (OBE bypass retained).
- Ran a dead-code-after-removal sweep (skill): deleted the dead function+constant and cleaned every dependent — the orphaned test import, the ~6 obsolete tests, and 3 stale comments in the service file; verified zero remaining references.
- Reworked the test files based on these updates.

jonnalley:
@weimiao67 @Santi-3rd would appreciate any review on this you're able to provide at your convenience so we can get this PR taken care of.

Santi-3rd (approved):

weimiao67 (approved):
Reviewed the full diff and ran the touched test suites locally — frontend unit tests (132 passed) and the backend edit/delete/helper suites (63 passed), all green.

The editability unification into `compute_bli_editable` / `compute_bli_is_deletable` / `get_bli_locked_message` removes the drift-prone duplicates and keeps the meta in lockstep with the write path. The delete-as-change-request routing (200/202/ValidationError) is handled consistently across all call sites, and the expunge/snapshot ordering in `change_requests.py update()` correctly handles the CASCADE-on-delete. Dead-code sweep is clean — no dangling references to the removed helpers.

One minor, non-blocking cosmetic note: for a budget-team user who both edits financials directly AND deletes a PLANNED line, the "Pending Changes" list in `showSuccessMessage` would list the directly-applied edits alongside the genuinely-pending deletion. Narrow combo case, safe to defer.

LGTM 🚀

PR #5884, opened 2026-07-01
fix: removed edit button for read only user and alert now closes

Santi-3rd:
## What changed

Fixed the success alert bug and removed edit button from the Can Funding tab when logged in as the Read-Only User.

## Issue

#2728 

## How to test

- log in s Read-Only user
- go to any Can Funding page to verify that the edit button does not render
- log in as budget team member
- go to user admin and change a user's role
- Verify that the alert renders, is closable and disappears on its own after 6 seconds

## A11y impact

## Storybook

### DIFF
--- a/frontend/src/components/Users/UserInfo/UserInfo.jsx
+++ b/frontend/src/components/Users/UserInfo/UserInfo.jsx
@@ -1,10 +1,8 @@
 import React, { useEffect } from "react";
-import { useDispatch } from "react-redux";
 import { useGetDivisionsQuery, useUpdateUserMutation } from "../../../api/opsAPI.js";
 import { useGetRolesQuery } from "../../../api/opsAuthAPI.js";
 import constants from "../../../constants.js";
 import useAlert from "../../../hooks/use-alert.hooks.js";
-import { setIsActive } from "../../UI/Alert/alertSlice.js";
 import ComboBox from "../../UI/Form/ComboBox/index.js";
 import { USER_STATUS } from "./UserInfo.constants.js";
 import { useNavigate } from "react-router-dom";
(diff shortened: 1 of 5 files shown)


josbell (approved):
Approving — the two production changes are correct. The alert fix properly resolves the re-appearing-alert bug (the removed `dispatch(setIsActive(true))` was redundant since `setAlert` already sets `isActive`; the real fix is `updateUserResult.reset()` clearing the lingering `isSuccess`/`isError` that caused the effect to re-fire the alert on every render). The read-only gating in `CanFunding.jsx` is logically and structurally correct.

Per our discussion, the remaining gap is test coverage — both behavior changes in this PR are currently unverified. These specs are needed before merge:

**1. Read-only gating (`CanFunding.test.jsx`) — the PR's headline feature is untested**
The existing mock hardcodes `useIsUserReadOnly: () => false`, so only the button-shown path runs. Add:
- A test with `useIsUserReadOnly: () => true` asserting the Edit button is **absent** (e.g. `expect(screen.queryByRole("button", { name: /edit/i })).not.toBeInTheDocument()`).
- A test with `useIsUserReadOnly: () => false` asserting the Edit button **renders**.

This also protects the 90% coverage gate now that a new `!isReadOnly` branch exists.

**2. Alert lifecycle (`UserInfo`) — the exact bug being fixed has no regression guard**
The current `UserInfo` test stubs `setAlert` as a no-op and the mutation mock always returns `isSuccess/isError: false`, so the fixed branches never execute. Add a test that:
- Simulates a successful mutation (`isSuccess: true`) and asserts `setAlert` is called with the success heading **and** `updateUserResult.reset()` is invoked (guards against the alert re-firing / not closing).
- Optionally covers the error branch similarly.

Non-blocking notes for context (no action required for this PR):
- `useAlert` returns a fresh `setAlert` closure each render, so the effect runs on every render — currently harmless (branches are guarded) but worth memoizing later.
- On the error path, `UserInfo` both sets an error alert and navigates to `/error`, so the error `setAlert` is largely moot — pre-existing, not introduced here.

rajohnson90 (approved):

Issue #5435, opened 2026-04-06
Edit Notes on Procurement Tracker


kimschulke:
### Goals
* Notes should always be editable on procurement tracker steps (instead of saved during step completion)

### Tasks
- [x] Adjust designs
- [x] Ready for dev in Figma

Dev tasks:
- [ ] Update frontend

### Additional Context
* This was discovered during a recent demo/stakeholder feedback

### Resources
*

kimschulke:
This is ready for dev - [updated in the figma workflow for procurement tracker](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=30901-20137&t=1xKo86HyOt4vNYNQ-1)

A few notes:
- users should be able to update/save notes independently of step completion
- when a step is active/current, a user can provide notes, click save and see their notes have been saved (changes to read/view display)
- If a user enters notes as a part of their step completion (task completed by, date completed, etc) but they do not click "save notes" before clicking "complete step" - then we will save their notes for them, same as the interaction we have for Target Completion Date
- when a step is active/current, a user can edit previously saved notes
- For a future step that is inactive/didnt happen yet, a user cannot provide notes because the step is disabled
- For a past step that is completed/already happened, a user can edit notes because the edit notes button displays next to the note
- The edit notes text always displays at the end of the user input (dynamic placement depending on text length)

kimschulke:
A few screenshots below -- please make sure you review the entire workflow as there are additional interactions mocked up.

<img width="732" height="286" alt="Image" src="https://github.com/user-attachments/assets/d1db4065-24d1-4580-bebe-d716ce7a2eb2" />

<img width="818" height="438" alt="Image" src="https://github.com/user-attachments/assets/69447b22-3e1f-4829-a896-5c4676b15342" />

rajohnson90:
Merged into main! Should be in dev/staging shortly and ready for a UX review.

jonnalley:
@andressacurvelo @npage123 @kimschulke see Reed's note above!  😄

kimschulke:
testing in staging now! its working but with a few inconsistencies described below. additionally, the only scenario that doesnt seem to be working is when I try to add a note into a future step. In that case, I can top into the notes component, click save notes, and I receive a toast notification that my notes were saved. However, if I navigate away and come back, my note is not saved

kimschulke:
[...]

PR #5885, opened 2026-07-01
feat(procurement): enable saving and editing notes on procurement tracker steps

Santi-3rd:
## What changed

Enables users to save notes on procurement tracker steps independently of completing the step, and to edit notes on steps that are already completed.

**Frontend — `frontend/src/components/Agreements/ProcurementTracker/ProcurementTrackerStep{One..Six}/`**
- The notes `TextArea` is now enabled independently of the step's completion / pre-solicitation state, with a dedicated **Save Notes** button (`data-cy="save-notes-button"`) that PATCHes only the `notes` field and surfaces a success/error alert.
- Notes state hydrates from `stepXData?.notes` and stays in sync via a `useEffect`, so previously saved notes are shown.
- Completed steps render an inline **Edit Notes** / **Cancel** flow, allowing notes to be edited and re-saved without reopening the whole step.
- Added `handleSaveNotes` to each step's hook and wired it through the component.

**Backend — `backend/ops_api/ops/validation/rules/procurement_tracker_step.py`**
- `NoUpdatingCompletedProcurementStepRule` now allows a notes-only update on a completed step (via `EDITABLE_AFTER_COMPLETION = {"notes"}`) while still blocking any other field change.
- `AcquisitionPlanningRequiredFieldsRule` only enforces `task_completed_by` / `date_completed` when the step is actually being moved to complete, and treats fields already persisted on the step as satisfied — so notes-only edits and partial updates aren't wrongly rejected.

**Tests**
- Updated BDD feature + step definitions to cover "notes-only update on a completed step succeeds" and "cannot complete acquisition planning step without required fields".
- Updated the API event test to send the now-required completion fields.
- Extensive unit tests added/updated across all six step components and their hooks.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5435

## How to test

Backend:
```bash
cd backend/ops_api
pipenv run pytest tests/ops/features/test_validate_procurement_tracker_steps.py
pipenv run pytest tests/ops/procurement_tracker/test_procurement_tracker_steps_api.py
```

Frontend:
```bash
cd frontend
bun run test --watch=false src/components/Agreements/ProcurementTracker
```

Manual:
1. Open an agreement's Procurement Tracker.
2. On an incomplete step, type in the Notes field and click **Save Notes** — confirm a success alert and that the note persists after reload.
3. Complete a step, then click **Edit Notes**, change the text, and **Save Notes** — confirm the note updates and the step stays completed.
4. Click **Edit Notes** then **Cancel** — confirm edits are discarded.
5. Confirm you still cannot change non-notes fields on a completed step.

## A11y impact

## Storybook

## Screenshots

image

## Links

_N/A_


backend/ops_api/tests/ops/features/test_validate_procurement_tracker_steps.py
```diff
@@ -101,7 +101,7 @@ def test_validate_updating_procurement_tracker_step_with_valid_status(): ...
 
 @scenario(
     "validate_procurement_tracker_steps.feature",
-    "When no presolicitation package is sent to proc shop, the request is valid with unfilled request",
```
rajohnson90, line 104: This change is confusing. Doesn't seem like this makes sense anymore? Based on test name.

backend/ops_api/tests/ops/features/validate_procurement_tracker_steps.feature
```diff
@@ -68,7 +68,7 @@ Feature: Validate Procurement Tracker Steps
 
     Then I should get a validation error
 
-Scenario: When no presolicitation package is sent to proc shop, the request is valid with unfilled request
+Scenario: Cannot complete acquisition planning step without required fields
```
rajohnson90, line 71: Seems like we should double check this.

frontend/src/components/Agreements/ProcurementTracker/ProcurementTrackerStepOne/ProcurementTrackerStepOne.jsx
```diff
@@ -223,7 +246,69 @@ const ProcurementTrackerStepOne = ({
...
+                                            disabled={isDisabled}
+                                        >
+                                            Cancel
+                                        </button>
+                                        <button
+                                            type="button"
+                                            className="usa-button usa-button--unstyled"
+                                            data-cy="save-notes-button"
+                                            onClick={async () => {
+                                                await handleSaveNotes(stepOneData?.id);
```
weimiao67: **Edit mode exits even when the save fails (minor UX/correctness).**

`handleSaveNotes` swallows its error (catch → alert, no rethrow), so this `await` always resolves and `setIsEditingNotes(false)` runs unconditionally. On a failed save the editor collapses back to the label showing the *old* server value, while the user's edited text silently remains in `step1Notes` state.

Consider only exiting edit mode on success — e.g. have `handleSaveNotes` return a boolean and gate the toggle:
```js
onClick={async () => {
    const ok = await handleSaveNotes(stepOneData?.id);
    if (ok) setIsEditingNotes(false);
}}
```

This same pattern is duplicated in the completed-state Save button of all six step components.

frontend/src/components/Agreements/ProcurementTracker/ProcurementTrackerStepOne/ProcurementTrackerStepOne.hooks.js
```diff
@@ -41,6 +44,27 @@ export default function useProcurementTrackerStepOne(stepOneData, handleSetCompl
 
     let validatorRes = suite.get();
 
+    const handleSaveNotes = async (stepId) => {
```
weimiao67: **Significant duplication (main cleanup opportunity).**

This `handleSaveNotes` is copy-pasted verbatim across all six step hooks (`ProcurementTrackerStep{One..Six}.hooks.js`), and the ~65-line Edit/Cancel/Save JSX block is likewise duplicated across all six components. That's a lot of surface area to keep in sync.

Consider extracting:
- a shared `useSaveNotes(patchMutation, notes, setAlert)` hook, and
- a shared `<StepNotesEditor>` component for the completed-state edit flow.

Besides cutting the duplication, this would let one test suite cover all six and would localize fixes (e.g. the "don't exit edit mode on failed save" issue) to a single place.

rajohnson90 (changes requested):
AI PR review identified at least one problem that needs to be fixed: 


  frontend/.../ProcurementTrackerStepOne.hooks.js:24 — useEffect overwrites in-progress draft on RTK Query 
  refetch (CONFIRMED, same in all 6 hooks)
  handleSaveNotes invalidates the "ProcurementTrackers" cache tag, triggering an immediate refetch. When
  the new stepOneData.notes prop arrives, the useEffect fires setStep1Notes(stepOneData?.notes ?? "") —
  overwriting keystrokes typed since the last save. Race: user clicks Save → types more text → refetch
  resolves → effect resets the field to the older saved value. Fix: skip the effect when the field is dirty
  (track with a isDirty ref, reset after a successful save).

frontend/src/components/Agreements/ProcurementTracker/useSaveNotes.js
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }, [serverNotes]);
+
+    /**
+     * Updates the notes value and marks the field dirty so it won't be
+     * overwritten by a subsequent server sync.
+     * @param {string} value - The new notes value.
+     */
+    const setNotes = React.useCallback((value) => {
+        isDirtyRef.current = true;
```
weimiao67, line 41: **`setNotes` marks the field dirty unconditionally, so Cancel (and completion) leave it permanently dirty.**

`setNotes` always sets `isDirtyRef.current = true`, even when the caller is restoring the server value rather than typing a fresh edit. `StepNotesEditor`’s Cancel handler calls `setNotes(savedNotes ?? "")`, which brings the field value back in line with the server but leaves the dirty flag `true`.

Per this hook’s own docstring, the sync-from-server effect is supposed to resume once the field is clean again. After a Cancel it never will (the flag is only cleared on a *successful save*), so a later external server update — another user editing the note, or a refetch delivering a newer value — won’t flow into the field. The completion handlers (`handleStep1Complete`, etc.) have the same issue: they send `notes` but never clear the flag.

Suggestion: expose a `resetNotes(value)` (or `clearDirty`) from `useSaveNotes` that sets the value *and* clears `isDirtyRef`, and use it in the Cancel path (and after completion). Edge-case severity, but it contradicts the documented intent and none of the current tests cover it.

Santi-3rd:
Duplicated inline "Save Notes" block across all 6 step files — Medium

  The active-step editing markup (<div className="display-table"> + TextArea with textAreaStyle={{ height: "8.5rem", minWidth: "30rem" }} + maxLength={750} + the faCheck "Save Notes" button) is copy-pasted verbatim into all six
  ProcurementTrackerStep*.jsx files, and is nearly identical to the editing branch already inside StepNotesEditor.jsx. This is the exact repetition the branch set out to remove (refactor: removed repeated code). Extract a shared
  SaveNotesButton (or let StepNotesEditor cover the active-step form too) so the style constants and button live in one place.

backend/ops_api/tests/ops/features/validate_procurement_tracker_steps.feature
rajohnson90: Resolved as not an issue in meeting.

backend/ops_api/tests/ops/features/test_validate_procurement_tracker_steps.py
rajohnson90: Resolved as not an issue in meeting.

weimiao67 (approved):

josbell (approved):

rajohnson90 (approved):

Issue #5879, opened 2026-06-30
request to move/convert some agreements


jonnalley:
### Goals
* Convert some contracts to AAs with a new proc shop.  We were also sent some BLs to add/update.

### Tasks
- [ ] Check out the BLs for the one agreement that we were asked to replace/remove the BLs on
- [ ] Upsert/convert the AAs
- [ ] Upset the BLs

### Additional Context
* 

### Resources
* see ops support email

jonnalley:
running into data issues that need clarity from OPRE as well as script issues with the agreement_history updates. will look into it further

PR #5891, opened 2026-07-03
Preserve related records during AA conversion & BLI delete history

jonnalley:
## What changed

This PR supports client-requested data updates (converting existing Contract/IAA agreements into AA agreements, and removing budget lines) tracked in issue #5879. While running the `aas` and `remove_budget_lines` data-import scripts against real data, several bugs and data-loss issues surfaced. This PR fixes them and these changes were already used for the live data modifications.

Background: the `aas` importer converts an existing agreement into a new `AaAgreement` by repointing the old agreement's budget lines onto the new AA and then **deleting** the old agreement. Any table that still references the old agreement's id was either being cascade-deleted or left orphaned. This PR repoints all of those references to the new AA *before* the delete. It also fills a gap where the budget-line removal script never recorded agreement history.

**`backend/data_tools/src/load_aas/utils.py`** (agreement conversion)
- On re-import of an existing AA, preserve fields the CSV does not carry (`awarding_entity_id`, `product_service_code_id`, `project_officer_id`, `alternate_project_officer_id`) so the upsert no longer nulls them out.
- Before deleting the converted agreement, repoint its related records to the new AA so they survive and stay associated:
  - **CLINs** (`clin.agreement_id`) — otherwise cascade-deleted (`ondelete=CASCADE`), which also left repointed budget lines with dangling `clin_id`s.
  - **Procurement actions & trackers** (`agreement_id`) — the cascade delete previously violated the `procurement_tracker → procurement_action` FK and crashed the import.
  - **Agreement history** (`agreement_id` + `agreement_id_record`) — kept associated with the new AA instead of nulled/orphaned.
  - **Team members** (`agreement_team_members.agreement_id`) — otherwise removed with the old agreement.

**`backend/models/agreement_history.py`** (shared history logic — used by the live API too)
- Null-guard the procurement-shop fee-total computation (`(item.amount or 0)`) in both the agreement-update and change-request paths. A budget line with a `NULL` amount previously raised `TypeError: unsupported operand type(s) for *: 'NoneType' and 'decimal.Decimal'`.
- Remove the hardcoded word "Draft" from the BLI-deletion history message. The message assumed every deleted budget line was in DRAFT status, but any status can be deleted (and the code has no access to status here). Now reads `deleted the BL {id}`.

**`backend/data_tools/src/load_remove_budget_lines/utils.py`** (budget-line removal)
- After creating the `DELETE_BLI` OpsEvent, fire `agreement_history_trigger_func` so deletions produce a user-facing `AgreementHistory` record — matching newer loaders (`load_aas`, `load_grants`). Flushes first so the event's `id`/`created_on` are populated before the trigger reads them. The `DRY_RUN` env var now drives both the trigger's commit and the rollback/commit (single source of truth).

**Tests** — regression coverage added/updated in the three corresponding test files (field preservation on re-import, `None`-amount fee guard, BLI-delete history record, and the updated "deleted the BL" message assertion). I was pretty selective about adding new tests because some of these scenarios are highly unusual and infrequent. 

## Issue

https://github.com/HHS/OPRE-OPS/issues/5879

## How to test

Automated tests (from the repo root, using the project venv or pipenv):

```bash
# data_tools tests
cd backend/data_tools
pipenv run pytest tests/load_aas/test_load_aas.py
pipenv run pytest tests/load_remove_budget_lines/test_load_remove_budget_lines.py

# shared model / API history tests
cd ../ops_api
pipenv run pytest tests/ops/agreement/test_agreement_history.py
```

Manual / end-to-end (against a seeded DB):

```bash
cd backend
# AA conversion — should complete without a ForeignKeyViolation or TypeError.
# Spot-check that a converted agreement's CLINs, procurement actions/trackers,
# agreement_history rows, and team-member rows now carry the new AA's id.
python data_tools/src/load_data.py --env <env> --type aas --input-csv <aas.tsv>

# Budget-line removal — confirm an AgreementHistory row (BUDGET_LINE_ITEM_DELETED)
# is created per deleted BLI, with a message like "... deleted the BL <id>." (no "Draft").
python data_tools/src/load_data.py --env <env> --type remove_budget_lines --input-csv data_tools/test_csv/budget_line_items_to_delete.tsv
```

## A11y impact

## Storybook

## Screenshots

N/A — backend / data-tools changes only.

## Links

N/A

### DIFF
--- a/backend/data_tools/src/load_aas/utils.py
+++ b/backend/data_tools/src/load_aas/utils.py
@@ -230,9 +230,6 @@ def create_models(data: AAData, sys_user: User, session: Session) -> None:
 
         session.merge(aa)
         session.flush()
-        # Set Dry Run true so that we don't commit at the end of the function
-        # This allows us to rollback the session if dry_run is enabled or not commit changes
-        # if something errors after this point
         agreement_history_trigger_func(ops_event, session, sys_user, dry_run=False)
         session.commit()
 
--- a/backend/data_tools/src/load_aas/utils.py
+++ b/backend/data_tools/src/load_aas/utils.py
@@ -259,11 +259,69 @@ def create_models(data: AAData, sys_user: User, session: Session) -> None:
 
                     session.add(new_budget_line_item)
 
+            # Repoint COR/FPO, alternate, nick_name, SRT, procurement shop, agreement reason, and PSC ID from the existin agreement
+            aa.project_officer_id = existing_agreement.project_officer_id
+            aa.alternate_project_officer_id = existing_agreement.alternate_project_officer_id
+            aa.awarding_entity_id = existing_agreement.awarding_entity_id
+            aa.product_service_code_id = existing_agreement.product_service_code_id
+            aa.nick_name = existing_agreement.nick_name
+            aa.service_requirement_type = existing_agreement.service_requirement_type
+            aa.agreement_reason = existing_agreement.agreement_reason
+
             # Repoint CLINs to the new AA so they survive the agreement delete
             # (clin.agreement_id has ondelete=CASCADE, so otherwise they'd be deleted).
             clins = session.execute(select(CLIN).where(CLIN.agreement_id == existing_agreement.id)).scalars().all()
             for clin in clins:
                 clin.agreement_id = aa.id
+            session.commit()
+
+            # Repoint procurement actions/trackers to the new AA so they survive the delete
+            # (both have agreement_id under cascade="all, delete"; deleting the old agreement
+            # would cascade-delete them and violate procurement_tracker -> procurement_action FK).
+            proc_actions = (
+                session.execute(
+                    select(ProcurementAction).where(ProcurementAction.agreement_id == existing_agreement.id)
+                )
+                .scalars()
+                .all()
+            )
+            for proc_action in proc_actions:
+                proc_action.agreement_id = aa.id
+            proc_trackers = (
+                session.execute(
+                    select(ProcurementTracker).where(ProcurementTracker.agreement_id == existing_agreement.id)
+                )
+                .scalars()
+                .all()
+            )
+            for proc_tracker in proc_trackers:
+                proc_tracker.agreement_id = aa.id
+            session.commit()
+
+            # Repoint agreement history to the new AA so it stays associated after the delete
+            # (agreement_id has ondelete=SET NULL; agreement_id_record is a durable audit int).
+            history_records = (
+                session.execute(
+                    select(AgreementHistory).where(AgreementHistory.agreement_id_record == existing_agreement.id)
+                )
+                .scalars()
+                .all()
+            )
+            for history_record in history_records:
+                history_record.agreement_id = aa.id
+                history_record.agreement_id_record = aa.id
+
+            # Repoint team-member associations to the new AA so the links survive the delete
+            # (agreement_team_members rows are otherwise removed with the old agreement).
+            team_member_rows = (
+                session.execute(
+                    select(AgreementTeamMembers).where(AgreementTeamMembers.agreement_id == existing_agreement.id)
+                )
+                .scalars()
+                .all()
+            )
+            for team_member_row in team_member_rows:
+                team_member_row.agreement_id = aa.id
             session.commit()
 
             logger.info(f"Removing existing agreement {existing_agreement.name} with ID {existing_agreement.id}")
--- a/backend/data_tools/src/load_aas/utils.py
+++ b/backend/data_tools/src/load_aas/utils.py
@@ -191,6 +191,11 @@ def create_models(data: AAData, sys_user: User, session: Session) -> None:
             aa.id = existing_aa.id
             aa.created_on = existing_aa.created_on
             aa.created_by = existing_aa.created_by
+            # Preserve fields the CSV does not carry so the upsert does not null them out
+            aa.awarding_entity_id = existing_aa.awarding_entity_id
+            aa.product_service_code_id = existing_aa.product_service_code_id
+            aa.project_officer_id = existing_aa.project_officer_id
+            aa.alternate_project_officer_id = existing_aa.alternate_project_officer_id
 
             updates = generate_agreement_events_update(
                 existing_aa.to_dict(),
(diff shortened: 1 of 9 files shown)


backend/data_tools/src/load_remove_budget_lines/utils.py
```diff
@@ -101,7 +102,17 @@ def create_models(data: BudgetLineItemData, sys_user: User, session: Session) ->
...
-    session.commit()
+    session.flush()  # Populate ops_event.id and created_on before the history trigger reads them
+
+    # Fire the agreement-history trigger so the deletion is recorded in AgreementHistory.
+    # DRY_RUN drives both the trigger's commit and our rollback/commit (single source of truth).
+    dry_run = bool(os.getenv("DRY_RUN"))
+    agreement_history_trigger_func(ops_event, session, sys_user, dry_run=dry_run)
+    if dry_run:
+        logger.info("Dry run enabled. Rolling back transaction.")
+        session.rollback()
```
josbell: **Highest-priority note (safety):** this `session.rollback()` runs too late to protect the delete. The `session.delete(budget_line_item)` + `session.commit()` up at lines 85-86 run unconditionally, *before* this `DRY_RUN` branch — so when `DRY_RUN` is set, the budget line is already permanently deleted, and this rollback only discards the newly-added `OpsEvent` + `AgreementHistory`. Net effect of a dry run: the irreversible delete happens *and* the audit record is thrown away, which contradicts the comment above ("DRY_RUN drives both the trigger's commit and our rollback/commit (single source of truth)").

`load_remove_agreements/utils.py` (lines 65-81) has the right shape — it checks `DRY_RUN` *before* `session.delete(...)`. Suggest moving the `delete`/`commit` (and the ops_event build) into the non-dry-run branch so the whole unit of work is a single transaction gated by `DRY_RUN`.

backend/data_tools/src/load_aas/utils.py
```diff
@@ -228,7 +233,7 @@ def create_models(data: AAData, sys_user: User, session: Session) -> None:
         # Set Dry Run true so that we don't commit at the end of the function
         # This allows us to rollback the session if dry_run is enabled or not commit changes
         # if something errors after this point
-        agreement_history_trigger_func(ops_event, session, sys_user, dry_run=True)
+        agreement_history_trigger_func(ops_event, session, sys_user, dry_run=False)
```
josbell, line 233: Stale comment: it says "Set Dry Run true so that we don't commit at the end of the function," but the call now passes `dry_run=False` and is immediately followed by `session.commit()`. Behavior's fine — just update/remove the comment so it doesn't mislead.

backend/data_tools/src/load_aas/utils.py
```diff
@@ -254,6 +259,71 @@ def create_models(data: AAData, sys_user: User, session: Session) -> None:
...
+            aa.alternate_project_officer_id = existing_agreement.alternate_project_officer_id
+            aa.awarding_entity_id = existing_agreement.awarding_entity_id
+            aa.product_service_code_id = existing_agreement.product_service_code_id
+            aa.nick_name = existing_agreement.nick_name
+            aa.service_requirement_type = existing_agreement.service_requirement_type
+            aa.agreement_reason = existing_agreement.agreement_reason
+
+            # Repoint CLINs to the new AA so they survive the agreement delete
+            # (clin.agreement_id has ondelete=CASCADE, so otherwise they'd be deleted).
+            clins = session.execute(select(CLIN).where(CLIN.agreement_id == existing_agreement.id)).scalars().all()
```
josbell, line 270: The four new repoint blocks (CLINs, procurement actions/trackers, agreement history, team members) don't have test coverage — `test_existing_agreement_handling` only sets up budget lines, so each of these loops iterates an empty list and passes vacuously. Since the PR calls out a procurement-tracker FK violation crashing imports in production, a regression test that attaches a CLIN + ProcurementAction/Tracker + team member to the source agreement and asserts they survive (repointed to the new AA, old agreement deleted without an FK error) would lock the fix in.

backend/data_tools/src/load_aas/utils.py
```diff
@@ -254,6 +259,71 @@ def create_models(data: AAData, sys_user: User, session: Session) -> None:
...
+                .scalars()
+                .all()
+            )
+            for team_member_row in team_member_rows:
+                team_member_row.agreement_id = aa.id
+            session.commit()
+
             logger.info(f"Removing existing agreement {existing_agreement.name} with ID {existing_agreement.id}")
             session.delete(existing_agreement)
             session.commit()
```
josbell, line 326: Minor integrity note: `create_models` commits at several points inside the `try` (here and above), but the `except` only logs and re-raises with no rollback. If a later step fails, the earlier commits persist (BLIs moved, new AA created, old agreement still present + partially repointed), leaving a half-migrated state. Not necessarily for this PR, but wrapping the repoint+delete sequence in a single transaction would make re-runs safer.

backend/data_tools/tests/load_aas/test_load_aas.py
```diff
@@ -236,6 +236,48 @@ def test_create_models_upsert(db_for_aas):
...
+    aa_model.project_officer_id = sys_user.id
+    db_for_aas.commit()
+
+    # Re-import the same AA row.
+    create_models(data, sys_user, db_for_aas)
+
+    aa_model = db_for_aas.execute(select(AaAgreement).where(AaAgreement.name == "Family Support Research")).scalar()
+    assert aa_model.awarding_entity_id == proc_shop.id
+    assert aa_model.project_officer_id == sys_user.id
+
```
josbell, line 279: `test_create_models_preserves_existing_fields` asserts `awarding_entity_id` and `project_officer_id`, but the preservation logic also covers `product_service_code_id` and `alternate_project_officer_id` (utils.py:195-198). Worth adding assertions for those two so a future regression on either is caught.

josbell (approved):
Thanks for this — the core fixes are solid: the `(item.amount or 0)` fee null-guard, the BLI-delete history wiring, the "Draft" removal, and the related-record repointing on AA conversion all do what the description says, and the fee-guard + delete-message changes are well tested.

**Approving** so this isn't blocked, and leaving the fixes below to your discretion. One I'd call out as the highest priority: the `DRY_RUN` path in `load_remove_budget_lines` doesn't actually prevent the destructive delete — a dry run still permanently deletes the BLI and only rolls back the audit rows, which is the opposite of a safe preview and contradicts the "single source of truth" comment. Details inline. The rest are minor notes (test coverage for the new repoint blocks, a stale comment, a couple of test assertions).

jonnalley:
@josbell good findings. will look at and address these later in the week

rajohnson90 (approved):

Issue #5762, opened 2026-06-01
Add warning that Read-Only users should have no other roles


rajohnson90:
### User Story
User Admins need to know that if they assign 'read-only' to a user that in order for it to work properly a user should have _only_ the read-only user role. Adding extra roles may distort its permissions. 

### Acceptance Criteria
- [ ] The admin console contains text warning that read-only role needs to be assigned alone to function properly.

-----------------------------------------

#### UX
- [x] Come up with warning to inform user that Read-Only role needs to be specified in a certain way to function properly

#### Dev
- [ ] Implement changes in the User Admin panel per Figma design

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] Usability validated (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] New design/tech debt documented (if applicable)
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

kimschulke:
[We have 2 design options coming up shortly](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=31886-23441&t=YNnhkKV57jvjZmIE-1). We'll definitely need to address the user admin page at some point as its current implementation is not a very good UX and has some usability issues.

npage123:
Figma designs are ready for dev and can be found [here](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=32076-18437&t=WCjo2mBO71xFXqa1-4).

kimschulke:
Hi @rajohnson90 @npage123 I see the designs in Figma, but not seeing this implemented when testing in staging. Let me know if anyone is working on this or we can move back to Sprint Backlog or Product Backlog. I'm also about to update the tags to ready for dev. Thanks!

npage123:
After discussing in standup on 6/26 we decided to move this to the sprint backlog as we have other priorities currently but design is done and it is ready for dev.

jonnalley:
@andressacurvelo @kimschulke @npage123 all changes should be implemented at this time in dev and staging and ready for any review

npage123:
@Santi-3rd @jonnalley Reviewed in staging and all looks good and working great! Thank you!
CC: @andressacurvelo @kimschulke 

https://github.com/user-attachments/assets/d46da6b5-887b-4aff-871b-c2941ad7ca29

PR #5893, opened 2026-07-03
feat: added new modal and button

Santi-3rd:
## What changed

- Added Save Changes and Cancel buttons to the UserInfo cards. 
- Added two modals: A Read-Only role warning and a Cancel warning. 
- Added backend safeguards that ensure that users cannot combine roles with the Read-Only role.

## Issue

#5762 

## How to test

- All tests pass
- log in as Budget team and go to /user-admin
- Change a user's role and test saving and canceling
- Select a user and add the Read-only role combined with other roles. Verify that it triggers the modal and the options work as intended. When saving, it should remove all the roles except the Read-Only role. 
- Try the last step but with multiple users selected. Ensure that saving and cancelling only affects the user that was edited. 

## A11y impact

## Storybook

## Screenshots


frontend/src/components/Users/UserInfo/UserInfo.jsx
```diff
@@ -101,20 +110,73 @@ const UserInfo = ({ user, isEditable }) => {
...
+        });
+        // Record the persisted values so the dirty check resets the buttons to disabled
+        // once the save completes (the `user` prop won't reflect this change).
+        setLastSaved({ division, status, roleNames });
+    };
+
+    // Selecting Read Only requires confirmation because it strips all other roles.
+    // Any other change saves directly.
+    const handleSave = () => {
+        if (selectedRoleNames.includes(READ_ONLY_ROLE)) {
```
weimiao67: **Logic bug — modal fires for existing Read-Only users**

This condition is too broad. If the user already has `READ_ONLY` as their only role and an admin changes only their `division`, clicking Save will open the modal with _"existing role(s) will be removed"_ — even though there are no roles to remove.

The guard should only fire when READ_ONLY is *being newly assigned*:

```js
const handleSave = () => {
    if (selectedRoleNames.includes(READ_ONLY_ROLE) && !persistedRoleNames.includes(READ_ONLY_ROLE)) {
        setShowModal(true);
        return;
    }
    saveUser(selectedRoleNames);
};
```

A test covering this case would be: render with a `baseUser` already holding only `READ_ONLY`, change the division, click Save — assert no modal appears and `mockUpdateUser` is called directly.

weimiao67 (approved):

josbell (changes requested):
## Review: Request Changes

Good overall structure and solid test coverage for the happy paths. Four issues need addressing before merge — three are quick fixes.

---

### Block: Fix before merge

**1. `resolve_roles` can propagate `None` for unknown role names** (`backend/ops_api/ops/services/users.py` line 24)

`session.scalar(...)` returns `None` when no row matches. The list comprehension propagates it silently into the roles list, which will cause a cryptic SQLAlchemy crash rather than a clean 400 response. This is a security-sensitive code path and should fail loudly.

```python
# current — crashes downstream on unknown role name
return [session.scalar(select(Role).where(Role.name == role_name)) for role_name in role_names]

# fix — raise early with a useful message
resolved = []
for name in role_names:
    role = session.scalar(select(Role).where(Role.name == name))
    if role is None:
        raise BadRequest(f"Role '{name}' does not exist.")
    resolved.append(role)
return resolved
```

---

**2. No frontend guard when an already-READ_ONLY user gets a second role added** (`UserInfo.jsx` lines 158–164)

The modal check only fires when transitioning *into* READ_ONLY (`!persistedRoleNames.includes(READ_ONLY_ROLE)`). If a user is already READ_ONLY and an admin adds a second role, the condition is false, the modal is skipped, and `saveUser` calls the backend directly. The backend correctly returns 400, but the frontend handles that by navigating to `/error` — the admin gets a generic error page instead of the inline validation this PR was designed to provide.

Fix: add an upfront guard in `handleSave` that mirrors the backend rule:

```js
const handleSave = () => {
    if (selectedRoleNames.includes(READ_ONLY_ROLE) && selectedRoleNames.length > 1) {
        // show validation error inline instead of relying on backend 400
        return;
    }
    if (selectedRoleNames.includes(READ_ONLY_ROLE) && !persistedRoleNames.includes(READ_ONLY_ROLE)) {
        setShowModal(true);
        return;
    }
    saveUser(selectedRoleNames);
};
```

---

**3. Duplicate `id="save-changes"` when multiple users are selected** (`UserInfo.jsx` line 303)

`UserAdmin.jsx` maps over `selectedUsers` and renders one `UserInfo` per user. Selecting 2+ users produces multiple `<button id="save-changes">` in the same document — invalid HTML that can break CSS selectors, `getElementById`, and assistive-technology focus management. The button already has `data-cy="save-btn"` so the `id` attribute is unused.

Fix: remove `id="save-changes"` from the button element.

---

**4. No backend test for `create_user` enforcing the READ_ONLY exclusivity rule** (`test_users_post.py`)

`resolve_roles` is called in `create_user` too, but `test_users_post.py` has no READ_ONLY combination tests. Worth adding two:
- POST with `["READ_ONLY", "SYSTEM_OWNER"]` → expect 400
- POST with `["READ_ONLY"]` alone → expect 201

**5. `setLastSaved` is called before the mutation resolves** (`UserInfo.jsx` line 152)

On API failure the component navigates to `/error` so the form doesn't get permanently stuck, but the success baseline is recorded optimistically. Consider moving `setLastSaved` into the `isSuccess` branch of the `updateUserResult` effect to keep the dirty-check baseline accurate.

rajohnson90 (approved):

josbell (approved):

Issue #668, opened 2023-02-11
View Project Details


jonnalley:
### User Story
As an OPRE staff member, I want to view all the details of a research or admin/support project so I can learn more about that project and its attributes.

### Preconditions
* Editing a research project is not in scope yet

### Acceptance Criteria
- [ ] User can see how the project is being funded across portfolios for the selected FY (the amount contributed from each portfolio) 
- [x] User can see a funding breakdown by CAN for the selected FY
- [ ] User can see the cumulative funding to date (lifetime funding since the project start/first award) and a breakdown over the last 3 fiscal years
- [x] User can see a description of the project
- [ ] User can see the First Award date, which is the earliest date of any agreement associated to that project
- [ ] User can see a listing of all portfolios associated with the project
- [ ] User can see the CAN IDs/numbers for the project
- [ ] User can see a listing of methodologies and populations for the project
- [ ] User can select and access these details of a project from the original listing of research projects on a portfolio

-----------------------------------------

#### UX
- [x] Update designs for view, edit projects #4210 
- [x] Get feedback from other OPS users
- [x] Refine design based on feedback
- [x] Annotate designs and mark ready for dev

#### Dev
- [ ] Update data model / DB architecture for any new attributes and relationships
- [ ] Populate any sample data needed
- [ ] Create additional backend components (i.e. API endpoints, etc..) to accommodate the AC
- [ ] Create any frontend components or other changes to accommodate the AC
- [x] #5224 
- [x] #5275 
- [x] #5363 

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] New design/tech debt documented (if applicable)
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

### Additional Context & Resources
* On the research projects tab of portfolios, users can see a list of and some limited info about each RP, but this shows all the details of the RP.
* Also see #4210 
* [Designs in Figma](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=27966-34646&t=mzTQQ2SiaLiBnnGI-1)

kimschulke:
Design is ready for development! https://www.figma.com/file/dsQFe6G2KjcTWNq7NNZujm/View-Portfolio-Details?node-id=2287%3A8820&t=r9ps6wF3I5i6so8p-1

[...]

PR #5898, opened 2026-07-06
fix: highlight percentage tags on data viz hover (OPS-668)

weimiao67:
## What changed

Fixes the missing hover highlight on two project-detail data visualizations so the percentage value gets the same background color as the hovered chart segment — matching the behavior of the "Budget Across Portfolios" viz that serves as the reference standard.

**`frontend/src/pages/projects/detail/ProjectSpending.jsx`**
- The "Project Spending By Agreement Type" donut legend now assigns each agreement type its matching `tagStyleActive` (`whiteOnContractBlue`, `darkOnPartnerGreen`, `darkOnGrantOrange`, `whiteOnDirectObligationPink`). These active styles already exist in the `Tag` component, so on hover the percentage tag is filled with the segment color. Previously the legend text bolded but the percentage tag stayed white because agreement-type labels didn't match `Tag`'s hardcoded label cases and no `tagStyleActive` was passed.

**`frontend/src/components/Projects/ProjectFundingByPortfolioCard/ProjectFundingByPortfolioCard.jsx`**
- The "FY Project Funding by Portfolio" legend percentage now takes the segment's `backgroundColor` (with legible light/dark text) when active, mirroring `PortfolioLegend`. Previously the percentage span was hardcoded to `bg-white` and only bolded on hover. Since this card assigns palette colors sequentially rather than by abbreviation, the light/dark text decision keys off a `LIGHT_BACKGROUND_COLORS` set derived from the same light-background portfolio list used by `PortfolioLegend`.

## Issue

https://github.com/HHS/OPRE-OPS/issues/668

## How to test

1. Run the frontend unit tests:
   ```bash
   cd frontend
   bun run test --watch=false ProjectFundingByPortfolioCard
   bun run test --watch=false DonutGraphWithLegendCard LegendItem Tag
   ```
2. Start the app (`podman compose up --build`) and open a Project Details page.
3. On the **Project Spending** tab, hover over the "Project Spending By Agreement Type" donut segments — the corresponding legend percentage tag should be highlighted with the segment's color (text remains legible).
4. On the **Project Funding** tab, hover over the "FY Project Funding by Portfolio" bar segments — the corresponding legend percentage should get a rounded rectangle filled with the segment color, matching the Budget Across Portfolios viz on the portfolios list page.

## A11y impact

## Storybook

### DIFF
--- a/frontend/src/components/Projects/ProjectFundingByPortfolioCard/ProjectFundingByPortfolioCard.jsx
+++ b/frontend/src/components/Projects/ProjectFundingByPortfolioCard/ProjectFundingByPortfolioCard.jsx
@@ -28,6 +28,21 @@ import { buildPortfolioChartData } from "./ProjectFundingByPortfolioCard.helpers
  * @param {FundingByPortfolioItem[]} props.fundingByPortfolio
  * @returns {JSX.Element}
  */
+// Portfolio color vars with light fills that need dark text for readability
+// when used as the active percentage-tag background (mirrors the
+// lightBackgroundPortfolios list in PortfolioLegend, keyed by color instead of
+// abbreviation since this card assigns colors sequentially).
+const LIGHT_BACKGROUND_COLORS = new Set([
+    "var(--portfolio-bar-graph-cc)",
+    "var(--portfolio-bar-graph-hs)",
+    "var(--portfolio-bar-graph-hmrf)",
+    "var(--portfolio-bar-graph-hv)",
+    "var(--portfolio-bar-graph-dd)",
+    "var(--portfolio-bar-graph-none-opre)",
+    "var(--portfolio-bar-graph-ocdo)",
+    "var(--portfolio-bar-graph-otip)"
+]);
+
 const ProjectFundingByPortfolioCard = ({ fiscalYear, fundingByPortfolio = [] }) => {
     const [activeId, setActiveId] = useState(0);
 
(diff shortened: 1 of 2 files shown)


josbell (approved):
Approving — the fix is correct and the hover behavior works as intended. A few things worth following up on:

**Potential accessibility issue**
`ProjectFundingByPortfolioCard.jsx` — `LIGHT_BACKGROUND_COLORS` is missing `"var(--portfolio-bar-graph-dv)"` (`#619ad8`, slot 8 in the sequential palette). Projects with 8+ portfolio types can land this medium-bright blue on a segment; white text on it fails WCAG AA contrast (~2.9:1). Worth adding to the set. The fallback color `var(--data-viz-bl-by-status-1)` (light gray) has the same gap but is lower risk since it only appears for unknown portfolios.

**No tests for the new active-state branch**
Neither the `isActive === true` branch in `ProjectFundingByPortfolioCard` nor the new `tagStyleActiveMap` logic in `ProjectSpending` are covered by tests. `PortfolioLegend.test.jsx` sets the pattern (explicit tests for light vs. dark background text color selection). This leaves the bug we just fixed open to silent regression.

**Duplication worth a follow-up**
The `tagStyleActiveMap` in `ProjectSpending.jsx` is identical to what `AgreementSpendingSummaryCard.jsx` already has inline. The codebase pattern is a constants file (see `PROJECT_TYPE_TAG_STYLE_ACTIVE` in `ProjectTypes.constants.js`). Not blocking, but worth extracting to `AgreementSpendingCards.constants.js` in a follow-up.

rajohnson90 (approved):

Issue #5894, opened 2026-07-03
Sticky Obligate By date validation on editing a BL


npage123:
After getting a validation error on the Obligate By date when attempting to create or edit a budget line as one user role, when I then sign in with a different user role, and navigate to an agreement, the same Obligate By validation error appears by default without doing anything.

## Expected Behavior
The Obligate By date validation error should not appear for different users or on different agreements when navigated to. 

## Steps to Reproduce
1. Log into staging as a DD and go to edit an agreement
2. Create a new BL or edit an existing BL by entering an Obligate By date in the past to trigger the validation error
3. Log out
4. Log into staging as a Budget Team user and go to edit an agreement
5. See the validation error on the Obigate By date picker by default without editing anything

https://github.com/user-attachments/assets/59c72870-21b9-43e8-8a45-c8b95c33a394

weimiao67:
@npage123 the bug is fixed and ready for review and testing.

npage123:
@weimiao67 Reviewed in staging and looks good to me.
https://github.com/user-attachments/assets/44b92b6a-9be4-4326-8364-23643d592000

PR #5903, opened 2026-07-06
fix: reset validation suites on mount/unmount to prevent sticky Obligate By error

weimiao67:
## What changed

Fixes a bug where the **Obligate By** date validation error would "stick" across budget-line forms — appearing by default for a different user or on a different agreement without any interaction.

**Root cause:** The budget-line form uses three Vest validation suites (`datePickerSuite`, `budgetFormSuite`, and the page-level `suite`) declared as **module-level singletons**. Vest retains a suite's last result until `.reset()` is called. Because OPS is a SPA, logging out and back in as a different user never reloads the JS modules, so a failing "Obligate By" result from one user survived into the next session and was painted onto the DatePicker on mount (`BudgetLinesForm.jsx` reads `datePickerSuite.get()` during render).

**Fix** — `CreateBLIsAndSCs.hooks.js` (the sole mount point for `BudgetLinesForm`, covering the create/edit/review-edit/details flows):

- Added a mount + unmount `useEffect` that resets all three suites, so a freshly mounted form always starts with clean validation state. Follows the existing in-repo idiom at `AgreementEditForm.hooks.js`.

- **CreateBLIsAndSCs.hooks.test.js** — added a regression test asserting each suite is reset once on mount and once on unmount.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5894

## How to test

Automated:
```bash
cd frontend
bun run test --watch=false src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.test.js
bun run test --watch=false src/components/BudgetLineItems/
```

Manual (mirrors the issue's repro steps):
1. Log in as one user (e.g. a Division Director), edit an agreement, and create/edit a budget line with an **Obligate By** date in the past to trigger the validation error.
2. Log out.
3. Log in as a different user (e.g. a Budget Team user) and navigate to edit an agreement.
4. Confirm the **Obligate By** date picker shows **no** validation error by default, without editing anything.

## A11y impact

_Logic-only change; no UI markup or component behavior changes._

## Storybook

### DIFF
--- a/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
+++ b/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
@@ -109,6 +109,19 @@ const useCreateBLIsAndSCs = (
     const activeUser = useSelector((state) => state.auth.activeUser);
     const isSuperUser = activeUser?.is_superuser ?? false;
 
+    // Reset validation suites on mount and unmount so stale results from a
+    // previous agreement or user session never paint errors on a fresh form. (issue #5894)
+    React.useEffect(() => {
+        suite.reset();
+        budgetFormSuite.reset();
+        datePickerSuite.reset();
+        return () => {
+            suite.reset();
+            budgetFormSuite.reset();
+            datePickerSuite.reset();
+        };
+    }, []);
+
     React.useEffect(() => {
         if (currentStep != 0) {
             setBlockerDisabledForCreateAgreement(true);
--- a/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
+++ b/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
@@ -109,12 +109,22 @@ const useCreateBLIsAndSCs = (
     const activeUser = useSelector((state) => state.auth.activeUser);
     const isSuperUser = activeUser?.is_superuser ?? false;
 
+    // Snapshot the page-suite result in state. The suites are module-level singletons
+    // read during render (here and in BudgetLinesForm), so a stale result from a prior
+    // session/agreement can be painted before the reset effect below fires. Storing the
+    // result in state lets the effect call setState after reset(), which schedules a
+    // re-render with clean validation instead of relying on an incidental re-render. (issue #5894)
+    const [pageSuiteResult, setPageSuiteResult] = React.useState(() => suite.get());
+
     // Reset validation suites on mount and unmount so stale results from a
     // previous agreement or user session never paint errors on a fresh form. (issue #5894)
     React.useEffect(() => {
         suite.reset();
         budgetFormSuite.reset();
         datePickerSuite.reset();
+        // Force a re-render with the freshly-cleared state so the render-time reads
+        // (this hook's `res` and BudgetLinesForm's datePickerSuite.get()) repaint clean.
+        setPageSuiteResult(suite.get());
         return () => {
             suite.reset();
             budgetFormSuite.reset();
@@ -155,11 +165,14 @@ const useCreateBLIsAndSCs = (
     }, [tempBudgetLines, servicesComponents]);
 
     // Validation
+    // Review mode re-runs the suite every render against the current budget lines.
+    // Non-review mode reads the state-backed snapshot so a mount-time reset() repaints
+    // clean instead of surfacing a stale singleton result. (issue #5894)
     const res = isReviewMode
         ? suite.run({
               budgetLines: tempBudgetLines
           })
-        : suite.get();
+        : pageSuiteResult;
     const pageErrors = res.getErrors();
     // Filter page errors to only include "Budget line item" errors and consolidate into single message
     const budgetLineErrors = Object.entries(pageErrors).filter((error) => error[0].includes("Budget line item"));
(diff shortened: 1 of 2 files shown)


josbell (changes requested):
## Review: Incomplete fix for SPA navigation case

The fix correctly solves the logout/hard-refresh repro in the issue (module reload clears the singleton), but leaves the more common case broken: **clicking a React Router `<Link>` from one agreement to another still shows stale errors on first render.**

### Root cause of the gap

`useEffect` fires *after* the browser has committed the DOM. But the suites are read *during* render — before the effect has a chance to fire:

- `BudgetLinesForm.jsx` line 60: `let dateRes = datePickerSuite.get();` — top-level, synchronous
- `BudgetLinesForm.jsx` lines 135, 155–157, 177, 193: four `budgetFormSuite.getErrors()` calls in JSX
- `CreateBLIsAndSCs.hooks.js` ~line 162: `suite.get()` in non-review mode (synchronous render path)

Because `suite.reset()` has no React state side effect, no re-render is scheduled after it fires. The stale errors computed on the first render stay painted until the next user interaction.

**Concrete scenario:** User edits a BLI on Agreement A, triggers a past-date error, navigates via a `<Link>` to Agreement B. React renders Agreement B while Agreement A is still mounted (React 18 concurrent transitions). `datePickerSuite.get()` returns Agreement A's errors. The Add BLI button is disabled; field-level error labels appear on untouched fields. They persist until the user interacts with any field.

### Suggested fix

Store the initial suite result in `useState` so calling `setState` inside the effect schedules a re-render with clean state:

```js
// In CreateBLIsAndSCs.hooks.js
const [suiteResult, setSuiteResult] = React.useState(() => suite.get());

React.useEffect(() => {
    suite.reset();
    budgetFormSuite.reset();
    datePickerSuite.reset();
    setSuiteResult(suite.get()); // forces re-render with clean state
    return () => {
        suite.reset();
        budgetFormSuite.reset();
        datePickerSuite.reset();
    };
}, []);
```

Then derive `res` / `pageErrors` from `suiteResult` instead of calling `suite.get()` directly in the render body.

### Test gap

The new test only asserts `.reset()` call counts. The suite mocks unconditionally return `{ hasErrors: () => false }` regardless of whether `reset()` was called — so the test would pass even if the fix were removed. A meaningful regression test should:

1. Pre-seed the suite mock to return stale errors before mount
2. Mount the hook
3. Assert `result.current.budgetLinePageErrorsExist === false` after the effect settles

### Nit

`toHaveBeenCalledTimes(1)` silently relies on `beforeEach` calling `vi.clearAllMocks()`. `resetForm` (and `handleSave`) also call `suite.reset()`, so a future test added before this one that triggers those handlers could inflate the count. A delta assertion (`const before = suite.reset.mock.calls.length; ...; expect(suite.reset).toHaveBeenCalledTimes(before + 1)`) would be robust to ordering.

### Note on prior art

`AgreementEditForm.hooks.js` has the same timing gap (same `useEffect` reset + synchronous `suite.get()` in render). That's a pre-existing issue worth a follow-up, but the BLI fix should be corrected in this PR since that's the stated scope.

Santi-3rd (approved):

weimiao67:
@josbell Thanks — great catch on the render-vs-effect timing. You're right that the suites are read during render (`datePickerSuite.get()` at `BudgetLinesForm.jsx:60/156`, `suite.get()` in the hook) before the reset effect fires, so a stale singleton result can appear on the first render. I've applied your `useState` suggestion in `f722886`:

- `pageSuiteResult` is now stored in state and set via `setPageSuiteResult(suite.get())` inside the mount-reset effect, so the render-time reads repaint against cleared state deterministically instead of relying on an incidental re-render.
- Non-review `res` derives from `pageSuiteResult`; review mode still re-runs the suite each render (unchanged).

On the **test gap** — fully agreed the old assertions only checked `reset()` call counts. Added a behavioral test that models the singleton as *actually stateful* (`get()` returns stale Budget line item errors until `reset()` clears it), mounts the hook, and asserts `budgetLinePageErrorsExist === false` / `pageErrors === []` after the effect settles. Confirmed it fails (waitFor times out) if the reset effect is removed, so it's a real guard.

One empirical note for the record: before this change I instrumented the real hook + real suites, seeded a stale past-date error, and logged every render. The stale error appeared only on render 1 and was cleared by the committed render — because the subsequent `setTempBudgetLines` effect forced a re-render within the same commit batch, so under normal (non-transition) navigation it was a single-render flash rather than a persistent error. Your fix still improves this by making the repaint explicit/deterministic rather than incidental, so I've taken it regardless.

Re **prior art** (`AgreementEditForm.hooks.js`): agreed it has the same pattern — tracking as a follow-up, out of scope here. Left the `toHaveBeenCalledTimes` nit as-is since the new behavioral test is the substantive coverage; happy to convert to a delta assertion if you'd prefer.

josbell (approved):
Reviewed by two independent agents + a deep investigation pass. The core fix is correct — resetting the Vest singletons in a mount/unmount effect and snapshotting `pageSuiteResult` into React state is the right approach for this class of stale-singleton bug.

Three minor items surfaced during review, all low severity and none blocking:

1. **`BudgetLinesForm` reads `datePickerSuite.get()` directly on first render** — the stale flash is real in code but resolves before user interaction via the indirect re-render triggered by `setPageSuiteResult`. The hook comment acknowledges this. Worth a follow-up to give `datePickerSuite` its own state slot for explicitness, but not urgent.

2. **`resetForm()` resets the suite singleton without calling `setPageSuiteResult`** — not a real bug. `suite.run()` is only called in review mode, so `pageSuiteResult` in non-review mode is set at mount and never needs updating. Harmless inconsistency.

3. **Call-count assertions in the new test are StrictMode-fragile** — `toHaveBeenCalledTimes(1/2)` would break if `wrapper: React.StrictMode` were added to the harness (production runs in StrictMode; tests don't). The behavioral assertion in the second new test is the better pattern to follow.

Issue #5910, opened 2026-07-08
Unsaved changes modal appears when saving an agreement draft


npage123:
## Expected Behavior
Just the success alert should appear messaging that a draft has been successfully saved.

## Current Behavior
The 'Unsaved Changes' modal appears with the success alert showing up behind the transparent background.

## Steps to Reproduce
1. Log into staging
2. Select 'Create' from the main nav and then select 'Agreement'
3. Select 'Save Draft' from any of the three create agreement steps
4. See 'Unsaved Changes' modal appearing with success alert behind transparent background

## Context
https://github.com/user-attachments/assets/0edc1334-b7fb-4029-8573-3bbd0d9b3200

weimiao67:
@npage123 the bug is fixed and ready for review/test.

npage123:
Successfully tested in staging. Thanks @weimiao67 !

https://github.com/user-attachments/assets/ba702344-a634-46b6-b3d1-f14d12f5051a

PR #5913, opened 2026-07-08
fix: bypass navigation blocker when saving agreement draft

weimiao67:
## What changed

Fixes the "Unsaved Changes" modal incorrectly appearing when a user clicks **Save Draft** in the Create Agreement wizard.

`handleDraft` (in `AgreementEditForm.hooks.js`) previously called `navigate("/agreements")` synchronously after saving. The `useNavigationBlocker` hook blocks any pathname change while `hasChanged && !isCancelling`, so this tripped the blocker and rendered the "You have unsaved changes" modal — with the success alert visible behind its transparent backdrop.

Two reasons it fired:
1. `setHasAgreementChanged(false)` only reset the parent's copy. The blocker's `hasChanged` is bound to the local `useHasStateChanged(agreement)` value, which stays `true` because the agreement still differs from its mount snapshot.
2. `handleDraft` never set the `isCancelling` bypass flag (unlike `handleCancel`, which does).

The fix makes `handleDraft` mirror the proven `handleCancel` pattern:
- Calls `setIsCancelling(true)` to bypass the blocker.
- Defers navigation via the success alert's `redirectUrl: "/agreements"` instead of a synchronous `navigate()`, so a re-render propagates `isCancelling=true` before `useBlocker` re-evaluates (its predicate is captured at render).
- New draft: adds `redirectUrl` to the "Agreement Draft Saved" alert.
- Existing draft: calls `saveAgreement("/agreements")` so its own success alert carries the redirect.
- Keeps a direct `navigate()` only for the no-op case (existing agreement, no changes) where the blocker is already inactive.
- Error path unchanged.

Only step 2 (Agreement) of the wizard was affected: step 1 (Select Project) has no Save Draft button, and step 3 (Budget Lines) uses a "Create Agreement" button whose blocker is already disabled past step 0 and navigates via `redirectUrl`.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5910

## How to test

1. Log into staging (or run locally).
2. Select **Create** from the main nav, then **Agreement**.
3. Fill out enough to enable **Save Draft** and click it on the Agreement step.
4. Confirm only the success alert appears ("Agreement Draft Saved") and you are redirected to the Agreements list — the "Unsaved Changes" modal should **not** appear.

## A11y impact

## Storybook

### DIFF
--- a/frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js
+++ b/frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js
@@ -529,7 +529,14 @@ const useAgreementEditForm = (
         }
 
         try {
-            const result = await saveAgreement();
+            // Bypass the navigation blocker: saving a draft is an intentional, saved exit,
+            // so the "You have unsaved changes" modal must not appear. Note that
+            // setHasAgreementChanged(false) only resets the parent's copy — the blocker's
+            // hasChanged is bound to the local useHasStateChanged(agreement) value, which
+            // stays true here. Navigation is deferred to the success alert's redirectUrl so a
+            // re-render propagates isCancelling before useBlocker re-evaluates.
+            setIsCancelling(true);
+            const result = await saveAgreement("/agreements");
             if (result === false && !agreement.id) {
                 const data = {
                     ...agreement,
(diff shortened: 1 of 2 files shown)


jonnalley (approved):
The handleDraft handler now mirrors the proven handleCancel pattern: it sets
  setIsCancelling(true) to bypass the useBlocker predicate and defers navigation to the success alert's redirectUrl rather than calling navigate() synchronously. The two-part mechanism is genuinely load-bearing and the three navigation paths are mutually exclusive and each navigate exactly once.

rajohnson90 (approved):
Looks good to me. Manual testing passes as well

Issue #4148, opened 2025-07-25
Implement OPS User Metrics for UXers


kimschulke:
### Goals
Implement OPS User Metrics for UXers

### Tasks
- [ ] Determine what metrics can be provided without a lot of effort
- [ ] Determine how/when/where metrics can be shared

### Additional Context
* UXers need some visibility into OPS use so they have better understanding into OPS users, context and how to tailor communications 

### Resources
* TBD

kimschulke:
starting to think through this some more - brainstorm ideas for what metrics would be helpful

weimiao67:
# Proposal — OPS Use Metrics for UXers

## The problem
UXers need visibility into how OPS is actually used — who the users are, what they do, and where they spend time — so they can tailor communications and design decisions. Today they have no way to see this.

## What we can deliver with low effort
OPS already records a rich activity stream in the production database (the `ops_event` log) — we don't need to build new tracking to get started. Existing data can answer:

- **Who's using OPS** — active users over time (daily/weekly/monthly) and login frequency, broken down by **division** and **role**.
- **What they do** — counts of key actions we actually log today: agreements (created / edited / viewed), budget lines (created / edited), CANs (created / edited), projects (created / edited).
- **User lifecycle** — new users, and deactivations (a user's status flipping to inactive).
- **Session behavior** — how often sessions time out vs. users logging out deliberately (a UX signal about session length).

What we **can't** get easily: fine-grained UI interactions (clicks, scrolls, which buttons). That would need new frontend instrumentation — a good follow-up, but out of scope for a first version. A couple of the actions we *don't* currently log as distinct events (e.g. "CAN viewed", "project viewed") simply won't appear in v1; we'd report only what the activity log actually captures.

## The key constraint that shaped this
**UXers don't have production access.** That rules out a dashboard *inside* OPS, because they can't log into the environment where the real data lives. So instead of UXers pulling a report from the app, the system has to **push the report out** to somewhere they can already reach.

## Options we considered

| Option | Self-serve? | Effort | Verdict |
|---|---|---|---|
| **Dashboard page inside OPS** | ❌ needs prod login | High | **Rejected** — UXers can't access prod |
| **Email the report weekly** | ✅ arrives in inbox | High | **Rejected for v1** — OPS has no email capability today; net-new infra + secrets + approvals |
[...]

PR #5960, opened 2026-07-20
feat: add scheduled usage metrics report job for UX team (OPS-4148)

weimiao67:
## What changed

Adds the first *scheduled* job in the repo: an Azure Container App Job that runs on the **last Friday of each two-week sprint**, aggregates existing `ops_event` activity into a usage report, uploads it to Azure Blob Storage, and emails the UX team a time-limited download link. The report is delivered as a single **two-sheet `.xlsx`** each run:

- an **"Aggregate"** sheet — per-day × division × role counts (logins, logouts, active users, agreements viewed, BLIs created, deactivated users, etc.), and
- a **"Per-user"** sheet — each named user who signed in during the reporting window (`name, email, division, roles, sign_in_count, last_sign_in_utc`).

> **Privacy note:** the per-user sheet names individual users — an explicit, approved requirement change (#4148) that reverses the original counts-only posture. IP addresses are excluded from both sheets. Read access to the `reports/` prefix should be granted with the awareness that the `.xlsx` contains named user data.

### Aggregation & report

- **`backend/data_tools/src/usage_metrics/utils.py`** — aggregation module. Reads `ops_event` rows within a lookback window (SUCCESS-only), attributes each event to a division/role, and buckets counts. `aggregate_user_sign_ins` builds the per-user view: sign-ins are counted on the resolved actor id **only** (not fanned out per role) from SUCCESS `LOGIN_ATTEMPT` events over the same lookback window, falling back to the `event_details` snapshot for users no longer in the live table. `build_workbook` renders both sheets with openpyxl directly (no pandas). Uploads the workbook as dated + stable `-latest` blobs (two blobs), or writes it locally when no storage account is configured.
- **`backend/data_tools/src/azure_utils/utils.py`** — adds `upload_blob` (managed identity / RBAC) mirroring the existing download path, with an optional `content_type` so the `.xlsx` is uploaded with the spreadsheet MIME type (clean browser download via SAS link), and `build_blob_sas_url`, which mints the read-only SAS download link.
- **`backend/data_tools/src/disable_users/disable_users.py`** — the automated deactivation path now writes `status: INACTIVE` into its `ops_event` details (mirroring the manual UI payload) so the usage-metrics `deactivated_users` detector counts automated deactivations too.

### Schedule (last Friday of each sprint)

Cron cannot express "every other Friday" — there is no week-parity field, and restricting day-of-month alongside day-of-week makes standard cron parsers **OR** the two fields rather than AND them. So the job's cron is `50 23 * * 5` (23:50 UTC **every** Friday = Friday evening US Central, after the workday) and `should_generate_report` in `src/usage_metrics/utils.py` exits early on the Fridays that do not end a sprint — logging why, and touching neither the database nor Blob storage.

- **`USAGE_METRICS_SPRINT_ANCHOR_DATE`** (default `2026-09-11`) is a known sprint-end Friday; sprint ends are every 14 days from it in both directions. It's validated as a Friday at run time, because a non-Friday anchor would never line up with the cron and the job would silently no-op forever. The default came from the team's own boundaries — the sprint 106 and 107 release-note commits landed Friday 2026-08-28 and Friday 2026-09-11, exactly 14 days apart. A boundary shift needs only an env-var update, no code change.
- **`USAGE_METRICS_LOOKBACK_DAYS` moves 7 → 14** to match the sprint length, so consecutive reports tile the calendar with no gap or overlap.
- **`USAGE_METRICS_FORCE_RUN=true`** bypasses the filter, which is how the job gets test-fired off-schedule; local/dev/pytest configs default it to `True` so a manual run always produces a report.

**Action item at merge:** the deploy workflows update only the job'''s *image*, and cron/env vars are applied at job creation — so the existing staging job keeps its old Monday cron. Since a Monday can never match the Friday sprint anchor, it would silently skip every run until someone applies the one-time `az containerapp job update` in the runbook ("Rescheduling a job that already exists"). Production is unaffected; its job has not been created yet.

Two consequences documented in the runbook: about half the scheduled runs are deliberate no-ops that still show as `Succeeded` in `az containerapp job execution list` (check the logs for the skip line before treating one as a missed report), and activity after ~18:50 Central on the sprint's last Friday lands in the next sprint's report rather than being lost.

### Email delivery (SAS link)

- **`backend/data_tools/src/usage_metrics/email_delivery.py`** — after the workbook is safely in Blob storage, `deliver_report_link` generates a time-limited **SAS download link** to that sprint's dated report and emails it to the UX team via **Azure Communication Services (ACS)**. `parse_recipients` splits/de-dupes the comma-separated recipient list. The report is already persisted before this runs; a send failure is logged and raised so the run surfaces it.
- **Both secrets are read from Key Vault by the job's managed identity at run time — neither is stored on the job.**
  - **ACS auth uses the connection string**, not AAD/RBAC. The ACS resources are provisioned by [OPRE-OPS-Data#59](https://github.com/HHS/OPRE-OPS-Data/pull/59), which distributes the connection string as a **Key Vault secret** per environment and leaves RBAC out of scope; ACS's AAD data-plane auth would require the **`Contributor`** role on the ACS resource (there is no narrower built-in send role), which the job's MI does not hold.
  - **The SAS is signed with the storage account key**, not a user-delegation SAS, because Azure caps MI-signed user-delegation SAS at **7 days** and the link needs ~90.
- **One-time Azure prerequisite:** `opre-ops-stg-app-kv` uses **access policies** (not RBAC) and the `storageAccountUser` MI is not in its policy list, so it must be granted `get` on secrets before email delivery can work (one grant covers both secrets). Steps are in the runbook.
- Configured via `USAGE_METRICS_ACS_CONNECTION_STRING_SECRET`, `USAGE_METRICS_EMAIL_SENDER`, `USAGE_METRICS_EMAIL_RECIPIENTS`, `USAGE_METRICS_SAS_EXPIRY_DAYS` (default 90), plus `VAULT_URL` / `VAULT_FILE_STORAGE_KEY`. Email delivery **no-ops** (report still uploads to Blob) unless the ACS secret name, sender, and recipients are all set — so local/dev/staging runs stay silent until wired.

### Config, scripts & CI

- **`backend/data_tools/environment/*.py` + `types.py`** — adds `usage_metrics_*` config properties to `DataToolsConfig` and each environment (storage account/container/prefix, lookback days, sprint anchor date + force-run, SAS expiry, and the ACS connection-string secret name/sender/recipients). Azure reads env vars; local/dev/test use defaults.
- **`backend/data_tools/scripts/azure/create_usage_metrics_job.sh`** — one-time script to create the scheduled job. Configured to match how the real staging jobs run (public ghcr.io image, DB password as a container-app secret, shared user-assigned MI for Blob write). Validates required env vars up front so a missing password fails immediately rather than at cron time.
- **`backend/data_tools/scripts/usage_metrics.sh`** — wrapper the job executes.
- **`backend/data_tools/scripts/azure/USAGE_METRICS_JOB.md`** — the full "Scheduled Usage Metrics Report Job" runbook (create/deploy, verified staging + prod values, ACS/Key Vault wiring, privacy & semantics notes). Split out of `data_tools/CLAUDE.md`, which now carries a short pointer to it.
- **`.github/workflows/stg_be_build_and_deploy.yml`** and **`.github/workflows/prod_be_build_and_deploy.yml`** — update the `usage-metrics-job` image on each merge / prod deploy (guarded so it's a no-op until the job exists).
- **`backend/data_tools/Pipfile` / `Pipfile.lock`** — adds `openpyxl` (workbook rendering) and `azure-communication-email` (email delivery); pins `greenlet` to keep it in the arm64 relock. Relocked.

### Tests

- **`backend/data_tools/tests/usage_metrics/test_utils.py`** — unit tests for aggregation, attribution, per-user sign-in counting (incl. multi-role no-double-count, failed/out-of-window exclusion, deleted-user fallback), workbook shape, xlsx-only upload, and the sprint-schedule guard (anchor parsing/Friday validation, sprint-end detection before and after the anchor, off-sprint skip, force-run bypass).
- **`backend/data_tools/tests/usage_metrics/test_email_delivery.py`** — recipient parsing, email message shape, the connection-string send path, and that the connection string is never logged.
- **`backend/data_tools/tests/azure_utils/test_utils.py`** — `upload_blob`, `build_blob_sas_url` (shape + non-positive expiry rejection), and `get_secret` MI client-id pinning.

### Verified on staging

The job was created and test-fired against staging (`opre-ops-stg-app-rg`): it aggregated ~4,200 `ops_event` rows and uploaded the workbook to `data/reports/`. Fixes that came out of verification:

- **CSV dropped in favor of xlsx-only.** The `.xlsx` "Aggregate" sheet already carries the same per-day × division × role counts the CSV did, so the CSV was redundant. `build_csv`/`generate_report_csv` and the CSV uploads are removed; each run now produces two blobs (`usage-metrics-latest.xlsx` + dated) instead of four.
- **`create_usage_metrics_job.sh` `--args` fix.** The script passed the command as one comma-quoted string (`"/bin/ash, -c, ./…"`), which `az` stored as a single argument — the container exited 128 (StartError) with no logs. Now passed as separate tokens (`/bin/sh` + script path), matching the working data-tools jobs.
- **Deactivation metric counting fix.** The automated `disable_users` path wasn't writing the `status` field the `deactivated_users` detector keys on, so automated deactivations were undercounted; corrected (see `disable_users.py` above).
- **Schedule changed from weekly to per-sprint.** The job originally ran `50 4 * * 1` (Sunday night Central) with a 7-day window; the team wants one report per sprint, on the sprint's last Friday, so the cron, the lookback window, and the new sprint filter replace it.
- **ACS auth reworked to the Key Vault connection string.** The first cut authenticated ACS with the job's managed identity (`USAGE_METRICS_ACS_ENDPOINT` + `DefaultAzureCredential`); that path needs `Contributor` on the ACS resource, which the infra repo does not grant. Switched to the connection-string secret it *does* provision.

## Issue

https://github.com/HHS/OPRE-OPS/issues/4148

## How to test

Run the unit tests:

```bash
cd backend/data_tools
pipenv run pytest tests/usage_metrics/ tests/azure_utils/test_utils.py -v
```

Enabling / deploying on staging: see the runbook at **`backend/data_tools/scripts/azure/USAGE_METRICS_JOB.md`** (`data_tools/CLAUDE.md` → "Scheduled Usage Metrics Report" points to it). In brief:

```bash
export PGUSER=ops PGHOST=opre-ops-stg-db-pg-server.postgres.database.azure.com PGPORT=5432 PGDATABASE=postgres
export PGPASSWORD='<ops DB password — the pgpassword secret on the other staging jobs>'
export USAGE_METRICS_STORAGE_ACCOUNT_URL="https://opreopsstgappsa.blob.core.windows.net"

# optional — email delivery (needs the one-time Key Vault access-policy grant in the runbook):
export VAULT_URL="https://opre-ops-stg-app-kv.vault.azure.net/"
export VAULT_FILE_STORAGE_KEY='<secret name holding the storage account key>'
export USAGE_METRICS_ACS_CONNECTION_STRING_SECRET="opre-ops-sdlc-comms-acs-connection-string"
export USAGE_METRICS_EMAIL_SENDER="DoNotReply@7b9d729e-13e1-43ab-b12d-fa0ea1793a56.azurecomm.net"
export USAGE_METRICS_EMAIL_RECIPIENTS="ux1@example.gov,ux2@example.gov"

./scripts/azure/create_usage_metrics_job.sh opre-ops-stg-app-rg storageAccountUser opre-ops-stg-app-cae

# test-fire without waiting for the sprint-end Friday (force the run, then unset it afterwards):
az containerapp job update -n usage-metrics-job -g opre-ops-stg-app-rg --set-env-vars USAGE_METRICS_FORCE_RUN=true
az containerapp job start -n usage-metrics-job -g opre-ops-stg-app-rg
az containerapp job update -n usage-metrics-job -g opre-ops-stg-app-rg --set-env-vars USAGE_METRICS_FORCE_RUN=""
```

Then confirm the report appears at `data/reports/usage-metrics-latest.xlsx` (the two-sheet workbook with the Aggregate and Per-user sheets), and — if the ACS email vars are configured — that the UX team receives the SAS download link (first send may land in junk; see the deliverability notes in the runbook).

## A11y impact

- [ ] Test coverage maintained/improved
- [ ] Security tests passed


josbell (changes requested):
## Review findings

### 🔴 Critical — `create_usage_metrics_job.sh:76`: Bash `#` comment truncates the `az` command

The multi-line comment block (lines 76–81) is embedded *inside* the `\`-continued `az containerapp job create` invocation. In bash, when a `\<newline>` continuation is consumed, `#` on the next physical line starts a comment that ends at the physical EOL — and since line 76 has no trailing `\`, the continuation chain stops there. The `az` command is submitted after `--cron-expression "50 4 * * 1"` only, missing `--args`, `--environment`, `--mi-user-assigned`, `--secrets`, and `--env-vars`. Azure CLI requires `--environment` and will fail; with `set -euo pipefail` the script exits immediately and the job is never created. This means the script cannot currently be used to provision the job on any environment.

**Fix:** Move the comment block above the `az containerapp job create` call (before line 68), then remove lines 76–81 from inside the continued command.

---

### 🟠 High — `utils.py:197`: `is_deactivating_update` silently excludes all automated bulk deactivations

`is_deactivating_update` detects a deactivation by checking `event_details.get("request.json")` for a `status` field. This only matches manual UI-driven deactivations (where the HTTP request payload is captured under `request.json`). The automated batch deactivation path in `disable_users.py` creates `UPDATE_USER` events with `event_details={"user_id": ..., "message": "User deactivated via automated process."}` — no `request.json` key. `details.get("request.json")` returns `None`, and the function returns `False` for every automated deactivation. In a production environment where the weekly disable_users job is the primary deactivation mechanism, the `deactivated_users` metric will systematically undercount with no error or warning.

**Fix:** Also check for the automated deactivation format, e.g. look for the `message` key containing `"deactivated"`, or check `event_details.get("user_id")` combined with absence of `request.json`.

---

### 🟡 Medium — `utils.py:366`: `parse_lookback_days` accepts `0`, silently producing an empty report

`parse_lookback_days` validates that the configured value is a parseable integer but does not check `> 0`. With `lookback_days=0`, the cutoff is `datetime.now()` and `OpsEvent.created_on >= cutoff` matches no historical rows. The job succeeds and uploads a valid `.xlsx` with column headers but no data rows — no error or warning is raised. An operator who accidentally sets `USAGE_METRICS_LOOKBACK_DAYS=0` would receive a blank report from the UX team with no indication the configuration was wrong.

**Fix:** Add `if days <= 0: raise ValueError(f"usage_metrics_lookback_days must be > 0, got {days}")` after the `int()` conversion in `parse_lookback_days`.

weimiao67:
Thanks for the thorough review — all three findings were valid and are now fixed in c94cc0685.

**🔴 #1 — bash comment truncating the `az` command:** Confirmed (reproduced the truncation in a scratch script). Moved the `--args` explanatory comment block above the `az containerapp job create` call and removed it from inside the `\`-continued invocation, so all flags (`--args`, `--environment`, `--mi-user-assigned`, `--secrets`, `--env-vars`) are now passed. `bash -n` passes.

**🟠 #2 — automated deactivations undercounted:** Fixed at both ends. The `disable_users` job now emits an explicit top-level `"status": UserStatus.INACTIVE.name` on its `UPDATE_USER` event (symmetric with the manual UI path's `request.json.status`), and `is_deactivating_update` now reads `status` from either `event_details['request.json']` (manual) or top-level `event_details['status']` (automated). Chose the explicit-status approach over string-matching `"message"` to avoid brittleness. Note: automated events already written to the DB won't be counted until they age out of the lookback window — only events emitted after this deploys carry the new `status`.

**🟡 #3 — `parse_lookback_days` accepting `0`:** Added a `days <= 0` guard raising `ValueError` (also catches negatives), so a misconfigured window fails fast instead of silently producing an empty report.

Added unit tests for the automated-format detection and the non-positive lookback rejection. `pytest tests/usage_metrics/ tests/disable_users/` → 29 passed.

jonnalley:
@josbell can you please re-review so we can avoid stale PRs? Thank you!

josbell:
Two additional bugs found during review (also closing issues #6122/#6123/#6124 that were incorrectly auto-filed — see those for context):

---

**🟠 `deactivated_users` overcounted for multi-role admins**

In `aggregate_events`, the `is_deactivating_update` check fires inside the `for role in attribution["roles"]` fan-out loop:

```python
for role in attribution["roles"]:
    key = (date_iso, attribution["division"], role)
    ...
    if is_deactivating_update(event):
        counts[key]["deactivated_users"] += 1  # fires N times for N roles
```

An admin holding 2 roles who deactivates one user produces 2 increments across 2 role buckets — the report shows 2 deactivated users for 1 real event. The `active_users` metric avoids this correctly by accumulating into a `set` outside the loop. `deactivated_users` needs the same treatment: track deactivating events in a separate set keyed on `(actor_id, date, division)` outside the role loop, then fold in after.

---

**🟡 `parse_lookback_days` accepts `0`, silently producing an empty report**

`parse_lookback_days` validates that the value is a parseable integer but not that it is `> 0`. With `lookback_days=0`, the cutoff equals `datetime.now()` and `OpsEvent.created_on >= cutoff` matches no historical rows. The job exits successfully and uploads a valid `.xlsx` with headers but no data — no warning or error is raised. A simple guard after the `int()` conversion fixes it:

```python
if days <= 0:
    raise ValueError(f"usage_metrics_lookback_days must be > 0, got {days}")
```

jonnalley:
@weimiao67 One thing that stands out to me here is that the existing backend/data_tools/CLAUDE.md file is already over Anthropic's recommended max size for a CLAUDE.md file.  We probably don't want to extend and make it even bigger. would recommend we condense or split into separate files. 
https://code.claude.com/docs/en/memory.md

weimiao67:
@josbell Reviewed both findings against current HEAD — no code change needed. Summary of why:

**🟡 `parse_lookback_days` accepts `0`** — legitimate concern, but **already fixed** (stale finding). The exact guard proposed already exists at `utils.py:379-381`:

```python
if days <= 0:
    raise ValueError(f"usage_metrics_lookback_days must be > 0, got {days}.")
return days
```

Covered by `test_parse_lookback_days_non_positive_raises`. Landed in `c94cc0685`, which predates this comment.

**🟠 `deactivated_users` overcounted for multi-role admins** — the symptom is real but misdiagnosed, and the proposed fix would introduce a new bug:

1. **It's intended, documented, and universal.** Every event-count metric (`logins`, `logouts`, `new_users`, `agreements_edited`, `blis_created`, `projects_created`) fans out per actor role in the same loop. The module docstring (`utils.py:37-39`) explicitly documents that role sums may exceed division totals and are left unresolved by design. Within a single `(date, division, role)` bucket the count is correct; only the cross-role sum "doubles." Nothing here is specific to `deactivated_users`.

2. **The `active_users` comparison is inaccurate.** `active_users[key].add(actor_id)` is *inside* the role loop (`utils.py:250`) and keyed per role, so `active_users` fans out across roles too. Only the `len()` fold (`utils.py:253-254`) is outside the loop. The set dedups repeated events by the same actor *within one bucket* — it does not prevent role fan-out.

3. **Different semantics; proposed fix undercounts.** `active_users` = distinct active users (dedup by actor is correct). `deactivated_users` = count of deactivation actions (parallel to `new_users`; each should count). The suggested set keyed on `(actor_id, date, division)` keys on the *actor*, so an admin deactivating 5 different users in one day would collapse to 1 — a real undercount. Any dedup would have to key on the *target* user, not the actor.

Net: #2 is already implemented + tested; #1 rests on an inaccurate reading of `active_users` and proposes a change that trades documented/intended fan-out for an undercount. Leaving the code as-is.

josbell (approved):

rajohnson90 (approved):
Looks good. Had an AI review with a little bit of suggestions that I implemented.

backend/data_tools/scripts/azure/create_usage_metrics_job.sh
```diff
@@ -0,0 +1,145 @@
...
+    PGHOST="${PGHOST}" \
+    PGPORT="${PGPORT}" \
+    PGDATABASE="${PGDATABASE}" \
+    USAGE_METRICS_STORAGE_ACCOUNT_URL="${USAGE_METRICS_STORAGE_ACCOUNT_URL}" \
+    USAGE_METRICS_CONTAINER_NAME="${USAGE_METRICS_CONTAINER_NAME:-data}" \
+    USAGE_METRICS_REPORT_PREFIX="${USAGE_METRICS_REPORT_PREFIX:-reports}" \
+    USAGE_METRICS_LOOKBACK_DAYS="${USAGE_METRICS_LOOKBACK_DAYS:-14}" \
+    USAGE_METRICS_SPRINT_ANCHOR_DATE="${USAGE_METRICS_SPRINT_ANCHOR_DATE:-2026-09-11}" \
+    USAGE_METRICS_FORCE_RUN="${USAGE_METRICS_FORCE_RUN:-}" \
+    USAGE_METRICS_ACS_CONNECTION_STRING_SECRET="${USAGE_METRICS_ACS_CONNECTION_STRING_SECRET:-}" \
```
jonnalley, line 140: @weimiao67 would recommend considering making the name of this secret something more generic to ACS or email sending out in general as other things besides usage metrics would use the same secret for outbound email

Issue #5928, opened 2026-07-13
Create Budget Lines


rajohnson90:
Create a vertical slice that involves creation, editing, and deletion of Grant BudgetLines

<img width="622" height="545" alt="Image" src="https://github.com/user-attachments/assets/437abb9c-1d70-49ae-a3d2-003408f0c08f" />

- [ ] Create Grant BudgetLine
- [ ] Delete Grant BudgetLine
- [ ] Edit Grant BudgetLine

rajohnson90:
Merged

PR #5969, opened 2026-07-21
OPS-5928/grant bli numbers

Santi-3rd:
## What changed

Adds **Grant Numbers** as the grant-agreement analog of a Contract's Services Component, and lets budget lines on grant agreements be created, edited, reassigned, and deleted against a Grant Number (mirroring the Contract SC path end-to-end).

**Backend**
- New `GrantNumber` model + `grant_number` table (migration + `*_version` history), plus `grant_number_id` FK on `GrantBudgetLineItem` (migration).
- Grant Number CRUD resource/service (`resources/grant_number.py`, `services/grant_number.py`), routes in `urls.py`, and a new `OpsEventType` for grant-number events.
- Schema support for `grant_number_id` / `grant_number_ref` with mutual-exclusivity validation (a BLI links to at most one of services component / grant number), threaded through the atomic-create and edit-bundle paths.
- `GrantBudgetLineItem.get_required_fields_for_status_change()` now requires `grant_number_id` (not `services_component_id`) so status changes validate against the grant-specific grouping field.
- New read-only role wiring + user data updates; OpenAPI spec regenerated.

**Frontend**
- New `GrantNumbers` component suite: form, list, list item, select, hooks, and constants.
- Budget-line create/edit flow (`CreateBLIsAndSCs`) renders a Grant Number variant for grant agreements, grouped by Grant Number in an accordion (replacing the previous placeholder).
- Editor context + hooks thread `grant_number_number` for not-yet-persisted grant numbers (the `ref` mechanism), and reseed grant numbers on the Agreement Details page for correct grouping after refetch.
- New types (`GrantNumbers.d.ts`) and `AgreementTypes` updates.

**Test fix (this branch)**
- `test_change_requests.py`: the BLI change-request tests target agreement 1, which is a *contract* agreement, so they now create `ContractBudgetLineItem` instead of `GrantBudgetLineItem`. They previously passed only because grant BLIs fell back to the base-class required fields (`services_component_id`); with grant BLIs now requiring `grant_number_id`, the mismatched type failed validation (400 instead of 202).

## Issue

#5928

## How to test

1. Create (or open) a **Grant** agreement.
2. On the budget-lines step, add one or more **Grant Numbers**.
3. Add budget lines and associate each with a Grant Number; confirm lines are grouped by Grant Number in the accordion.
4. Edit a budget line and reassign its Grant Number; delete a budget line.
5. Save via both the new-agreement create path and the edit-bundle path, reload, and confirm `grant_number_id` persists and grouping is correct.
6. Backend: `cd backend/ops_api && pipenv run pytest` (see grant-number, linkage, and change-request tests).
7. Frontend: `cd frontend && bun run test --watch=false`.

## A11y impact

## Storybook

- [x] N/A — new components are page-specific (grant agreement wizard/details), not shared `src/components/UI/`

## Links

- Story: `docs/stories/OPS-5928/grant-budget-lines-story.md`
- Plan: `docs/stories/OPS-5928/grant-budget-lines-plan.md`


weimiao67 (approved):

josbell (changes requested):

A multi-agent adversarial review (53 agents, 19 raw findings → 11 confirmed) found several significant issues that must be addressed before merge. The backend model and migration design are solid, but the frontend save path has a critical data-loss bug and the backend has a missing ownership check and a PATCH data-loss bug.

---

### 🔴 Critical — Must Fix

#### 1. Grant number edits on the Agreement Details page are silently discarded
**File:** `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js`, lines 912–968

`handleSave()` has three paths: new agreement creation, edit-bundle (review flow), and existing-agreement editing (`workflow="none"`). The first two paths correctly include grant number API calls. The third — the `else` branch used on the Agreement Details page — builds SC and BLI mutations but has **zero grant number API calls**. `useAddGrantNumberMutation`, `useUpdateGrantNumberMutation`, and `useDeleteGrantNumberMutation` are not imported in this file at all.

The `GrantNumbers` component renders for `workflow="none"` and dispatches `ADD_GRANT_NUMBER`, `UPDATE_GRANT_NUMBER`, `DELETE_GRANT_NUMBER` to editor context correctly — so users can make changes in the UI — but those changes are never flushed to the API on save. They are silently lost on the next page load.

**Fix:** Import the three GN mutations into `CreateBLIsAndSCs.hooks.js`. In the existing-agreement branch of `handleSave`, read `grant_numbers` and `deleted_grant_numbers_ids` from editor context and fire the API promise chains, mirroring the existing SC mutation structure in that same branch.

---

### 🔴 High — Must Fix

#### 2. Grant number deletions silently dropped in `handleDeletions`
**File:** `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js`, lines 236–255

`handleDeletions()` (called by the existing-agreement save path) processes `deletedServicesComponentsIds` and `deletedBudgetLines` but never processes `deleted_grant_numbers_ids`. The `useDeleteGrantNumberMutation` import is missing. Even after fixing finding #1, this function must also be updated, or deletions will still be dropped.

The edit-bundle path (`getSlice()` in `CreateBLIsAndSCs.jsx`) correctly includes `delete: deletedGrantNumbersIds`, confirming this omission in `handleDeletions` is an oversight.

**Fix:** Add `useDeleteGrantNumberMutation` to imports. Add a deletion loop inside `handleDeletions` for `deleted_grant_numbers_ids`, mirroring the existing `deletedServicesComponentsIds` loop.

#### 3. PATCH of a single field NULLs all other nullable columns
**File:** `backend/ops_api/ops/services/grant_number.py`, lines 132–145

`GrantNumberService.update()` constructs a new `GrantNumber(**updated_fields)` instance from only the fields in the PATCH body, then calls `session.merge()`. Because `GrantNumber` has no Python-level defaults, unspecified fields are `None` on the transient instance. `session.merge()` copies **all** attributes — including those `None` values — onto the persistent row, overwriting whatever the database holds.

A `PATCH /grant-numbers/1 {"description": "updated"}` will NULL out `period_start` and `period_end`. The existing test always sends all fields and does not catch this.

**Fix:** Replace `GrantNumber(**updated_fields)` + `session.merge()` with a fetch-then-setattr pattern: load the existing row by `obj_id`, then call `setattr(gn, key, val)` for each key in `updated_fields`. Add a test that sends a partial PATCH and asserts the omitted fields remain unchanged.

---

### 🟠 Medium — Should Fix

#### 4. IDOR: cross-agreement `grant_number_id` accepted on BLI create without ownership check
**File:** `backend/ops_api/ops/services/budget_line_items.py`, lines 111–144

`BudgetLineItemService.create()` validates that the user is associated with the BLI's `agreement_id` but never checks that `grant_number_id` belongs to that same agreement. A user with write access to Agreement A can POST a BLI with `grant_number_id` pointing to a grant number from Agreement B. The response includes the full nested `grant_number` object, disclosing data the requester has no authorization to read.

The correct check already exists in `_validation()` (lines 791–795) but is only called from `update_with_change_request_ids()`, never from `create()`.

**Fix:** In `create()`, after resolving `grant_number_id`, add: `if gn.agreement_id != agreement_id: raise ValidationError(...)`. Extract the ownership check into a shared helper and call it from both paths.

#### 5. `GrantNumbers.hooks.js` crashes when `grantNumbers` is undefined
**File:** `frontend/src/components/GrantNumbers/GrantNumbers.hooks.js`, lines 77, 112, 124

Line 24 destructures `grant_numbers` from `useEditAgreement() || {}`. If `useEditAgreement()` returns `null`, `grantNumbers` is `undefined`. Lines 77, 112, and 124 call `.map()` / `.findIndex()` on it unconditionally, crashing at runtime. The component is always mounted inside `EditAgreementProvider` today so this is latent, but it will cause a confusing crash in tests or if the component is reused.

**Fix:** Change line 24 to `const { grant_numbers: grantNumbers = [] } = useEditAgreement() || {};`

---

### 🟡 Low — Clean Up

#### 6. `GrantNumberListResponse` is dead code
**File:** `backend/ops_api/ops/schemas/grant_number.py`, line 30

Defined but never imported anywhere. The list endpoint uses `GrantNumberItemResponse(many=True)`. The class implies a `{"items": [...]}` envelope that the API doesn't produce, which misleads readers.
**Fix:** Delete the class.

#### 7. Silent validation pass when `grant_number_id` references a nonexistent grant number on BLI update
**File:** `backend/ops_api/ops/services/budget_line_items.py`, lines 791–795

Guard is `if gn and gn.agreement_id != ...` — when `gn is None`, the check is skipped and the nonexistent FK is written to the DB, causing a 500 from an unhandled `IntegrityError`. The `can_id` check directly below uses the correct pattern (`if can_id and not can: raise ResourceNotFoundError`).
**Fix:** Mirror the `can_id` pattern: if `gn is None`, raise `ResourceNotFoundError`.

#### 8. `SYSTEM_OWNER` role missing `DELETE_GRANT_NUMBER` permission in seed SQL
**File:** `backend/data_tools/initial_data/003-role.sql`

PR adds `GET/PUT/PATCH/POST_GRANT_NUMBER` to `SYSTEM_OWNER` but omits `DELETE_GRANT_NUMBER` — same pre-existing gap as `DELETE_SERVICES_COMPONENT`. All other roles have both delete permissions after this PR.
**Fix:** Confirm if the omission is intentional policy. If not, add both `DELETE_SERVICES_COMPONENT` and `DELETE_GRANT_NUMBER` to `SYSTEM_OWNER` in seed SQL and migration.

#### 9. PR description overstates scope
The description states CreateBLIsAndSCs "renders a Grant Number variant for grant agreements, grouped by Grant Number in an accordion (replacing the previous placeholder)." The code has an explicit placeholder paragraph and inline comment confirming BLI frontend work is deferred to a follow-on PR. The story doc and plan doc reflect this correctly — only the PR description is wrong.
**Fix:** Update the PR description to accurately state that BLI create/edit for grants is deferred to the next PR.

#### 10. `setAlert` exported from `useGrantNumbers` but never used
**File:** `frontend/src/components/GrantNumbers/GrantNumbers.hooks.js`, line 133
`GrantNumbers.jsx` (the only consumer) doesn't destructure it. **Fix:** Remove from the return statement.

---

All 11 findings are confirmed by independent adversarial verification. The backend model design, migration approach, and API structure are well-reasoned — the primary gaps are in the frontend save path (findings 1–2) and two specific backend validation points (findings 3–4).

frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
```diff
@@ -894,7 +903,8 @@ const useCreateBLIsAndSCs = (
                     const createAgreementPayload = {
                         ...cleanData,
                         budget_line_items: cleanBudgetLines,
-                        services_components: newServicesComponents
+                        services_components: newServicesComponents,
+                        grant_numbers: newGrantNumbers
                     };
```
josbell, line 933: **🔴 Critical: Grant number changes silently discarded on Agreement Details page**

The new-agreement creation path (above) correctly includes `grant_numbers: newGrantNumbers` in the payload. However, the `else` branch of `handleSave()` — the path used when editing an **existing** agreement from the Agreement Details page (`workflow="none"`) — has no grant number API calls at all. `useAddGrantNumberMutation`, `useUpdateGrantNumberMutation`, and `useDeleteGrantNumberMutation` are not imported in this file.

The `GrantNumbers` component renders for `workflow="none"` and dispatches `ADD_GRANT_NUMBER`/`UPDATE_GRANT_NUMBER`/`DELETE_GRANT_NUMBER` to the editor context correctly — so users can make changes in the UI — but those changes are never flushed to the API on save. They are silently lost on next page load.

**Fix:** Import the three GN mutations. In the existing-agreement `else` branch, read `grant_numbers` and `deleted_grant_numbers_ids` from editor context and fire the equivalent API promise chains, mirroring the SC mutation structure in that same branch.

frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
```diff
@@ -103,7 +103,8 @@ const useCreateBLIsAndSCs = (
     const {
         agreement,
         services_components: servicesComponents,
-        deleted_services_components_ids: deletedServicesComponentsIds
+        deleted_services_components_ids: deletedServicesComponentsIds,
+        grant_numbers: grantNumbers
     } = useEditAgreement();
```
josbell, line 115: **🔴 High: `handleDeletions` never fires DELETE for removed grant numbers**

`deleted_grant_numbers_ids` is NOT destructured here alongside `deleted_services_components_ids`, and `useDeleteGrantNumberMutation` is not imported in this file. `handleDeletions()` (called by the existing-agreement save path) processes SC and BLI deletions but has no grant number deletion loop.

The edit-bundle path (`getSlice()` in `CreateBLIsAndSCs.jsx`) correctly includes `delete: deletedGrantNumbersIds`, confirming this omission is an oversight.

**Fix:** Also destructure `deleted_grant_numbers_ids` here. Import `useDeleteGrantNumberMutation` and add a deletion loop inside `handleDeletions` mirroring the `deletedServicesComponentsIds` loop.

backend/ops_api/ops/services/grant_number.py
```diff
@@ -0,0 +1,190 @@
...
+        updated_fields["id"] = obj_id  # Ensure ID is included for update
+
+        updated_grant_number = GrantNumber(**updated_fields)
+
+        try:
+            self.db_session.merge(updated_grant_number)
+            if commit:
+                self.db_session.commit()
+            else:
+                self.db_session.flush()
```
josbell, line 138: **🔴 High: `session.merge` on a partial instance NULLs all unspecified nullable columns**

`GrantNumber(**updated_fields)` constructs a transient instance from only the fields in the PATCH body. `GrantNumber` has no Python-level defaults for `description`, `period_start`, `period_end`, and `agreement_id`, so those attributes are `None` on the transient instance. `session.merge()` copies **all** instrumented attributes — including those `None` values — onto the persistent row.

A `PATCH /grant-numbers/1 {"description": "updated"}` will NULL out `period_start` and `period_end`. The existing test always sends all fields and doesn't catch this.

**Fix:** Replace with a fetch-then-setattr pattern:
```python
grant_number = self.db_session.get(GrantNumber, obj_id)
for key, value in updated_fields.items():
    setattr(grant_number, key, value)
```
Add a test that sends a partial PATCH and asserts omitted fields remain unchanged.

backend/ops_api/ops/services/budget_line_items.py
```diff
@@ -787,6 +788,12 @@ def _validation(self, budget_line_item, updated_fields):
         if sc and sc.agreement_id != budget_line_item.agreement_id:
             raise ValidationError({"services_component_id": "Services Component does not belong to the Agreement."})
 
+        grant_number_id = updated_fields.get("grant_number_id")
+        if grant_number_id:
+            gn = self.db_session.get(GrantNumber, grant_number_id)
+            if gn and gn.agreement_id != budget_line_item.agreement_id:
+                raise ValidationError({"grant_number_id": "Grant Number does not belong to the Agreement."})
```
josbell: **🟠 Medium: Two issues with this ownership check**

**1. IDOR — this check is only in `_validation()`, not in `create()`**

`_validation()` is only called from `update_with_change_request_ids()`, never from `create()`. A user with write access to Agreement A can POST a new `GrantBudgetLineItem` with `agreement_id=A` and `grant_number_id` pointing to any grant number from Agreement B. The response includes the full nested `grant_number` object, disclosing data the requester has no access to. Sequential IDs make enumeration straightforward.

**Fix:** Add the same ownership check directly inside `create()` after resolving `grant_number_id`.

---

**2. Low: `if gn and ...` silently passes when grant number doesn't exist**

When `grant_number_id` references a nonexistent row, `gn` is `None`, the `if gn and` guard skips the check, and the nonexistent FK is written to the DB — causing an unhandled `IntegrityError` (500) at commit. The `can_id` check directly below uses the correct pattern: `if can_id and not can: raise ResourceNotFoundError(...)`.

**Fix:** Add an explicit existence check: `if not gn: raise ResourceNotFoundError(...)`

frontend/src/components/GrantNumbers/GrantNumbers.hooks.js
```diff
@@ -0,0 +1,144 @@
...
+        actionButtonText: "",
+        secondaryButtonText: "",
+        handleConfirm: () => {}
+    });
+    const [formKey, setFormKey] = React.useState(Date.now());
+    const { setAlert } = useAlert();
+
+    const dispatch = useEditAgreementDispatch();
+    const { grant_numbers: grantNumbers } = useEditAgreement() || {};
+
```
josbell, line 25: **🟠 Medium: `grantNumbers` will be `undefined` if context returns `null`**

`useEditAgreement()` defaults to `null` (from `createContext(null)`). The `|| {}` fallback has no `grant_numbers` property, so `grantNumbers` is `undefined`. Lines 78, 113, and 125 below call `.findIndex()` and `.map()` on it unconditionally, crashing at runtime.

The component is always mounted inside `EditAgreementProvider` today so this is latent, but it will crash in isolation or unit tests.

**Fix:**
```js
const { grant_numbers: grantNumbers = [] } = useEditAgreement() || {};
```

frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.jsx
```diff
@@ -413,7 +467,11 @@ export const CreateBLIsAndSCs = ({
                 </div>
             )}
 
-            {groupedBudgetLinesByServicesComponent.length > 0 ? (
+            {isGrant ? (
+                // Grant budget lines are #5928 — this slice only shows the empty-state
```
josbell, line 480: **🔴 High: PR description overstates scope — BLI form and accordion are absent**

The PR description states CreateBLIsAndSCs "renders a Grant Number variant for grant agreements, grouped by Grant Number in an accordion (replacing the previous placeholder)." This code renders only a static placeholder paragraph for grants — there is no `GrantNumberAccordion`, no BLI entry form, and no per-grant-number grouping anywhere in the frontend.

The inline comment and story doc (`docs/stories/OPS-5928/grant-budget-lines-story.md`) correctly describe this as intentional deferral to a follow-on PR — only the PR description is wrong.

**Fix:** Update the PR description to accurately state that BLI create/edit for grant agreements is deferred to the next PR.

josbell (commented):
Inline annotations for the findings from my earlier review — each comment is pinned to the relevant line in the diff.

backend/ops_api/ops/schemas/grant_number.py
```diff
@@ -0,0 +1,92 @@
...
+
+class GrantNumberItemResponse(GrantNumberSchema):
+    """Schema for single grant number response."""
+
+    # Extends the base schema to add additional fields for responses if needed
+    # Currently identical to base schema
+
+
+class GrantNumberListResponse(Schema):
+    """Schema for listing multiple grant numbers."""
```
josbell: **🟡 Low: `GrantNumberListResponse` is dead code**

This class is never imported or used anywhere in the codebase. The list endpoint uses `GrantNumberItemResponse(many=True)` directly. The class implies a `{"items": [...]}` response envelope that the API does not produce, which will mislead future readers about the actual response shape.

**Fix:** Delete this class.

frontend/src/components/GrantNumbers/GrantNumbers.hooks.js
```diff
@@ -0,0 +1,144 @@
...
+    return {
+        formData,
+        setFormData,
+        grantNumbers,
+        showModal,
+        setShowModal,
+        modalProps,
+        setModalProps,
+        setAlert,
```
josbell: **🟡 Low: `setAlert` is exported from the hook but no consumer uses it**

`GrantNumbers.jsx` (the only caller) does not destructure `setAlert` from the hook's return value. The hook uses `setAlert` internally — that's fine — but exposing it on the public API surface is unnecessary.

**Fix:** Remove `setAlert` from the return statement.

backend/data_tools/initial_data/003-role.sql
```diff
@@ -1,6 +1,6 @@
...
+INSERT INTO ops.role (id, name, permissions, created_on, updated_on) VALUES (2, 'VIEWER_EDITOR', '{GET_AGREEMENT,PUT_AGREEMENT,PATCH_AGREEMENT,POST_AGREEMENT,GET_BUDGET_LINE_ITEM,PUT_BUDGET_LINE_ITEM,PATCH_BUDGET_LINE_ITEM,POST_BUDGET_LINE_ITEM,DELETE_BUDGET_LINE_ITEM,GET_SERVICES_COMPONENT,GET_GRANT_NUMBER,PUT_SERVICES_COMPONENT,PUT_GRANT_NUMBER,PATCH_SERVICES_COMPONENT,PATCH_GRANT_NUMBER,POST_SERVICES_COMPONENT,POST_GRANT_NUMBER,DELETE_SERVICES_COMPONENT,DELETE_GRANT_NUMBER,GET_BLI_PACKAGE,GET_CAN,GET_DIVISION,GET_NOTIFICATION,PUT_NOTIFICATION,PATCH_NOTIFICATION,GET_PORTFOLIO,GET_RESEARCH_PROJECT,POST_RESEARCH_PROJECT,PATCH_RESEARCH_PROJECT,GET_USER,PUT_USER,PATCH_USER,GET_HISTORY,GET_WORKFLOW,GET_CHANGE_REQUEST,PATCH_CHANGE_REQUEST,POST_CHANGE_REQUEST,GET_CHANGE_REQUEST_REVIEW,GET_UPLOAD_DOCUMENT,POST_UPLOAD_DOCUMENT,PATCH_UPLOAD_DOCUMENT}', current_timestamp, current_timestamp);
```
josbell, line 2: **🟡 Low: `SYSTEM_OWNER` is missing `DELETE_GRANT_NUMBER`**

This PR adds `GET/PUT/PATCH/POST_GRANT_NUMBER` to `SYSTEM_OWNER` but omits `DELETE_GRANT_NUMBER`. All other roles (`VIEWER_EDITOR`, `REVIEWER_APPROVER`, `BUDGET_TEAM`) have `DELETE_GRANT_NUMBER` after this PR. `SYSTEM_OWNER` also lacks `DELETE_SERVICES_COMPONENT` (pre-existing gap).

**Fix:** Confirm whether this is intentional policy. If not, add both `DELETE_SERVICES_COMPONENT` and `DELETE_GRANT_NUMBER` to `SYSTEM_OWNER` in the seed SQL and update the migration accordingly.

Santi-3rd:
@josbell Thanks for pointing those out. I pushed some changes. Here is some info about the changes I didn't make, let me know if you have any more insight about them.

 - SYSTEM_OWNER missing DELETE_GRANT_NUMBER — this is intentional and consistent, not a gap. The migration derives each role's grant-number permissions from its services-component permissions (gn_perm = sc_perm.replace("GRANT_NUMBER", "SERVICES_COMPONENT"), adding the GN perm only when the role already holds the matching SC perm). SYSTEM_OWNER doesn't have DELETE_SERVICES_COMPONENT, so by design it also doesn't get DELETE_GRANT_NUMBER — GN mirrors SC exactly. Adding only DELETE_GRANT_NUMBER would make GN diverge from SC for this role. If the underlying SYSTEM_OWNER delete policy should change, that's a separate decision that should move both delete perms together.

 - PR description overstates scope — correct, doc-only. Grant BLI create/edit and the per-grant-number accordion are deferred to the follow-on PR; the isGrant branch here is an intentional placeholder. Updating the PR description now.

josbell:
### Code review

Found 3 issues:

1. **`grant_number_id` missing from `ALWAYS_DIRECT_EDIT_FIELDS`** — updates to `grant_number_id` on PLANNED/EXECUTING `GrantBudgetLineItem`s are silently dropped. This is the same bug that was fixed for `services_component_id` in commit `132c82c3f`; `grant_number_id` is the direct grant-BLI analog and must be added to the set.

https://github.com/HHS/OPRE-OPS/blob/65b49959cd853637090db3ee82d000886b730d54/backend/ops_api/ops/services/budget_line_items.py#L720-L724

2. **No MessageBus subscribers registered for grant number events** — `CREATE_GRANT_NUMBER`, `UPDATE_GRANT_NUMBER`, and `DELETE_GRANT_NUMBER` are emitted but no `subscribe_globally` calls are registered and `agreement_history_trigger_func` has no `case` branches for them. Grant number create/update/delete will produce no `AgreementHistory` audit trail, unlike the Services Component pattern they mirror (see lines 242–244 below).

https://github.com/HHS/OPRE-OPS/blob/65b49959cd853637090db3ee82d000886b730d54/backend/ops_api/ops/__init__.py#L240-L246

3. **IDOR in `_update_grant_numbers`** — the method calls `self._grant_numbers.update(gn_id, ...)` without verifying that `gn_id` belongs to the bundle's target `agreement_id`. `GrantNumberService._gn_associated_with_agreement` only checks the user has access to the grant number's own parent agreement; it does not verify that parent matches the bundle's target. A user with access to two agreements can submit a bundle edit for Agreement A containing a grant number ID from Agreement B and mutate it.

https://github.com/HHS/OPRE-OPS/blob/65b49959cd853637090db3ee82d000886b730d54/backend/ops_api/ops/services/agreement_edit_bundle.py#L239-L251

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rajohnson90 (approved):

josbell (approved):

Issue #5881, opened 2026-07-01
Procurement Tracker Dashboard: Filter by Division


jonnalley:
### User Story
As a user with access to the Procurement Tracker Dashboard, I want to filter the dashboard by division so that I can focus on procurement activity relevant to a specific team within OPRE.

### Acceptance Criteria
- [ ] The dashboard provides a way to filter data by one or more divisions
- [ ] Divisions are derived from the Project Officer (COR) assigned to each agreement
- [ ] When one or more divisions are selected, only procurement activity for agreements whose Project Officer belongs to one of the selected divisions are shown
- [ ] When no division filter is active, all agreements are shown (default/unfiltered state)
- [ ] The division filter coexists with the Procurement Shop filter so that both can be applied at the same time
- [ ] Applying a division filter updates both the summary cards and the details table to reflect only the filtered agreements
- [ ] The filter state behaves consistently with how the existing Procurement Shop filter behaves (session-only; no URL persistence)
- [ ] The division filter is accessible to users navigating by keyboard or using assistive technology

-----------------------------------------

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] New design/tech debt documented (if applicable)
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

#### Manual Test Cases

**Positive cases:**
- With no division filter active, all agreements on the dashboard are visible and summary cards show aggregate totals
- Selecting a single division shows only agreements whose Project Officer belongs to that division; summary cards update to reflect only those agreements
- Selecting multiple divisions shows agreements whose Project Officer belongs to any of the selected divisions (OR logic)
- Clearing the division filter restores the full unfiltered view
- With both a division filter and a Procurement Shop filter active simultaneously, only agreements matching both criteria are shown
- Navigating away from the dashboard and returning resets the division filter to the unfiltered default (consistent with Procurement Shop behavior)

**Negative / edge cases:**
- Selecting a division that has no agreements with a matching Project Officer results in an empty state in the table (and zero-value summary cards), with no error
- An agreement with no Project Officer assigned does not appear when any division filter is active
- A user without dashboard access cannot reach the filtered view (authorization check)

#### Automated Test Coverage

**Frontend unit/integration tests:**
- Division filter selection updates the displayed agreements and summary data
- Multiple divisions can be selected and the combined result set is correct
[...]

PR #5975, opened 2026-07-22
feat: filter Procurement Dashboard by division

weimiao67:
## What changed

Replaces the single procurement-shop dropdown on the Procurement Dashboard with a multi-select **Filter** panel that supports filtering by **procurement shop** and/or **division**, plus removable filter tags.

**Frontend**
- New `ProcurementDashboardFilterButton` (with `.hooks.js`) providing a filter panel with multi-select `ProcShopComboBox` and `DivisionComboBox` inputs.
- New `ProcurementDashboardFilterTags` (with `.hooks.js`) rendering removable tags for active filters.
- New reusable `DivisionComboBox` and `ProcShopComboBox` components under `src/components/UI/Form/` (each with stories + tests).
- `ProcurementDashboardPage` now holds a `{ procShop, division }` filter state and passes both to `useGetAllAgreements`; removed the old `ProcShopFilter` component.
- Added `useGetDivisionsQuery` usage and a `division` param to the agreements query in `opsAPI.js`.

**Backend**
- Added a `division` list filter to the agreements API. Filters agreements by the division of their project officer (COR) via a subquery through `User`, combining multiple values with OR logic. Agreements without a project officer are excluded when the filter is active.
- Updated `AgreementRequestSchema`, `AgreementFilters`, and `openapi.yml` accordingly.

## Issue

#5881 

## How to test

1. Run the app locally and navigate to the Procurement Dashboard.
2. Click the **Filter** button next to Export.
3. Select one or more procurement shops and/or divisions and apply — verify the summary cards and agreement list update to only include matching agreements.
4. Verify a removable tag appears for each active filter and that removing a tag updates the results.
5. Combine both filters and confirm results reflect procurement shop **and** division constraints.
6. Backend: call the agreements endpoint with `?division=<id>` (and multiple `division` values) and confirm OR-combined results, excluding agreements without a project officer.

## A11y impact

## Storybook

- [x] Story added for new component in `src/components/UI/`

## Screenshots
Screenshot 2026-07-22 at 4 33 48 PM


Santi-3rd (approved):

Santi-3rd:
Some clean-up recommendations from Claude: 

  1. ProcurementDashboardFilterTags.hooks.js:19 — useTagsList keeps fully-derivable state in useState + two useEffects. This is the classic "you might not need an effect" pattern; a useMemo over filters would produce the same list synchronously.
  2. ProcurementDashboardFilterTags.hooks.js:39 — updateTags is memoized on the whole filters object, so changing one filter redundantly rebuilds both filters' tags. Folds away if the first one is applied.
  3. ProcShopComboBox.jsx / DivisionComboBox.jsx — near-identical ~55-line copies differing only in label/namespace/accessor; a single parameterized combobox would remove the duplication.

josbell (changes requested):

A multi-agent adversarial review found **6 confirmed issues**, all introduced by this PR. Two require fixes before merge; four are straightforward cleanups that should be addressed here since this code will be used as a template for future filters.

---

### 🔴 Must Fix

#### 1. E2E division filter test passes vacuously on zero results
**File:** `frontend/cypress/e2e/procurementDashboard.cy.js`, line 165

The test only asserts `expect(filteredCount).to.be.at.most(allCount)`. If the division filter is broken and returns 0 agreements, `0 <= allCount` passes trivially — the test gives a false green. The analogous proc-shop test (lines 131–132) correctly asserts both bounds.

```js
// Add this line:
expect(filteredCount).to.be.at.least(1);
```

Also verify the seed data reliably has at least one active-procurement agreement whose project officer belongs to whichever division appears first in the dropdown. If that's uncertain, select a specific division by name rather than relying on first-option ordering.

#### 2. `updateTags` useCallback depends on the full `filters` object — both effects fire on every single filter change
**File:** `frontend/src/pages/procurementDashboard/ProcurementDashboardFilterTags/ProcurementDashboardFilterTags.hooks.js`, lines 26–48

`useCallback(..., [filters])` depends on the entire filter object reference. Every `setFilters(prev => ({ ...prev, ... }))` call creates a new reference, so `updateTags` is recreated on any filter change. Both `useEffect` hooks depend on `updateTags`, so changing only `procShop` also fires the `division` effect (and vice versa), producing a redundant `setTagsList` call with an unchanged result. In React concurrent mode this can cause a transient intermediate render between the two `setTagsList` updates.

Fix: depend on the specific slice each effect cares about rather than the whole object:

```js
const updateTags = useCallback((filterKey, getTagText) => { ... }, [filters.procShop, filters.division]);
```

Or split into two separate callbacks, each closing over only its own slice.

---

### 🟡 Should Fix (Low — but straightforward and this code sets the pattern for future filters)

#### 3. `resetFilter` replaces entire filter state instead of using functional updater
**File:** `frontend/src/pages/procurementDashboard/ProcurementDashboardFilterButton/ProcurementDashboardFilterButton.hooks.js`, lines 45–50

`setFilters({ procShop: [], division: [] })` silently drops any future filter keys not listed here. `applyFilter` (the sibling function) correctly uses `setFilters(prev => ({ ...prev, ... }))`. Adding a third filter key later would mean Reset silently doesn't clear it.

Fix: use the functional updater form, or derive the empty state from a shared `DEFAULT_FILTERS` constant.

#### 4. `useGetDivisionsQuery({})` creates a separate RTK Query cache entry
**File:** `frontend/src/pages/procurementDashboard/ProcurementDashboardPage.jsx`, line 35

The `getDivisions` endpoint ignores its argument, but RTK Query uses the serialized argument as the cache key — `{}` and `undefined` are different keys. Any future caller that writes `useGetDivisionsQuery()` (no argument) will trigger a second network request for identical data.

Fix: call `useGetDivisionsQuery()` with no argument at all call sites.

#### 5. `Modal.setAppElement('#root')` called in render body
**File:** `frontend/src/pages/procurementDashboard/ProcurementDashboardFilterButton/ProcurementDashboardFilterButton.jsx`, line 65

`Modal.setAppElement` is a side effect called directly in the component function body on every render. `FilterButton` (the inner component) already makes the same call, making this doubly redundant. Calling side effects in render violates React's rules and can produce unexpected behavior under concurrent rendering.

Fix: remove this call entirely from `ProcurementDashboardFilterButton` (covered by `FilterButton`). Ideally, move the single canonical call to `main.jsx`.

#### 6. Lodash `groupBy` + `flat` is a no-op
**File:** `frontend/src/pages/procurementDashboard/ProcurementDashboardFilterTags/ProcurementDashboardFilterTags.jsx`, lines 17–18

```js
const tagsListByFilter = _.groupBy(tagsList, 'filter');
const tagsListByFilterMerged = Object.values(tagsListByFilter).flat();
```

`updateTags` already maintains contiguous grouping by filter key, so `groupBy` + `flat` returns items in the same order as the input — it's a structural no-op. Lodash is imported solely for this transformation.

Fix: remove lines 17–18, remove the `_` import, and pass `tagsList` directly to `<FilterTags>`.

---

All 6 issues were confirmed by independent adversarial verification and are introduced by this PR (commits `e4de620`, `821687f`). No pre-existing issues were included.

weimiao67:
Thanks @Santi-3rd — all three addressed in 3c7aa6b2a:

1. **`useTagsList` derivable state** — replaced the `useState` + two `useEffect`s with a single `useMemo`, so the tags list is now derived synchronously during render (no extra render pass / effect machinery).
2. **`updateTags` over-memoization** — folded away by the rewrite above. The hook now destructures `{ procShop, division }` and depends on those array references directly, so changing one filter no longer rebuilds the other's tags.
3. **`ProcShopComboBox` / `DivisionComboBox` duplication** — collapsed into a single parameterized `FilterComboBox` (`label` / `namespace` / `options` / `selected` / `setSelected` / `optionText`). Wired `ProcurementDashboardFilterButton` to it for both filters, removed the two duplicated wrappers, and moved the tests + stories onto the shared component.

Lint clean and all affected unit tests pass.

weimiao67:
Thanks for the thorough review, @josbell. Went through all 6:

- **#1 (vacuous E2E test)** — Fixed. Added `expect(filteredCount).to.be.at.least(1)` to the division test to match the proc-shop test's bounds.
- **#2 (`updateTags` useCallback deps)** — Already resolved. Commit `3c7aa6b2` ("refactor: simplify filter tags hook and unify filter comboboxes") landed after your review and replaced the effect-mirrored tags state with a pure `useMemo(..., [procShop, division])` in `useTagsList`. There's no longer an `updateTags` callback or the two effects. No change needed.
- **#3 (`resetFilter`)** — Fixed. Now uses the functional updater `setFilters(prev => ({ ...prev, procShop: [], division: [] }))`, consistent with `applyFilter`.
- **#4 (`useGetDivisionsQuery({})` cache key)** — Fixed. Changed to `useGetDivisionsQuery()` (no arg).
- **#5 (`Modal.setAppElement` in render body)** — Left in place with a `TODO` noting the redundant/repeated pattern. Every filter button in the app (Project, Portfolio, Agreements, BLI, CAN) does this identically, so consolidating into a single canonical call in `main.jsx` is better done as one global refactor rather than a one-off here.
- **#6 (lodash `groupBy` + `flat` no-op)** — Fixed. Removed the no-op and the `lodash` import; `tagsList` is passed directly to `<FilterTags>`.

Lint passes and all 90 procurementDashboard unit tests pass.

josbell (approved):
Approving. The change is well-structured, follows the existing filter-modal/tags patterns, and is thoroughly tested (unit tests for both new components + hooks, a backend service test class covering the no-PO exclusion and division+awarding-entity intersection, and updated Cypress specs).

Verified clean:
- Backend subquery filters by the project officer's division via `project_officer_id IN (SELECT User.id WHERE User.division IN (...))` — imports present, `User.division` is a nullable integer FK, and the WHERE adds no joins so it doesn't interact badly with the base query's `.distinct()`/outer joins.
- NULL project-officer exclusion is intentional and documented in the code comment + openapi.yml, with a dedicated test.
- Frontend buffer/effect design doesn't clobber in-progress edits, and Reset works via the re-seed effect. No dead code left behind.

Two non-blocking notes:
- `Modal.setAppElement("#root")` duplication is acknowledged via TODO and correctly deferred to a global refactor.
- Division semantics (filter by the project officer's division, excluding agreements with no PO) is documented in openapi.yml — worth confirming with the ticket owner that "no PO ⇒ excluded" matches intent.

Santi-3rd (approved):

PR #5977, opened 2026-07-22
OPS-5928: Grant budget lines — frontend (PR 2)

weimiao67:
## What changed

**PR 2 of 2** for grant budget lines (OPS-5928) — all user-facing frontend. Stacked on the backend data-layer PR (#5969); base branch is `OPS-5928/grant-bli-numbers`.

Budget lines can now be created, edited, deleted, and grouped on **grant** agreements, working exactly like contract budget lines except the "Services Component" concept is replaced by "Grant Number."

**Write UI**
- New `AllGrantNumberSelect` (reads grant numbers from the AgreementEditor context) and `GrantNumberAccordion`.
- `BudgetLinesForm` gains an `isGrant` variant; the Vest suite registers only the applicable select group (bidirectional skip) so the inactive select never disables the Add button.
- Editor hook threads `grant_number_number` through add/edit/edit-by-id/hydration/baseline-decoration/duplicate; adds `groupByGrantNumber` grouping.
- Grant link wired into all three save paths (new-agreement `grant_number_ref`, existing-agreement id resolution, edit-bundle ref/id).

**Read/approval parity** (grants are not gated out of approval)
- Group + validate grant BLIs by Grant Number across Agreement Details, Review, Approve (before/after), and Pre-award.
- Review suite is type-aware (requires `grant_number_id` for grants); all-BLI table + XLSX export show the grant number.

**Grant display rules (Step 3 – Grants design)**
- Grant budget-line tables omit the Fee and Total columns; `AgreementTotalCard` shows only the total (grants have no procurement shop).
- `ProjectAgreementSummaryCard` hides Procurement Shop and Current Fee Rate for grants.
- Breadcrumb restored on the create page; wizard step 3 renamed to "Budget Lines"; Create Grant Numbers instruction text updated; grant-number hint placed above the field with consistent spacing.

## Issue

[OPS-5928](https://jira.acf.gov/browse/OPS-5928) — Grant Numbers & Budget Lines (Step 3 of Create Grant Agreement)

## How to test

Requires the backend PR (#5969) merged/running and the full stack up.

1. Create a Grant agreement → add Grant Numbers → add budget lines, each associated with a Grant Number.
2. Verify budget lines group under the correct Grant Number accordion; grant tables show no Fee/Total columns; the Agreement Total card shows only the total; the summary card omits Procurement Shop / Fee Rate.
3. Edit a BLI's grant number + amount, delete a BLI, and save. Confirm `grant_number_id` persists after reload on all three save paths (new-agreement wizard, Agreement-Details edit, review edit-bundle).
4. Submit for approval; confirm Review/Approve/Pre-award pages group grant BLIs by grant number with no false "services component" validation errors.

E2E: `frontend/cypress/e2e/createGrantAgreement.cy.js` covers the create flow, grouping, `grant_number_ref` payload, breadcrumb, step-3 label, and instruction text.

## A11y impact

The Cypress spec runs `checkA11y` (axe) after each test and passes; changes reuse existing accessible USWDS components (Select hint uses `usa-hint`, accordions, tables).

## Storybook

New components (`AllGrantNumberSelect`, `GrantNumberAccordion`) are feature components under `GrantNumbers/`, not shared `src/components/UI/` components.

## Links

- Backend PR (base): #5969
- Plan: `docs/stories/OPS-5928/grant-budget-lines-plan.md`


josbell (changes requested):
## Code Review — Changes Requested

Reviewed against the full diff for branch `OPS-5928/grant-bli-numbers-frontend`. Ran 8 independent finder angles, verified each candidate independently. **7 issues confirmed**, 1 refuted after deep investigation.

---

### 🔴 Critical — Fix before merge

#### 1. `session.merge()` with partial payload silently nulls existing columns
**File:** `backend/ops_api/ops/services/grant_number.py:132`

`update()` constructs `GrantNumber(**updated_fields)` from only the patched fields, then calls `session.merge()`. SQLAlchemy merge copies ALL attributes from the transient object to the persistent row — including `None` for every column not in the payload. Any `PATCH /grant-numbers/{id}` with a partial body (e.g. just `{"number": 3}`) will silently null out `description`, `period_start`, and `period_end`.

If `agreement_id` is absent from the payload, the flush also hits a NOT NULL IntegrityError — but with the misleading message "A Grant Number with this number already exists for this agreement."

**Fix:** Replace the merge pattern with a fetch-then-setattr pattern (already used by most other services):
```python
grant_number = self.db_session.get(GrantNumber, obj_id)
for key, value in updated_fields.items():
    setattr(grant_number, key, value)
```

---

#### 2. Non-existent `grant_number_id` on BLI update silently passes validation → FK 500
**File:** `backend/ops_api/ops/services/budget_line_items.py:792`

```python
if grant_number_id:
    gn = self.db_session.get(GrantNumber, grant_number_id)
    if gn and gn.agreement_id != budget_line_item.agreement_id:
        raise ValidationError(...)
    # ← no "not found" guard when gn is None
```

When `grant_number_id` points to a deleted/non-existent row, `gn` is `None`, the condition is `False`, and the request passes validation — then hits a DB FK `IntegrityError` at flush, returning a raw 500. The `can_id` check 8 lines below has the correct pattern:

```python
if can_id and not can:
    raise ResourceNotFoundError("CAN", can_id)
```

**Fix:** Mirror that guard:
```python
if grant_number_id:
    gn = self.db_session.get(GrantNumber, grant_number_id)
    if not gn:
        raise ResourceNotFoundError("GrantNumber", grant_number_id)
    if gn.agreement_id != budget_line_item.agreement_id:
        raise ValidationError({"grant_number_id": "Grant Number does not belong to the Agreement."})
```

---

#### 3. `AgreementBudgetLines.jsx` is missing `grantNumbersReseedKey` — stale state after failed save
**File:** `frontend/src/pages/agreements/details/AgreementBudgetLines.jsx:261`

`EditAgreementProvider` is mounted without `grantNumbersReseedKey`, so the `RESEED_GRANT_NUMBERS` dispatch in `AgreementEditorContext.jsx` can never fire on this page. After a save failure the context's `grant_numbers` stays at the in-progress state; retrying re-submits the stale/failed edits.

`EditAgreementAndBudgetLines.jsx` already has the correct pattern:
```javascript
const [grantNumbersReseedKey, setGrantNumbersReseedKey] = useState(0);
// on catch: setGrantNumbersReseedKey((k) => k + 1)
// on provider: grantNumbersReseedKey={grantNumbersReseedKey}
```

**Fix:** Add the same three pieces to `AgreementBudgetLines.jsx`.

---

### 🟡 Should fix

#### 4. `put()` audit events always record `null` actor
**File:** `backend/ops_api/ops/resources/grant_number.py:48`

`put()` reads `grant_number.updated_by` from the sparse object returned by `service.update()` — which is `GrantNumber(**updated_fields)` and never contains `updated_by` (it's `dump_only` in the schema). Every `PUT` audit event records `updated_by=None`.

`patch()` already avoids this by re-fetching via `service.get(id)` after the merge. Apply the same re-fetch to `put()`.

---

#### 5. `servicesComponentNumber.toString()` without null guard
**File:** `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js:661`

```javascript
serviceComponentGroupingLabel: servicesComponentNumber.toString()
```

The initial `null` state is protected by the `budgetLineBeingEdited` guard, but if the user clears the services component selector mid-edit, `servicesComponentNumber` can return to `null` while `handleEditBLI` remains reachable — throwing `TypeError: Cannot read properties of null (reading 'toString')`.

**Fix:**
```javascript
serviceComponentGroupingLabel: (servicesComponentNumber ?? 0).toString()
```

---

#### 6. `handleDelete` and `setFormDataById` use `findIndex` result without `-1` guard
**File:** `frontend/src/components/GrantNumbers/GrantNumbers.hooks.js:78,113`

Both functions call `findIndex` then immediately index into `grantNumbers` without checking for `-1`. A double-click, rapid state update, or concurrent dispatch could call either with a value no longer in the list, causing `grantNumbers[-1]` → `undefined` → crash reading `.display_title` / `.period_start`.

**Fix:** Add an early return after each `findIndex` call:
```javascript
const index = grantNumbers.findIndex((item) => item.number === number);
if (index === -1) return;
```

---

#### 7. `addGrantNumberIdToBLI` silently returns `{grant_number_id: null}` on no match
**File:** `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js:783`

```javascript
grant_number_id: matchGrantNumber?.id ?? null,
```

If the grant number lookup fails (stale page data, partial creation failure), the BLI is silently saved without a grant number — no error surfaces to the user. This is the same pattern as the SC analog, so it's not unique to this PR, but it's new code introducing the same risk.

**Fix:** Throw an error if no match is found rather than returning `null` silently, or at minimum surface this case as a validation error before the save is attempted.

---

*Reviewed by Claude Code at high effort — 8 finder angles, 11 independent verifiers.*

Santi-3rd (approved):

rajohnson90:
This looks good to me aside from the fixes identified by Josbell. I'll give my approval as well once they're done.

rajohnson90 (approved):
Looks good to me. I verified that all the issues identified in Josbell's review are fixed or I fixed them myself this morning.

rajohnson90:
Dealing with test failures after the merge from main.

josbell (approved):

PR #6003, opened 2026-07-28
feat: procurement tracker step notes editor feedback

weimiao67:
## What changed

- Unified the notes editor across all six procurement tracker steps (`StepOne` through `StepSix`) by consolidating into a shared `StepNotesEditor` component, replacing the per-step `StepNotesForm` component (which has been removed).
- Fixed the Save Notes check icon to be greyed out when disabled (no unsaved changes or save in progress).
- Sized the notes TextArea and the read-only note text to a shared **664px** width (`STEP_NOTES_WIDTH`), overriding the USWDS `.usa-textarea` 30rem cap so the persisted note wraps at the same right edge as the editor.
- Rendered the read-only note text at the standard page text size (kept `font-12px` only on the "Notes" label) and constrained its width so long notes wrap.
- Moved the Cancel/Save Notes button row up onto the character-count line via the `margin-top-neg-205` utility class.
- Fixed invalid HTML: the Edit Notes button now lives inside the `<dd>` (below the note text) instead of being a bare `<button>` child of `<dl>`.
- Allowed clearing a previously saved note — Save is disabled only for a first-time empty entry; once a note exists the user can save an empty value, and read mode then shows "None".
- Read mode now trusts the live (trimmed) `notes` value, so a cleared note shows "None" immediately instead of briefly flashing the stale server value before the refetch lands (removed the redundant `notesLabel` prop).

### Review fixes

- **Step-level Cancel now collapses the notes editor.** `useSaveNotes` exposes a `notesResetKey` (bumped on every `resetNotes()`), threaded to `StepNotesEditor` as a `resetSignal` prop so the editor returns to read mode in lockstep with the value reset instead of leaving the textarea open.
- **Async-loaded notes are no longer wiped on Cancel.** When a step mounts before its saved note has fetched, the editor opens implicitly; `committedNotesRef` now syncs from the incoming `savedNotes` so Cancel discards back to the loaded value rather than clearing the note or keeping unsaved keystrokes.
- **Step-level Cancel no longer flashes "None" after a first-time save.** `useSaveNotes` tracks the last committed value (initial server value, then the just-saved value on each successful save); `resetNotes()` with no argument restores it, avoiding the stale-prop window before the RTK Query refetch lands.
- **Backend caps `notes` length.** The PATCH schema's `notes` field now uses `validate.Length(max=750)`, mirroring the frontend `STEP_NOTES_MAX_LENGTH` so a client bypassing the UI can't persist an arbitrarily long note.

## Issue

#5435 

## How to test

1. Navigate to a Procurement Tracker agreement.
2. Open any step (One through Six) and expand the notes section.
3. Verify the notes editor renders and functions correctly for all six steps.
4. Make a change and confirm the Save button/check icon is enabled; without a change, confirm it is greyed out/disabled.
5. Confirm the TextArea and the read-only note text share the same width (664px) and that long notes wrap at that edge.
6. Add a note and save; confirm it flips to read mode showing the note.
7. Edit an existing note, clear all text, and save; confirm read mode shows "None" (no flash of the old note) and the Save button was enabled for the clear.
8. On a first-time note (no saved value), confirm Save stays disabled until you type non-whitespace text.
9. With a saved note in the step form, open the notes editor, then click the step-level **Cancel** (confirm the modal); confirm the notes editor collapses back to read mode along with the other fields.
10. On a fresh step, add a first note and Save, then immediately click the step-level Cancel; confirm the just-saved note is preserved (no flash of "None").
11. Attempt to PATCH a step with a `notes` value longer than 750 characters (e.g. via API); confirm the request is rejected by server-side validation.

## A11y impact

## Storybook


josbell (changes requested):
## Two bugs require changes before merge

### 🐛 Bug 1 — Data overwrite race in `useSaveNotes.js:68` (correctness)

`isDirtyRef.current = false` runs unconditionally after the `await patchStep(...).unwrap()` succeeds. If the user types while the PATCH is in-flight, `setNotes` sets `isDirtyRef.current = true`. The success path then clears it, and the subsequent RTK Query refetch overwrites whatever the user typed.

**Reproduction** (Network → Slow 3G):
1. Type a note → Save
2. Immediately type more text while the request is pending
3. Wait for the save + refetch to complete → new text disappears

**Fix** — snapshot `notes.trim()` before the `await` and only clear the dirty flag if no new input arrived:

```js
const handleSaveNotes = async (stepId) => {
    const notesToSave = notes.trim();
    try {
        await patchStep({ stepId, data: { notes: notesToSave } }).unwrap();
        // Only clear dirty if the user hasn't typed anything new during the round-trip.
        if (notes.trim() === notesToSave) {
            isDirtyRef.current = false;
        }
        return true;
    } catch (error) { ... }
};
```

---

### 🐛 Bug 2 — Keyboard focus drops to `<body>` on every Save/Cancel (`StepNotesEditor.jsx`)

Edit mode returns `<div className="display-table">`, read mode returns `<dl>`. Different root element types force a full DOM unmount/remount on every mode switch. The focused Save Notes or Cancel button is destroyed, and the browser resets focus to `<body>`.

This is a direct WCAG 2.4.3/2.4.7 issue on a code path the PR explicitly validates as "Accessibility changes included and validated against WCAG 2.1 AA intent."

**Reproduction** (keyboard only):
1. Tab to **Save Notes** → press Enter
2. `document.activeElement` → `<body>` instead of the Edit Notes button

**Fix** — wrap both branches in a single stable root so React reconciles in place, then restore focus:

```jsx
const editNotesRef = React.useRef(null);

// After save or cancel, call: editNotesRef.current?.focus()

return (
    <div>  {/* stable root — React reconciles instead of unmounting */}
        {isEditingNotes ? (
            <div className="display-table">...</div>
        ) : (
            <dl>
                ...
                <button ref={editNotesRef} onClick={() => setIsEditingNotes(true)}>
                    Edit Notes
                </button>
            </dl>
        )}
    </div>
);
```

---

### ⚠️ Also worth addressing (non-blocking)

- **Accessibility gap**: the deleted `setAlert({ type: "success" })` was the only programmatic save-completion signal for screen-reader users (ARIA live region). The visual edit→read transition is not perceivable by AT. At minimum, an `aria-live` region announcement should replace it — especially given the a11y checkbox was checked.
- **`isReadOnly` branches in Steps 1–5** still have hand-rolled `<dt>/<dd>` Notes blocks that weren't migrated to `StepNotesEditor`. Step 3 already diverges (missing `font-12px`). These don't have to be in this PR but the drift is already visible.
- **`handleSaveNotes` not in `useCallback`** — unlike `setNotes`/`resetNotes`, this produces a new reference on every render. Worth making consistent.

rajohnson90 (approved):
I didn't find anything aside from Josbell's suggestion so I'll approve.

weimiao67:
Thanks for the thorough review, @josbell — all addressed in 2f04bbf.

### 🐛 Bug 1 — save race (`useSaveNotes.js`) ✅
Fixed, with one adjustment to the suggested patch. The `notes.trim() === notesToSave` comparison as written doesn't actually work: `handleSaveNotes` closes over `notes` from the render when it was created, so it never sees the in-flight keystrokes and the flag clears regardless. I track the live value in a `notesRef` and compare against `notesRef.current.trim()` after the await, so mid-flight edits genuinely survive the invalidation refetch. Added a unit test that types during a pending PATCH and asserts the keystrokes aren't clobbered.

### 🐛 Bug 2 — focus drops to `<body>` ✅
Fixed. Note the root cause is a bit different from the div-vs-`<dl>` framing: the focused Save/Cancel button is *conditionally unmounted* on the mode switch, so focus would drop to `<body>` even with a stable root element. The actual remedy is explicit focus restoration — an `editNotesRef` + effect that moves focus to **Edit Notes** after Save/Cancel (guarded so the initial read-mode mount doesn't steal focus). Added tests for save, cancel, and the initial-mount case.

### ⚠️ aria-live save signal ✅
Good catch — the deleted success alert rendered with `role="status"`, so it was a real AT signal. Added a persistent visually-hidden `role="status" aria-live="polite"` region that announces "Notes saved." on success (cleared on re-entering edit mode so repeat saves re-announce). Tests cover success, re-edit, and the failure path (no false announcement).

### ⚠️ `handleSaveNotes` not memoized ✅
Wrapped in `useCallback`. Composes with the Bug 1 fix since it now reads the live value from the ref rather than a stale closure.

### ⚠️ `isReadOnly` Notes blocks / Step 3 `font-12px`
Dug into this one and it turned out to be intentional rather than drift: Step 3's `isReadOnly` block *does* have `font-12px` (same as Steps 1/2/4/5), and there's an existing test (`ProcurementTrackerStepThree.test.jsx:1142`) that explicitly asserts `StepNotesEditor`'s read-mode `<dt>` must **not** have `font-12px`. The two are deliberately different label sizes for different contexts (whole-step read-only view vs. the completed/saved note editor), so I've left the styling as-is. Happy to revisit if you'd intended them to converge.

Full ProcurementTracker suite is green (813 tests). Thanks again!

josbell (changes requested):
## Review findings

Two confirmed bugs share the same root cause: `hasSavedNote` is derived from the `savedNotes` server prop rather than local state, so during the RTK Query refetch window after a first-time save, it is stale.

### Bug 1 — Cancel button vanishes after first-time save (a11y blocker)

**File:** `StepNotesEditor.jsx:134`

After saving a note for the first time on an active step, `savedNotes` prop is still `""` until the RTK Query refetch resolves. If the user clicks Edit Notes before that refetch lands, `canReturnToReadMode = false` → the Cancel button is not rendered. Keyboard users are trapped in edit mode with no escape — a WCAG 2.4.3 violation on a PR that explicitly claims WCAG 2.1 AA compliance.

### Bug 2 — Save disabled for clearing a just-saved note (same window)

**File:** `StepNotesEditor.jsx:158`

Same stale-`savedNotes` window. User re-enters edit mode and clears the textarea. `isDisabled={isDisabled || (!hasSavedNote && !notes.trim())}` evaluates to `true` because `hasSavedNote` is still `false`. The user cannot clear-and-save the note they just persisted until the refetch arrives.

### Root cause & fix

Both bugs share the same fix: seed a local `hasEverSaved` state from the prop and set it `true` on a successful save, rather than re-deriving from the stale prop on every render.

```jsx
// Initialize from prop, then promote on first successful save
const [hasEverSaved, setHasEverSaved] = React.useState(Boolean((savedNotes ?? "").trim()));

// In the Save onClick:
if (didSave) {
    setHasEverSaved(true);  // <-- add this
    setSaveAnnouncement("Notes saved.");
    returnToReadMode();
}

// Replace hasSavedNote with hasEverSaved on lines 65 and 158
```

### Low-severity note (not a blocker)

**File:** `ProcurementTrackerStepOne.hooks.js:24` (and equivalents 2–6)

`handleSaveNotes` and `handleStep1Complete` both fire the same `patchStepOne` mutation trigger with no `isLoading` guard. A user who clicks Save Notes then immediately clicks Complete Step can produce two concurrent PATCHes. Status and metadata are not at risk (the notes-only payload has no `status` field), but the `notes` field can transiently revert to a stale value if the notes-only PATCH arrives after the Complete PATCH. Low impact, but worth a mutual-exclusion guard (e.g. disable Save Notes when the Complete mutation is in-flight).

weimiao67:
Thanks @josbell — all three findings addressed in 1ff965fc4.

**Bugs 1 & 2 (stale `savedNotes` window).** Replaced the prop-derived `hasSavedNote` with local `hasEverSaved` state in `StepNotesEditor`: seeded from the prop, promoted to `true` the instant a save succeeds (and via an effect if a non-empty server value arrives later, e.g. async load / another user's edit). `canReturnToReadMode` and the Save-disabled condition now read `hasEverSaved`, so a first-time save keeps Cancel available (WCAG 2.4.3) and Save enabled for clearing the just-saved note without waiting for the refetch. Added two regression tests that reproduce the refetch window (`savedNotes` still `""` after the first save).

**Low-severity Save/Complete race.** Implemented the mutual-exclusion guard across all six steps. Each step hook already holds one `useUpdateProcurementTrackerStepMutation` instance shared by `handleSaveNotes` and the complete handler, so its `isLoading` (exposed as `isStepPatchInFlight`) is true for either in-flight PATCH. Threaded it into every Complete button's `disabled` and every `StepNotesEditor` `isDisabled`, so Save Notes and Complete can't fire concurrently — the notes-only PATCH can no longer land after Complete and revert `notes`. Also fixed the six `.hooks.test.js` mocks that returned `[trigger]` (the real RTK Query hook returns `[trigger, meta]`) and added a hook test asserting an in-flight PATCH disables the buttons.

One scoping note: I limited the guard to Save Notes ↔ Complete (the scenario you described). The per-step date-submit buttons also share the mutation, but their payloads carry no `notes` field so they can't revert notes — happy to guard those too if you'd prefer.

All 816 ProcurementTracker unit tests pass; lint and format are clean.

frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx
```diff
@@ -33,72 +47,180 @@ const StepNotesEditor = ({
...
+            setHasEverSaved(true);
+        }
+    }, [savedNotes]);
+
+    // The editor can return to read mode when a note has ever been saved or when
+    // the caller renders read mode by default (completed steps). A fresh active
+    // step with no saved note has no read view to cancel back to, so it stays in
+    // input mode and omits Cancel.
+    const canReturnToReadMode = hasEverSaved || startInReadMode;
+    const [isEditingNotes, setIsEditingNotes] = React.useState(!canReturnToReadMode);
```
josbell, line 81: **Bug (CONFIRMED): `isEditingNotes` is permanently stuck in edit mode when `savedNotes` arrives asynchronously after mount.**

`useState(!canReturnToReadMode)` captures the initial value only once. When `savedNotes` is `undefined` at mount (step data still loading), `hasEverSaved=false` → `canReturnToReadMode=false` → `isEditingNotes=true`.

When the fetch resolves with a non-empty note, the `useEffect` on line 68 promotes `hasEverSaved=true` — but nothing ever calls `setIsEditingNotes(false)`. The component is permanently stuck in edit mode for that mount.

All six step components mount `StepNotesEditor` unconditionally with `savedNotes={stepXData?.notes}` before the query resolves (e.g. `ProcurementTrackerStepOne.jsx` line 172), so every active step with an existing note is affected.

**Fix:** Add a `useEffect` that resets `isEditingNotes` to `false` the first time `canReturnToReadMode` transitions to `true` (i.e. when async data arrives), or guard the initial mode calculation on whether data has loaded.

frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx
```diff
@@ -33,72 +47,180 @@ const StepNotesEditor = ({
...
+
+    /** Enter edit mode, clearing any prior save announcement. */
+    const enterEditMode = () => {
+        setSaveAnnouncement("");
+        setIsEditingNotes(true);
+    };
+
+    // A persistent visually-hidden live region so the save announcement is spoken
+    // even as the editor swaps between the edit (<div>) and read (<dl>) subtrees.
+    const liveRegion = (
```
josbell, line 168: **Bug (CONFIRMED): `aria-live` region is not actually persistent — the save announcement is never spoken by screen readers.**

The comment at line 114 says the region is "rendered persistently below so its text change is announced", but this is not what happens. `liveRegion` is a local JSX variable included in *both* `return` branches (edit-mode line 130, read-mode here at line 198). When a save succeeds, `setSaveAnnouncement("Notes saved.")` and `setIsEditingNotes(false)` are batched, causing React to swap the entire subtree.

The old `<span>` (with `saveAnnouncement=""`) is **unmounted**. A fresh `<span>` (with `"Notes saved."` already present at insertion) is **mounted**. Screen readers (NVDA, JAWS, VoiceOver) fire `aria-live` announcements on *text mutations to existing DOM nodes* — content present when a node is first inserted is not announced. The save is completely silent to assistive technology.

**Fix:** Move the live region outside the conditional branches so a single stable DOM node receives the text change — restructure to a single `return` with conditional inner content, or hoist the `<span>` to a stable ancestor.

frontend/src/components/Agreements/ProcurementTracker/useSaveNotes.js
```diff
@@ -57,31 +62,43 @@ export default function useSaveNotes(patchStep, serverNotes, setAlert) {
...
+                await patchStep({
+                    stepId,
+                    data: { notes: notesToSave }
+                }).unwrap();
+                // The save succeeded, so the server value now matches the field —
+                // but only if nothing new was typed during the round-trip. Read the
+                // live value from the ref (not a closed-over `notes`) and clear the
+                // dirty flag only when the field still holds exactly what we saved,
+                // so future server updates can sync in again without losing edits.
+                if (notesRef.current.trim() === notesToSave) {
```
josbell, line 114: **Bug (CONFIRMED): Dirty-flag is incorrectly cleared when whitespace is typed during an in-flight empty-string PATCH.**

The guard `notesRef.current.trim() === notesToSave` is trivially true when `notesToSave = ""` (cleared note) regardless of whitespace typed during the round-trip (`"   ".trim() === ""`). Steps:

1. User clears note, clicks Save: `notesToSave = ""`
2. `TextArea` is **not** disabled during saves — `isStepPatchInFlight` does not flow to the `TextArea`'s `isDisabled` prop (see `StepNotesEditor.jsx` line 140: `isDisabled={isDisabled}`, not `isDisabled={isDisabled || isStepPatchInFlight}`)
3. User types `"   "` while PATCH is in-flight → `isDirtyRef=true`, `notesRef.current="   "`
4. PATCH resolves: `"   ".trim() === "" === notesToSave` → `isDirtyRef.current = false`
5. Refetch delivers `serverNotes=""` → `!isDirty` → `setNotesState("")`, silently overwriting user input

**Fix:** Either thread `isStepPatchInFlight` through to the `TextArea`'s `isDisabled` in `StepNotesEditor`, or tighten the guard to raw equality: `notesRef.current === notesToSave`.

josbell (commented):
Code review by Claude Code — 4 confirmed bugs + 3 test gaps found (all verified against the checked-out code). Inline comments below for issues on changed lines; issues on unchanged lines are summarized at the end of this comment.

---

### Finding 4 (CONFIRMED, unchanged line) — `ProcurementTrackerStepSix.jsx` line 119
**Notes row silently disappears in read-only view when no note has been saved.**

`{stepSixNotesLabel && <div>...` uses short-circuit rendering with no fallback. When `stepSixNotesLabel` is falsy (no saved note), the entire notes row is omitted — a read-only reviewer cannot distinguish "no notes" from "not loaded". All sibling fields in this `isReadOnly` block use `|| "Unknown"` fallbacks (e.g. completed-by on line 117). The non-`isReadOnly` completed branch renders via `StepNotesEditor` which correctly shows `"None"`.

**Fix:** `<strong>Notes:</strong> {stepSixNotesLabel || "None"}`

---

### Finding 5 (CONFIRMED, latent) — `ProcurementTrackerStepThree/ProcurementTrackerStepThree.hooks.test.js` line 28
### Finding 6 (PLAUSIBLE, latent) — `ProcurementTrackerStepFour/ProcurementTrackerStepFour.hooks.test.js` line 28
**Suite mock missing `.run` — any test exercising `runValidate()` will throw.**

Both mocks define `mockSuite.get` and `mockSuite.reset` but omit `mockSuite.run = vi.fn()`. The hooks files call `suite.run({...}, name)` (Vest v6 API — line 51 in StepThree, line 53 in StepFour). No existing test calls `runValidate()` so nothing throws today, but the mock is structurally defective. Compare with `ProcurementTrackerStepOne.hooks.test.js` which correctly includes all three methods.

**Fix:** Add `mockSuite.run = vi.fn();` to both mock blocks.

---

### Finding 7 (PLAUSIBLE) — `ProcurementTrackerStepOne/ProcurementTrackerStepOne.test.jsx` and all other step test fixtures
**`isStepPatchInFlight` absent from `defaultHookReturn` — the mutual-exclusion guard this PR adds is never tested at component level.**

When `isStepPatchInFlight` is absent from the mock, `isDisabled || undefined` evaluates to `isDisabled` (undefined is falsy), so the in-flight guard is always inactive in tests. Neither StepOne nor StepTwo test files contain `isStepPatchInFlight` anywhere — not in fixtures, not in per-test overrides.

**Fix:** Add `isStepPatchInFlight: false` to each `defaultHookReturn`, and add at least one test per step that sets `isStepPatchInFlight: true` and asserts both the Complete button and `StepNotesEditor` are disabled.

frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx
weimiao67: Looked into this — it doesn't reproduce against the current code, on two counts:

1. The parent (`AgreementProcurementTracker.jsx:94`) gates all step rendering on `isLoading`, so `StepNotesEditor` never mounts until the query resolves — `savedNotes` is already populated at mount, not `undefined`.
2. Even in the hypothetical where a non-empty note arrives async while the user is mid-edit, auto-forcing read mode would silently discard their in-progress typing. The `useEffect` deliberately promotes `hasEverSaved` (so Cancel/read affordances appear) *without* yanking the user out of edit mode. That's the intended behavior.

Leaving as-is.

frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx
weimiao67: This one is already handled by the structure. `liveRegion` is the **first** child of the fragment in both the edit-mode return (`{liveRegion}` then `<div>`) and the read-mode return (`{liveRegion}` then `<dl>`). React reconciles children by position + type, so across a save the `<span>` at index 0 is the same type in both trees and is updated in place — only the sibling at index 1 (`<div>`↔`<dl>`) unmounts/remounts. The `saveAnnouncement` text therefore mutates on a stable DOM node, which is exactly what NVDA/JAWS/VoiceOver announce.

There's an existing test asserting this ("announces the save to assistive tech via an aria-live region after a successful save" in `StepNotesEditor.test.jsx`). Leaving as-is.

frontend/src/components/Agreements/ProcurementTracker/useSaveNotes.js
weimiao67: The repro's step 2 doesn't hold against current code: the `TextArea`'s `isDisabled` does receive the in-flight state. Every step passes `isDisabled={isDisabled || isStepPatchInFlight}` to `StepNotesEditor` (e.g. `ProcurementTrackerStepOne.jsx:175`), and the editor forwards that same `isDisabled` to the `TextArea` (`StepNotesEditor.jsx:140`). So while a notes PATCH is in flight the textarea is disabled and the user can't type the `"   "` that the scenario depends on.

The guard also already compares the live ref value (`notesRef.current.trim() === notesToSave`) rather than raw equality — but since the input can't be edited mid-flight, the whitespace-clobber path isn't reachable. Leaving as-is.

(I did add component-level tests this round asserting Save Notes + the textarea are disabled while `isStepPatchInFlight` is true — see Finding 7.)

weimiao67:
Thanks @josbell — addressed in 82d5782.

## Findings 4–7 (fixed)

**Finding 4 — StepSix read-only Notes row silently omitted.** ✅
Changed the `isReadOnly` block to always render `<strong>Notes:</strong> {stepSixNotesLabel || "None"}`, matching the `|| "Unknown"` sibling fields and the `|| "None"` fallback in steps 1–5. Added a read-only test asserting the Notes row shows "None" when no note was saved.

**Findings 5 & 6 — Suite mocks missing `.run`.** ✅
Added `mockSuite.run = vi.fn()` to the StepThree and StepFour `.hooks.test.js` suite mocks so they match the other four steps and won't throw if a test exercises `runValidate()`.

**Finding 7 — `isStepPatchInFlight` untested at component level.** ✅
Added `isStepPatchInFlight: false` to all six step component fixtures, plus a per-step test that sets it `true` and asserts the mutual-exclusion guard: Save Notes (and the textarea) disable in every step, and the Complete button disables in Steps 2 and 6 (the two steps that gate Complete on `isStepPatchInFlight` at the component level; Steps 1/3/4/5 fold it into the hook-computed `disableStepNButtons`).

## The three inline findings (no change — replied inline)

After verifying against the current branch, the three "CONFIRMED" inline bugs don't reproduce — their preconditions were resolved by the earlier fix commits (2f04bbf / 1ff965f). Details in the inline replies:
- **Stuck in edit mode:** parent gates on `isLoading`, so `savedNotes` is present at mount; and force-switching to read mode mid-edit would discard the user's typing.
- **aria-live not persistent:** the live-region `<span>` is the first child in both branches, so React reconciles it in place — only the `<div>`↔`<dl>` sibling swaps. Existing test covers it.
- **Whitespace dirty-flag:** the textarea *is* disabled in-flight (`isDisabled || isStepPatchInFlight` → forwarded to `TextArea`), so the whitespace-clobber path is unreachable.

All 823 ProcurementTracker unit tests pass (+7 new); lint and format clean.

josbell (changes requested):
## Review findings

### Bug 1 — Step-level Cancel leaves the notes editor stuck in edit mode *(manually confirmed)*

**Steps to reproduce:**
1. Navigate to a Procurement Tracker agreement with a PENDING step that already has a saved note (so the notes section renders in read mode with an "Edit Notes" button).
2. Fill out the step form fields (checkbox, assigned user, completion date) so the step-level Cancel button becomes enabled.
3. Click **Edit Notes** in the notes section — the textarea opens.
4. Click the step-level **Cancel** button and confirm the modal.

**Expected:** All form fields reset AND the notes editor returns to read mode.
**Actual:** The checkbox, user picker, and date fields all reset correctly, but the **notes textarea remains open** in edit mode.

**Root cause:** `isEditingNotes` is local state inside `StepNotesEditor` with no escape hatch for the parent. The step cancel handlers (`cancelStep1`, etc.) call `resetNotesX(...)` to restore the notes value but have no way to call `setIsEditingNotes(false)`.

**Fix direction:** Add a callback prop (e.g. `onExternalReset`) that the parent can call during step cancel to collapse the editor, or drive a `key` prop from a reset counter so the component remounts on step cancel.

---

### Bug 2 — `committedNotesRef` seeded empty on async data load; Cancel wipes the saved note *(statically confirmed)*

**Root cause:** `StepNotesEditor.jsx` line 107:
```js
const committedNotesRef = React.useRef(savedNotes ?? "");
```
When `stepOneData?.notes` is `undefined` at mount (RTK Query fetch still in-flight), the ref seeds to `""` and the editor opens directly in input mode (`isEditingNotes = true`). When data arrives, `useSaveNotes` syncs `notes` to the saved value and the `useEffect` promotes `hasEverSaved → true`, making the Cancel button visible. But `committedNotesRef.current` is still `""` — `enterEditMode()` (the only other write site) was never called because the editor started in input mode at mount.

Clicking Cancel at this point calls `resetNotes("")`, clearing the note from the UI until the next refetch.

**Fix direction:** Add a `useEffect` that updates `committedNotesRef.current` from `savedNotes` while edit mode was entered implicitly (i.e. without a user clicking "Edit Notes"):
```js
React.useEffect(() => {
    if (isEditingNotes && !hasEnteredEditModeExplicitlyRef.current) {
        committedNotesRef.current = notes;
    }
}, [notes, isEditingNotes]);
```
Or simpler: track whether `enterEditMode` has ever been called and skip the sync only after that.

---

### Bug 3 — Backend missing length validation on `notes` field

The frontend enforces a 750-character limit via `STEP_NOTES_MAX_LENGTH`, but the PATCH schema has no corresponding server-side constraint:

```python
# ops/schemas/procurement_tracker_steps.py ~line 420
notes = fields.String(required=False, allow_none=True)  # ← no validate.Length
```

By contrast, `requestor_notes` and `reviewer_notes` both have `validate=validate.Length(max=150)`. A client bypassing the UI could write an arbitrarily long note.

**Fix:**
```python
notes = fields.String(required=False, allow_none=True, validate=validate.Length(max=750))
```

frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx
```diff
@@ -33,72 +47,191 @@ const StepNotesEditor = ({
...
+    // The editor can return to read mode when a note has ever been saved or when
+    // the caller renders read mode by default (completed steps). A fresh active
+    // step with no saved note has no read view to cancel back to, so it stays in
+    // input mode and omits Cancel.
+    const canReturnToReadMode = hasEverSaved || startInReadMode;
+    const [isEditingNotes, setIsEditingNotes] = React.useState(!canReturnToReadMode);
+
+    // Leaving edit mode (Save or Cancel) unmounts the focused Save Notes/Cancel
+    // button; without this the browser resets focus to <body> (WCAG 2.4.3/2.4.7).
+    // Flag the return-to-read transition and move focus onto the Edit Notes button
```
josbell, line 85: **Bug (manually confirmed):** `isEditingNotes` is local state with no parent escape hatch. When a user clicks Edit Notes then confirms the step-level Cancel modal, `cancelStep1()` calls `resetStep1Notes` to restore the value but has no way to collapse the notes editor — the textarea stays open while all other form fields reset.

**Steps to reproduce:**
1. Open a PENDING step that already has a saved note (notes section shows in read mode with "Edit Notes")
2. Fill out the form fields (checkbox, user, date) so the step-level Cancel button is enabled
3. Click **Edit Notes** → textarea opens
4. Click the step-level **Cancel** and confirm the modal

**Expected:** All fields reset, notes editor returns to read mode
**Actual:** Form fields reset but the notes textarea stays open

**Fix direction:** Pass a reset callback (e.g. `onExternalReset`) into `StepNotesEditor` that the step cancel handler can call, or drive a `key` prop from a reset counter to remount the editor on step cancel.

frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx
```diff
@@ -33,72 +47,191 @@ const StepNotesEditor = ({
...
+    // THIS, not the `savedNotes` prop: after a save the prop stays stale (still the
+    // pre-save server value) until the RTK Query invalidation refetch lands, so
+    // cancelling back to it would briefly revert read mode to the old note (or flash
+    // "None" for a first-time note). Seeded from the server value and refreshed each
+    // time the user (re)enters edit mode.
+    const committedNotesRef = React.useRef(savedNotes ?? "");
+
+    // Screen-reader announcement for a completed save. The visual edit→read flip
+    // is the only save-completion cue for sighted users; assistive tech gets this
+    // aria-live region instead (it replaces the removed "Notes Saved" success
```
josbell, line 150: **Bug:** `committedNotesRef` is seeded with `savedNotes ?? ""` at mount. If `savedNotes` is `undefined` when the component mounts (RTK Query fetch still in-flight) the ref is `""`. When data later arrives, `useSaveNotes` syncs `notes` to the saved value and the `hasEverSaved` effect fires — making the Cancel button visible. But `committedNotesRef.current` is still `""` because `enterEditMode()` (the only other write site) was never called. Clicking Cancel at this point calls `resetNotes("")`, wiping the note from the UI until the next refetch.

**Fix direction:** Add a `useEffect` that keeps the ref in sync with `notes` while in implicit edit mode (i.e. before the user has explicitly clicked Edit Notes):
```js
React.useEffect(() => {
    if (isEditingNotes && !hasEnteredEditModeExplicitlyRef.current) {
        committedNotesRef.current = notes;
    }
}, [notes, isEditingNotes]);
```

josbell:
**Bug 3 — Backend missing length validation on `notes` field**

The frontend enforces 750 chars via `STEP_NOTES_MAX_LENGTH` but the PATCH schema has no matching constraint:

```python
# backend/ops_api/ops/schemas/procurement_tracker_steps.py line 420
notes = fields.String(required=False, allow_none=True)  # ← no validate.Length
```

`requestor_notes` and `reviewer_notes` (lines 434–439) both have `validate=validate.Length(max=150)`. A client bypassing the UI could POST an arbitrarily long note.

**Fix:**
```python
notes = fields.String(required=False, allow_none=True, validate=validate.Length(max=750))
```

weimiao67:
Thanks for the thorough review, @josbell — all three findings are addressed in 54f3c9ce7.

**Bug 1 — Step-level Cancel left the notes editor stuck in edit mode**
`useSaveNotes` now exposes a `notesResetKey` that bumps on every `resetNotes()`. Each step passes it to `StepNotesEditor` as a new `resetSignal` prop; the editor watches it and collapses back to read mode in lockstep with the value reset (skipping the initial mount, and only when there's a read view to return to — a fresh active step with no saved note stays in input mode). Wired through all six steps.

**Bug 2 — `committedNotesRef` seeded empty on async data load**
Added the sync you suggested, keyed on whether edit mode was entered *implicitly* vs. via the "Edit Notes" button (`enteredEditModeExplicitlyRef`). While in implicit edit mode the ref now tracks the incoming **`savedNotes`** (not the live `notes`), so once the async value lands Cancel discards back to the loaded server value rather than wiping the note or keeping in-progress keystrokes.

While fixing this I also found the **step-level Cancel** had the same stale-prop hazard: it passed the raw `stepXData?.notes` prop into reset, which stays stale after a first-time save until the RTK Query refetch lands, briefly flashing "None". Fixed at the root — `useSaveNotes` now tracks the last committed value (initial server value, then the just-saved value on each successful save), and `resetNotes()` with no argument restores it. All six step Cancel handlers now call the no-arg form.

**Bug 3 — Backend missing length validation on `notes`**
Added `validate=validate.Length(max=750)` to the PATCH schema's `notes` field, mirroring the frontend `STEP_NOTES_MAX_LENGTH`.

Regression tests added for all of the above (`StepNotesEditor.test.jsx`, `useSaveNotes.test.js`, `test_procurement_tracker_steps_schema.py`); the full tracker suite (833 frontend tests) and backend schema tests pass.

rajohnson90 (approved):

josbell (changes requested):
## Code Review Findings

### Bug 1 (High) — Stale `canReturnToReadMode` closure in `resetSignal` effect

**File:** `frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx` line 117

`canReturnToReadMode` is used inside the `resetSignal` \`useEffect\` but is omitted from the dependency array (suppressed with `// eslint-disable-line react-hooks/exhaustive-deps`). After a first-time save flips `hasEverSaved` to `true` mid-session, the effect permanently holds the stale `false` value — so a subsequent step-level Cancel bumps `resetSignal`, the `if (canReturnToReadMode)` guard evaluates `false`, and `setIsEditingNotes(false)` is never called. The textarea stays open after the step-level Cancel.

**Repro:** Open a fresh active step (no saved note) → type a note → click Save → click Edit Notes → trigger step-level Cancel → textarea stays open instead of collapsing.

**Fix:** Add `canReturnToReadMode` to the effect's dependency array and remove the eslint-disable comment.

---

### Bug 2 (Medium) — `isStepPatchInFlight` missing from date/solicitation save button disabled conditions

**Files:**
- `ProcurementTrackerStepTwo.jsx` line 80 — `isTargetCompletionDateSaveDisabled`
- `ProcurementTrackerStepThree.jsx` ~line 93 — `isSolicitationDatesSaveDisabled`
- `ProcurementTrackerStepFive.jsx` ~line 83 — `isTargetCompletionDateSaveDisabled`
- `ProcurementTrackerStepSix.jsx` ~line 84 — `isTargetCompletionDateSaveDisabled`

These save buttons are not disabled while a notes PATCH is in-flight, allowing a concurrent PATCH to fire on the same shared RTK Query mutation instance. The two responses can land out of order and revert whichever field loses the race. The step Complete buttons and `StepNotesEditor` itself both correctly include `isStepPatchInFlight` — these date save buttons should too.

**Fix:** OR `isStepPatchInFlight` into each of these disabled conditions.

---

### Bug 3 (Low/Plausible) — `enterEditMode` snapshots un-trimmed notes; Cancel restores whitespace-padded value

**File:** `frontend/src/components/Agreements/ProcurementTracker/StepNotesEditor/StepNotesEditor.jsx` line 161

`handleSaveNotes` persists `notes.trim()` but never writes the trimmed value back to `notes` state — it waits for the RTK Query refetch. In the window between a successful save and that refetch landing, `enterEditMode()` snapshots the un-trimmed `notes` state into `committedNotesRef`. Clicking Cancel immediately after restores the whitespace-padded original to the field instead of the clean saved value.

**Fix:** After a successful save, set `notes` state to the trimmed value (i.e. `setNotesState(notesToSave)`) so the snapshot in `enterEditMode` is always clean.

---

### Process — `openapi.yml` not updated after backend schema change

**File:** `backend/ops_api/ops/schemas/procurement_tracker_steps.py` line 12

The PR adds `validate=validate.Length(max=750)` to the `notes` field, but `backend/openapi.yml` was not updated. Per `backend/ops_api/CLAUDE.md`: *"After adding, changing, or removing any endpoint (route, resource method, or request/response schema), run the /sync-openapi skill to update the spec."*

**Fix:** Run `/sync-openapi` and commit the resulting `openapi.yml` update.

weimiao67:
@josbell Thanks for the thorough review! Addressed in e735a2486:

**Bug 2 (date/solicitation save buttons) — fixed.** OR'd `isStepPatchInFlight` into the disabled conditions for `ProcurementTrackerStepTwo` (`isTargetCompletionDateSaveDisabled` **and** `isRevisedTargetDateSaveDisabled` — the revised-target button fires the same shared `patchStepTwo` mutation, so it had the identical race), `ProcurementTrackerStepThree` (`isSolicitationDatesSaveDisabled`), `ProcurementTrackerStepFive`, and `ProcurementTrackerStepSix`. StepFour already included it; StepOne has no date-save button.

**Bug 3 (un-trimmed snapshot on Cancel) — fixed.** Applied in `useSaveNotes.js` rather than `StepNotesEditor` (that's where the notes state setter lives): after a successful save, when the field still matches what was persisted, it now also `setNotesState(notesToSave)` so any subsequent `enterEditMode` snapshot / Cancel restores the clean trimmed value.

**Process (openapi.yml) — fixed.** Added `maxLength: 750` plus a "Max 750 characters." note to the `notes` property of `UpdateProcurementTrackerStepRequest`, matching the existing spec convention (the 150-char requestor/reviewer limits are documented the same way).

**Bug 1 (stale `canReturnToReadMode` closure) — respectfully, I believe this is a false positive and would leave it as-is.** The `resetSignal` effect only *runs* when `resetSignal` changes, which requires a re-render, and `canReturnToReadMode` is recomputed every render — so the closure React executes is always from the latest committed render and reads the current value. By the time a step-level Cancel bumps `resetSignal`, `hasEverSaved` is already `true` and the guard evaluates `true`, so the textarea does collapse. I wasn't able to reproduce the described repro.

Adding `canReturnToReadMode` to the deps would also make the effect fire whenever `hasEverSaved` flips true — including when a server value arrives asynchronously (another user's edit) while the current user is mid-typing in a fresh active step — calling `setIsEditingNotes(false)` and collapsing the editor out from under them. Happy to add a test or a clarifying comment if that would help, but I'd prefer not to change the behavior here.

Santi-3rd (approved):

josbell (approved):

Issue #5849, opened 2026-06-23
Procurement Tracker Step 1-5 UI Adjustments


kimschulke:
### Goals
* Some minor UI adjustments on procurement tracker steps 1-5

### Tasks
- [x] The Task Completed By component should be filtered to show valid users only in the dropdown (team members listed on the agreement, DDs listed on an agreement, Team Leaders listed on an agreement, anyone with a Procurement Team user role, anyone with a Budget Team user role)

- [x] Check that the success notifications for step 5 pre-award (sent to approval, etc) are consistent with other success notifications across the app (for example editing, status change, etc). Currently, the pre-award success notification displays for 6 seconds and then dismisses itself.

- [x] Notes box maxed out at 750 characters but there’s no indication to the user to let them know their text is cut off, need to see if this is app wide or just proc tracker - check USWDS guidelines (see below in comment). Ideally this change will go out to all open-text components with a character limit, for example agreement description, project description, etc
- [ ] Assess need to update the business rule documentation and do so

### Additional Context
* Bonus points for addressing #5435 

### Resources
* [Designs for Procurement Tracker](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=24912-20912&t=g5OIMEbV5b8op6MQ-1)

npage123:
Found an example of counting down the characters left in a text box and the validation that appears when you go over the limit https://designsystem.digital.gov/components/character-count/. If its possible to use this globally that would be fantastic!

rajohnson90:
A bit confused about checkbox #1 because we already are only listing the Authorized Users. 

Looking through our logic only show the following users for the dropdown:

Created By User
Division Director for agreement
COR + ACOR
Team Members
Portfolio owners for any portfolio associated with agreement
Budget Team
System Owner
Admin Power User

npage123:
Here are my findings regarding task 2...

PROCUREMENT TRACKER
The pre-award approval from the DD shows the success alert over the top of other content with a white overlay behind it. Disappears in 6 seconds.

The pre-award approval from the Budget Team shows the success alert over the top of other content with no white overlay behind it. Disappears in 6 seconds.

AGREEMENT
Editing SCs & BLs, the pattern is the success alert above the content with a white overlay behind it. Disappears in 6 seconds.

Editing an agreement’s details shows the success alert over the top of other content with no white overlay behind it. Disappears in 6 seconds.

[...]

PR #6021, opened 2026-07-30
fix: fix award/pre-award success alert timing and add award approval banner

rajohnson90:
## What changed

This PR fixes a race condition in the Award/Pre-Award success alerts and adds an inline banner that was missing for the Award approval step.

**Alert timing fix** (`ApproveAwardApproval.hooks.js`, `RequestAwardApproval.hooks.js`, `RequestPreAwardApproval.hooks.js`): these hooks previously dispatched `setAlert(...)` (Redux) and then manually called `navigate(...)` in the same synchronous handler, with no `redirectUrl` on the alert. Since `navigate()` unmounts the current route before the `<Alert />` component (mounted globally in `DefaultLayout`) ever gets a chance to render the Redux update, the success banner could be torn down before it painted. The fix removes the manual `navigate()` calls and instead passes `redirectUrl` to `setAlert`, letting `Alert.jsx`'s own effect perform the navigation — guaranteeing the alert renders first. This matches the pattern already used by `ApprovePreAwardApproval.hooks.js` / `ReviewBudgetTeamRequisition.hooks.js`.

**`Agreement.jsx`**: removed the now-redundant `location.state`-driven `showPreAwardSuccessAlert`/`showAwardSuccessAlert` inline `SimpleAlert` machinery (dead code once the hooks above switched to `redirectUrl`-based alerts), and wired in the new `AwardApprovalAlert` component alongside the existing `PreAwardApprovalAlert`.

**`AwardApprovalAlert.jsx`** (new component): the Pre-Award approval flow shows an inline success banner on the agreement page when a `PreAwardApprovalNotification` (Approved/Declined) notification exists, but no equivalent existed for Award approval — even though the backend already creates and correctly serves `AwardApprovalNotification` records for the same `/notifications?agreement_id=` query. This component mirrors `PreAwardApprovalAlert` (same auto-dismiss-after-6s + manual-close pattern via `useDismissNotificationMutation`), filtered to `notification_type === "AWARD_APPROVAL_NOTIFICATION"` and `approval_status === "APPROVED"` only — the Award step has no "Declined" outcome, so there's no decline branch to handle.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5849

## How to test

1. **Award/Pre-Award alert timing:**
   - As COR, submit an agreement for Pre-Award Approval, then for Award Approval — confirm the success banner ("Agreement Sent to Pre-Award Approval" / "Agreement Sent to Award Approval") is visible before/during the redirect to the procurement tracker page, not skipped.
   - As Budget Team, approve an Award — confirm "Agreement Approved for Award" banner shows before redirecting to `/agreements?filter=change-requests`.
2. **New Award approval banner:**
   - Advance an agreement through Award approval (Budget Team approves step 6).
   - As the requester/PO/team member, open the agreement detail page — confirm an inline "Award Approved" success banner renders (in addition to the existing bell-icon notification), and auto-dismisses after 6 seconds or can be closed manually.
3. Run the frontend unit tests:
   ```bash
   cd frontend
   bunx vitest run src/components/Agreements/AwardApprovalAlert/AwardApprovalAlert.test.jsx src/components/Agreements/PreAwardApprovalAlert/PreAwardApprovalAlert.test.jsx
   ```

## A11y impact

## Storybook

## Screenshots

_Frontend changes — screenshots to be added._

## Links

N/A


rajohnson90:
Tested that Agreement Award does, as desired, shows a success alert now.

weimiao67 (approved):

josbell (approved):
## ✅ Approving — solid change, with a few low-severity cleanup recommendations

Reviewed the migration from navigation-state success alerts to the notification-based + `redirectUrl` pattern. The core logic holds up: I initially flagged two "banner overlap" concerns (In-Review warning showing alongside the Approved banner) and a swallowed dismiss-error, but on deeper tracing all three are **non-issues** —

- The removed `!showPreAwardSuccessAlert` / `!showAwardSuccessAlert` guards only ever suppressed the old *submission-moment* inline `SimpleAlert` (removed in the same change); they never gated `PreAwardApprovalAlert` / `AwardApprovalAlert`. The notification banners fire at the *approval* moment, where the backend states are mutually exclusive (`is_award_approval_requested` flips to `false` once approved; the pre-award requester's APPROVED notification only arrives once `requisition_approved_by` is set, which is exactly when `isPreAwardInReview` turns false).
- `dismissNotification` has no optimistic update (only `invalidatesTags`), so a failed dismiss leaves the notification unread/in-list and the `isError` banner renders fine.

None of these block. A few **low-severity cleanup recommendations** for a follow-up (or this PR if convenient):

### 1. `scrollToTop()` is now inconsistent and effectively a no-op
`Alert.jsx` navigates *immediately* on `redirectUrl` (no delay before `navigate()`; the 6s only controls auto-dismiss), and the destination auto-scrolls via `useScrollToTop` in `App.jsx`. So `scrollToTop()` on the outgoing page has no observable effect.
- `ApproveAwardApproval.hooks.js` calls it (no-op), the two Request hooks don't → inconsistent.
- **Recommendation:** pick one convention. Since `redirectUrl` navigates immediately, dropping `scrollToTop()` from all three is cleanest. If the intent is for the user to see the banner on the *source* page first, delay navigation in `Alert.jsx` instead — but then it should be applied uniformly. Worth a note in `frontend/CLAUDE.md`, since the documented rule assumes explicit `navigate()` calls.

### 2. Dead ternary in `Alert.jsx`
```js
let waitTime = redirectUrl ? 6000 : 6000; // both branches identical
```
Collapse to `const waitTime = 6000;` (or make the branches actually differ if that was the intent).

### 3. Dead state in `Agreement.jsx`
```js
const [isAwardAlertVisible] = useState(true); // setter discarded, always true
```
Same for `isPreAwardAlertVisible`. Replace with a plain `const` or inline it.

### 4. `AwardApprovalAlert` duplicates `PreAwardApprovalAlert` (~85%)
The `isApprovedNotification` predicate, the `useRef(new Map())` + two-`useEffect` auto-dismiss timer block (~37 lines), `handleDismiss`, the error-banner JSX, and the PropTypes shape are near-verbatim copies. A third approval type would mean a third copy.
- **Recommendation:** extract a `useAutoDismissNotifications(notifications, onDismiss, delay)` hook plus a shared predicate/propType. Good follow-up refactor rather than a blocker.

### 5. Missing Storybook story
`AwardApprovalAlert` is under `frontend/src/components/**` with no co-located `AwardApprovalAlert.stories.jsx`. CI won't warn (the check only covers `src/components/UI/`), but `frontend/CLAUDE.md` documents the convention.

Nice work — approving.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #787, opened 2023-03-27
Create a Grant


jonnalley:
### User Story
As an OPRE staff member, I want to create a new grant so that I can facilitate research or program work in support of a project goal or topic.

### Acceptance Criteria
- [x] User can provide required attributes for and create a GrantAgreement with all required attributes, including attributes unique to a grant, that is persistently stored
- [x] User should see an indication of some kind that the creation was successful
- [x] User can add budget line(s) to the grant with attributes specific to a grant's BL
- [ ] All Claude temp story docs should be cleared before merging
- [ ] All review comments should be cleaned up

-----------------------------------------

#### UX
- [x] Adapt contract design to grants
- [x] Share designs with stakeholders and other OPRE team members for feedback
- [x] Refine the design based on feedback
- [x] Provide design assets with annotations to developers

#### Vertical Slices
- [x] #5925
- [x] #5926
- [x] #5927
- [x] #5928

#### Dev
***Backend***
- [x] Update backend model
  - [x] Update `GrantAgreement` to contain all the specific properties needed.
- [ ] Update Endpoints
  - [x] Create `GET`, `POST`, `PUT`, `PATCH` endpoints for `Grant` agreements. 
  - [ ] Create business logic rules
  - [ ] Validate business logic against any created rules.

***Frontend***

**DEV DISCUSS** - Determine how we want to change the controls displayed for different Agreement Types

- [ ] Update frontend components
  - [ ] Create `Grant` Component for `Create Agreement` ..OR.. determine how the individual controls should be displayed/hidden.

#### Manual Test Cases

**Positive cases:**
- A user can create a Grant agreement providing Type, Name, Nickname, and Description, and the created Grant is persistently stored and retrievable
- A user sees an indication of successful creation (e.g. success alert/redirect to agreement details)
- A user can add one or more Grant Numbers to a Grant agreement, and the added Grant Number(s) appear in the list
- A user can create one or more Grant Budget Lines on a Grant agreement with grant-specific attributes, and they appear on the agreement

**Negative / edge cases:**
- Attempting to create a Grant agreement without a required field (e.g. Name) shows a validation error and does not create the agreement
- Attempting to create a Grant agreement with a Name/Nickname that duplicates an existing agreement is rejected with a clear validation error
- Attempting to add a Grant Number or Grant Budget Line without first saving the parent Grant agreement is prevented or clearly disallowed in the UI
[...]

PR #6059, opened 2026-08-06
feat(OPS-787): grant creation workflow feedback fixes

rajohnson90:
## What changed

Addresses a round of feedback on the create-grant agreement workflow (OPS-787): makes NOFO Number optional, reworks ALN Numbers into a proper multiselect (1–10, ascending), converts Grant Funding Period into a dropdown, fixes budget-line/grant-number UX issues in the grant BLI creation flow, and aligns field widths with the design.

**Backend**
- New migration + model/schema change replacing the single `aln_number` column with an `aln_numbers` array field on agreements.
- `openapi.yml` and related schema tests updated to match.

**Frontend — Grant Details form (`AgreementEditForm`)**
- NOFO Number is no longer required to save a draft or continue.
- Added `AlnNumbersComboBox` — a multiselect combobox for choosing ALN Numbers 1–10 (fixed placeholder list, ascending order).
- Added `GrantFundingPeriodSelect` — Grant Funding Period is now a dropdown instead of free text.
- Shrunk NOFO Number, Grant Funding Period, and ALN Numbers to match the Team Members combobox width (15rem), including their dropdown widths.
- Fixed subheader copy/instructions across the grant creation steps.

**Frontend — Grant budget lines & numbers (`CreateBLIsAndSCs`, `AgreementEditorContext`)**
- Budget line items now live in the shared `AgreementEditorContext` (via a reducer) instead of local component state, fixing budget lines disappearing when navigating between steps 2 and 3 of grant creation.
- Budget line edit/delete lookups now match by `id` instead of array index, since the temp and original budget-line arrays can drift out of index alignment after add/delete/duplicate.
- Added a "no budget lines yet" message and creation instructions for grant BLIs.
- Grant numbers now show "N of M" numbering consistently across the approval, review, and pre-award views.
- Disabled the edit button from the agreements list where editing isn't yet supported, and fixed the FPO/description labeling for grant numbers.

## Issue

https://github.com/HHS/OPRE-OPS/issues/787

## How to test

Backend:
```bash
cd backend
pipenv run pytest ops_api/tests/ops/schemas/test_agreements.py ops_api/tests/ops/agreement/test_agreement.py ops_api/tests/ops/agreement/test_agreement_change_requests.py
```

Frontend:
```bash
cd frontend
bun run test --watch=false
bun run test:e2e --config-file ./cypress.config.js --headless --spec "cypress/e2e/createGrantAgreement.cy.js"
```

Manual verification:
1. Start a new agreement, select **Grant** as the type.
2. Confirm you can save a draft / continue without entering a NOFO Number.
3. Select multiple ALN Numbers (1–10) from the combobox — confirm the list renders ascending and matches the Team Members field width.
4. Select a Grant Funding Period from the dropdown.
5. Add multiple grant numbers and budget lines; navigate back and forth between steps 2 and 3 and confirm budget lines persist.
6. Edit/delete a budget line and confirm the correct line is affected (not a stale index).
7. Check grant number numbering ("Grant 1 of 2", etc.) on the approval, review, and pre-award requisition views.

## A11y impact

## Storybook

## Screenshots

_TODO: add before/after screenshots of the Grant Details form (field widths, ALN multiselect, funding period dropdown)_

## Links

N/A


weimiao67 (approved):

josbell (changes requested):
## Review

Good overall structure — the BLI state lift into `AgreementEditorContext` is well-designed and the id-based lookup fix is the right call. Two confirmed bugs need fixes before merge.

---

### 🔴 Bug 1 — `GrantFundingPeriodSelect`: deselecting the placeholder sends `NaN` to state

**File:** `frontend/src/components/Agreements/GrantFundingPeriodSelect/GrantFundingPeriodSelect.jsx` line 22

`Select.jsx` renders the placeholder as `<option value={null}>`. React removes the `value` attribute entirely, so the browser returns the option's *text content* as `e.target.value` (i.e. `"-Select Grant Funding Period-"`). The guard `selectedValue === ""` is never true for that string, so `Number("-Select Grant Funding Period-") = NaN` reaches `setFundingPeriodMonths` and eventually the API.

**Trigger:** User picks "18 Months", then re-selects the placeholder. NaN is saved to state and sent on the next API call.

**Fix:** Render the placeholder with an explicit empty value in `GrantFundingPeriodSelect`, or change the guard:
```js
onChange={(name, selectedValue) => onChange(name, !selectedValue || selectedValue === "-Select Grant Funding Period-" ? null : Number(selectedValue))}
```

---

### 🔴 Bug 2 — `DELETE_BUDGET_LINE_ITEM`: unsaved BLIs get queued for API deletion

**File:** `frontend/src/components/Agreements/AgreementEditor/AgreementEditorContext.hooks.js` line 188

`DELETE_BUDGET_LINE_ITEM` unconditionally pushes the full payload into `deleted_budget_line_items_ids` with no guard on whether the id is a real persisted id. The sibling cases `DELETE_SERVICE_COMPONENT` and `DELETE_GRANT_NUMBER` both guard with:
```js
action.payload.id ? [...push id] : [...no change]
```

**Trigger:** In the edit-agreement flow, user adds a new BLI (crypto-random string id), then deletes it before saving. `handleDeletions` calls `deleteBudgetLineItem("abc123xyz0").unwrap()` → 404 → `.unwrap()` throws → save error.

**Fix:** Apply the same guard as the sibling reducers:
```js
deleted_budget_line_items_ids: action.payload.id
    ? [...state.deleted_budget_line_items_ids, action.payload.id]
    : [...state.deleted_budget_line_items_ids]
```
Also: the field currently stores full objects while `deleted_services_components_ids` and `deleted_grant_numbers_ids` store plain ids — worth aligning the type while you're in here.

---

### 🟡 Minor — `AgreementTableRow.jsx` line 63: string literal instead of constant

`agreement?.agreement_type === "GRANT"` — the `AGREEMENT_TYPES.GRANT` constant is already imported and used correctly 6 lines earlier in the same file. Swap to the constant.

### 🟡 Minor — `AgreementEditForm.jsx` ~line 549: dead `messages` prop on NOFO Number input

The PR deletes the `nofo_number` Vest test, so `res.getErrors("nofo_number")` always returns `[]`. Remove the `messages` prop to avoid misleading future readers into thinking NOFO is still validated through the suite.

### 📝 Note — `AgreementTableRow.jsx` line 56: comment points to the wrong file

The comment says `isNotDevelopedYet` "gates whether GrantNumbers renders on the Agreement Details page" — but that rendering is gated by `isGrant` + grouped budget line existence, not `isNotDevelopedYet`. The actual downstream consumers are `DetailsTabs.jsx` (Procurement Tracker tab) and `AgreementDetails.jsx` (edit button). Worth correcting so the next person hunting here doesn't look in the wrong place.

rajohnson90:
Addressed Josbell's comments and re-requesting review.

josbell (approved):
Overall this is solid work — the BLI-state-to-context refactor is well-structured, the id-based edit/delete fix is correct, and the test coverage is thorough. A few suggestions below.

**Bug — sub-component BLI link cleared on edit** (`CreateBLIsAndSCs.hooks.js:673`)

`handleEditBLI` stamps the contract-path payload with:
```js
serviceComponentGroupingLabel: (servicesComponentNumber ?? 0).toString()
```
This produces a bare number string (e.g. `"2"`), but `addServiceComponentIdToBLI` matches against the full label including sub-component (e.g. `"2-A"`). For any BLI under an SC with a sub-component, the label won't match, `services_component_id` is set to `null`, and the BLI's SC link is silently cleared on save. The grant path is fine; this is the else-branch only. Pre-existing code, but the function was touched in this PR — good time to fix.

**Migration comment** (`replace_aln_number_with_aln_numbers.py:29`)

The migration drops `aln_number` without migrating existing values. Since this column was never populated in production, no data loss occurs — but a future reader won't know that. Consider adding a comment like:

```python
# aln_number was added in the previous migration but never populated in production,
# so no data migration is needed before dropping it.
```

**Schema — `load_default=[]` on `aln_numbers`** (`schemas/agreements.py:116`)

`allow_none=True, load_default=[]` means a PATCH body that omits `aln_numbers` will deserialize it as `[]` and potentially overwrite stored values, depending on whether the endpoint uses `partial=True` or `unknown=EXCLUDE`. Worth confirming the PATCH handler is safe here (if it already is, this is a no-op concern).

rajohnson90:
I don't think people generally look at the migration code because they are originally autogenerated so I'm going to leave that one. Verified that the aln_numbers is a non-issue because of partial=True being used for the marshmallow schema.

Issue #2280, opened 2024-05-06
Step 6 of Procurement Tracker - Award (Approver)


kimschulke:
### User Story
As a budget team member, I need to insure the accuracy and completeness of all contract artifacts in OPS and obligate budget lines.

### Preconditions
A contract has one or more budget line in Executing status and the contract has already advanced past the fifth step of the procurement tracker

### Acceptance Criteria
- [ ] User can view all artifacts uploaded previously for this contract
- [ ] User can view and edit all budget lines for the contract
- [ ] User can view and edit all contract metadata entered as part of this milestone (that is being submitted for review)
- [ ] User can approve all submissions
- [x] The affected budget lines will get updated to `Obligated` status
- [ ] The COR, project team, team lead, and budget team should all get notified of this final approval
- [x] The contract should be considered `awarded` if not already
- [ ] If any error is encountered in persisting the change(s), that error should be presented to the user
- [x] The relevant entries should be recorded in the event/history table(s) and be able to be properly redered properly in the history of changes to the agreement

-----------------------------------------

#### UX
- [x] Explore design options
- [x] Share design with team to discuss development considerations
- [x] Refine the design based on feedback
- [x] Share designs with stakeholders and other OPRE team members
- [x] Refine the design based on feedback
- [x] Provide design assets with annotations to developers

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

### Additional Context & Resources
A story pertaining to the prior step in the contract's lifecycle is #1518

jonnalley:
This issue originally existed as #1641

npage123:
@josbell The obligated date should be required before the 'Approve Award' button is enabled please. @kimschulke informed me that that date, if not entered, can't be assumed its the current date which is what happens currently in staging as it is generally first documented in another system.

https://github.com/user-attachments/assets/2d1c163d-8bfb-48b6-a8f2-8215aa0e3895

Santi-3rd:
@npage123 I worked on the issue and this is how it's looking so far. Let me know if you have any thoughts 

https://github.com/user-attachments/assets/004d524f-2b87-4a57-8ba7-9cbdb306b904

npage123:
Looks good @Santi-3rd , please move forward with the PR.

npage123:
[...]

PR #6079, opened 2026-08-10
Step 6: Obligated date is required before approval

Santi-3rd:
## What changed

Require an explicit **Obligated Date** when approving an AWARD step. Previously the date defaulted to today when omitted; it's now mandatory on both the API and the UI, since the real obligation date is documented in another system first.

**Backend**
- New `AwardApprovalObligatedDateRequiredRule` rejects an AWARD approval without `obligated_date` (400).
- Award approval no longer falls back to `date.today()` — the provided `obligated_date` is used to set BLI `date_needed` and the action's `date_awarded_obligated`.

**Frontend**
- New Vest suite validates the Obligated Date (required, `MM/DD/YYYY`, real calendar date).
- The date field shows inline errors and the **Approve Award** button stays disabled until a valid date is entered.

## Issue

#2280

## How to test

1. Open an agreement's AWARD step approval and leave **Obligated Date** blank — the **Approve Award** button is disabled.
2. Enter an invalid date (e.g. `13/45/2024` or `02/31/2024`) — an inline error appears and the button stays disabled.
3. Enter a valid `MM/DD/YYYY` date and approve — the request succeeds; IN_EXECUTION BLIs move to OBLIGATED with `date_needed` set to that date.
4. API check: `PATCH /api/v1/procurement-tracker-steps/{id}` with `{"approval_status": "APPROVED"}` and no `obligated_date` returns 400.

## A11y impact

## Storybook

## Screenshots

_N/A_

## Links

_N/A_


josbell (changes requested):

6 confirmed findings, ranked by severity.

---

### 🔴 [1] `bli.date_needed` unconditionally overwritten with None
**`backend/ops_api/ops/services/procurement_tracker_steps.py:1039`**

The `if obligated_date is not None:` guard was removed. `bli.date_needed = obligated_date` now fires unconditionally for every IN_EXECUTION BLI. Any caller that bypasses the validator (direct service call, test, admin script) with `obligated_date=None` silently nulls out every IN_EXECUTION BLI's date.

Seven existing tests in `test_award_approval_service.py` (lines 82, 94, 106, 118, 132, 162, 181) call `_handle_award_approval(step, "APPROVED", None, user)` directly with IN_EXECUTION BLIs — they still pass because they only assert `bli.status`, never `bli.date_needed`. The guard should be restored at the service layer as defense-in-depth.

---

### 🔴 [2] `date_awarded_obligated` can be stored as None
**`backend/ops_api/ops/services/procurement_tracker_steps.py:1051`**

`obligated_date or date.today()` → `obligated_date`. The column is `Optional[date]` with `nullable=True` (`models/procurement_action.py:115`). If `_handle_award_approval` is called without the validator (any direct/internal call), `None` is written to the column. `test_new_award_sets_awarded_status` exercises exactly this path with `obligated_date=None` but only asserts `status == AWARDED`, never the date — false confidence.

---

### 🟡 [3] Stale Vest singleton causes phantom errors on remount
**`frontend/src/pages/agreements/award-approval/ApproveAwardApproval.hooks.js:77`**

```js
const validatorRes = suite.get();  // reads module-level singleton on every render
```

The suite is a module-level singleton. If a user visits the page, triggers a validation error, navigates away, then returns — `suite.get()` on the fresh mount returns stale errors from the prior session, rendering phantom error messages before any input. There is no `suite.reset()` anywhere in the component lifecycle.

`RequestAwardApproval.hooks.js` avoids this with:
```js
const [validationResult, setValidationResult] = useState(suite.get());
// only updated via setValidationResult(...) inside runValidate
```
The same pattern should be used here.

---

### 🟡 [4] Validation test passes `obligated_date` as a string, not `datetime.date`
**`backend/ops_api/tests/ops/validation/test_award_approval_validation.py:197`**

`test_passes_when_obligated_date_provided` passes `{"obligated_date": "2026-07-15"}` — a string. In production, the PATCH request goes through `ProcurementTrackerStepPatchRequestSchema.load()` which deserializes `obligated_date` via `fields.Date` into a `datetime.date` before the validator runs. The test doesn't represent the real call path. Use `date(2026, 7, 15)` to match production and to stay consistent with `test_award_approval_service.py`.

---

### 🔵 [5] `isValidDateFormat` / `parseDateString` duplicated across 4 suite files; regex written twice in the same file
**`frontend/src/pages/agreements/award-approval/ApproveAwardApproval.suite.js:6`**

These helpers are already identical in `RequestAwardApproval.suite.js`, `ProcurementTrackerStepFive/suite.js`, and `ProcurementTrackerStepSix/suite.js`. This PR adds a 4th copy. Additionally, the regex `/^(0[1-9]|1[0-2])\/(0[1-9]|[12][0-9]|3[01])\/\d{4}$/` appears twice in the same file: line 7 (inside `isValidDateFormat`) and line 37 (inside `enforce().matches()`). StepFive/Six avoid this with a named `DATE_FORMAT_REGEX` constant. Consider extracting shared helpers to `frontend/src/helpers/` to prevent further drift.

---

### 🔵 [6] `React.memo` on `MemoizedDatePicker` is a no-op; and `DatePicker` is already memo'd at source
**`frontend/src/pages/agreements/award-approval/ApproveAwardApproval.hooks.js:12` / `ApproveAwardApproval.jsx:233`**

Two separate issues on the same component:

1. **Double-wrap**: `DatePicker` is already exported as `React.memo(DatePicker)` at its source. Wrapping it again at line 12 of the hooks file creates a *new anonymous component type* whose identity changes on hot-reload — React unmounts and remounts the picker instead of diffing it, losing focus and partially-typed input. The sibling `ProcurementTrackerStepSix.hooks.js` already has the fix: `const MemoizedDatePicker = DatePicker; // DatePicker is already React.memo'd at source`.

2. **Memo is a no-op anyway**: `runValidate` has no `useCallback`, so it gets a new reference every render. The `onChange` wraps it in a further inline arrow, meaning `MemoizedDatePicker` always sees a new prop and re-renders unconditionally. `messages={validatorRes.getErrors("obligatedDate") || []}` also produces a new array reference each render. Either stabilize the props with `useCallback` or remove the wrapper to avoid misleading future readers.

---

### 🔵 [7] `only(fieldName)` in the new suite violates the CLAUDE.md Vest v6 rule
**`frontend/src/pages/agreements/award-approval/ApproveAwardApproval.suite.js:26`**

```js
if (fieldName) {
    only(fieldName);
}
```

Per the project's Vest v6 conventions: "In vest v6, `only(fieldName)` skips all other tests and those tests retain their previous failed state, causing phantom validation errors." All three tests in this suite target the same `"obligatedDate"` field, so `only()` provides no benefit today. But if a second field is ever added, any `runValidate("obligatedDate", value)` call will leave the new field's prior error state intact, keeping the Approve button disabled even after both fields are validly filled. The recommended fix is to omit `only()` entirely for a single-field suite.

Santi-3rd:
@josbell Thank you for the feedback. I addressed all the findings you pointed out

  1. ✅ Restored `if obligated_date is not None:` guard before `bli.date_needed = obligated_date` (defense-in-depth for callers that bypass the validator).
  2. ✅ Guarded `date_awarded_obligated` assignment with `and obligated_date is not None`.
  3. ✅ Vest result now held in `useState(suite.get())` and updated only via `runValidate` — no more stale singleton errors on remount (matches `RequestAwardApproval.hooks.js`).
  4. ✅ Test now passes `date(2026, 7, 15)` instead of a string, matching the real post-deserialization call path.
  5. ✅ (partial) Added local `DATE_FORMAT_REGEX` constant, removing the intra-file regex duplication. Deferred the cross-file `helpers/` extraction as a separate cleanup.
  6. ✅ `MemoizedDatePicker = DatePicker` (removed the double `React.memo` wrap); dropped now-unused `React` import.
  7. ✅ Removed `only(fieldName)` (and unused `only` import/param) per the Vest v6 convention.

rajohnson90 (approved):

weimiao67 (approved):

josbell (approved):

PR #6097, opened 2026-08-14
ci: notify Slack when a user is @-mentioned in a GitHub comment

rajohnson90:
## What changed

Adds a GitHub Action that posts a Slack notification whenever a GitHub username is @-mentioned in an issue or PR comment.

**`.github/workflows/slack_mention_notify.yml`**
- Triggers on `issue_comment: created`, extracts `@username` mentions from the comment body (ignoring mentions inside fenced code blocks), and resolves each to a real Slack member ID via `.github/slack_user_map.json`, falling back to plain `@username` text for unmapped users.
- Posts the resulting message to a Slack channel via an Incoming Webhook (`secrets.SLACK_WEBHOOK_URL`).
- Includes hardening added after an adversarial security review (threat model: a malicious actor with only comment permissions, or a compromised/malicious insider):
  - **Access gate**: only runs for comments from `OWNER`/`MEMBER`/`COLLABORATOR`, so a random public-repo commenter can't trigger real Slack pings.
  - **Mrkdwn escaping**: the issue/PR title is HTML-entity-escaped (`&`, `<`, `>`) before being embedded in the Slack message, preventing an attacker-controlled title from breaking out of the link syntax to inject a fake link or a `<!channel>` mass-ping.
  - **Rate limiting**: a first step queries GitHub's own workflow-run history for the commenting actor and skips the notification if that actor has triggered this workflow more than 5 times in the last 10 minutes — bounding harassment/spam potential from a single compromised or malicious insider account.

**`.github/slack_user_map.json`**
- New mapping of GitHub usernames to Slack member IDs used to resolve mentions to real Slack pings.

## Issue

N/A

## How to test

1. Ensure the `SLACK_WEBHOOK_URL` repo secret is configured (Settings → Secrets and variables → Actions) pointing at the desired Slack channel's Incoming Webhook.
2. As an org member/collaborator, comment on any issue or PR with `@<a-github-username-from-slack_user_map.json>` and confirm a message appears in the target Slack channel with a real `<@member-id>` mention.
3. Comment with `@some-unmapped-username` and confirm the notification still posts, falling back to plain `@some-unmapped-username` text.
4. Comment with a title/body containing `<`, `>`, or `&` characters (e.g. open an issue titled `Test > <!channel> test`) and mention someone, then confirm in Slack that the message renders as plain text rather than as a live `@channel` ping or a broken/injected link.
5. Trigger the workflow more than 5 times within 10 minutes as the same actor and confirm the 6th+ run logs a `::warning::` and skips the checkout/notify steps (visible in the Actions run log).
6. Confirm the workflow does not fire at all when a non-collaborator/non-member comments (check `author_association` via a test account without repo permissions, if available).

## A11y impact

## Storybook

## Screenshots

N/A — backend/CI-only change, no UI

## Links

N/A

### DIFF
--- a/.github/slack_user_map.json
+++ b/.github/slack_user_map.json
@@ -1,10 +1,9 @@
 {
     "_comment": "Map GitHub usernames (lowercase) to Slack member IDs, used by slack_mention_notify.yml to turn @github-mentions into real Slack pings. Find a Slack member ID via a user's profile > More (kebab menu) > Copy member ID.",
-    "octocat": "U0123456789",
     "rajohnson90": "U070RCWHMCL",
     "weimiao67": "U04P2UN6N0K",
     "josbell": "U02D0HFJ725",
-    "Santi-3rd": "U074V0UH1D4",
+    "santi-3rd": "U074V0UH1D4",
     "jonnalley": "U02FKHMF2GN",
     "npage123": "U02933CUQ",
     "kimschulke": "U013RKN5CEN",
--- a/.github/slack_user_map.json
+++ b/.github/slack_user_map.json
@@ -1,0 +1,12 @@
+{
+    "_comment": "Map GitHub usernames (lowercase) to Slack member IDs, used by slack_mention_notify.yml to turn @github-mentions into real Slack pings. Find a Slack member ID via a user's profile > More (kebab menu) > Copy member ID.",
+    "octocat": "U0123456789",
+    "rajohnson90": "U070RCWHMCL",
+    "weimiao67": "U04P2UN6N0K",
+    "josbell": "U02D0HFJ725",
+    "Santi-3rd": "U074V0UH1D4",
+    "jonnalley": "U02FKHMF2GN",
+    "npage123": "U02933CUQ",
+    "kimschulke": "U013RKN5CEN",
+    "andressacurvelo": "U03U61WS6Q0"
+}
--- a/.github/workflows/slack_mention_notify.yml
+++ b/.github/workflows/slack_mention_notify.yml
@@ -1,0 +1,96 @@
+name: Notify Slack on PR/Issue Mention
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+  actions: read
+
+jobs:
+  notify-mention:
+    runs-on: ubuntu-latest
+    # Restrict to org members/collaborators so a public-repo commenter can't trigger real Slack pings
+    if: contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
+    steps:
+      - name: Rate limit check
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          ACTOR: ${{ github.event.comment.user.login }}
+          MAX_RUNS: "5"
+          WINDOW_MINUTES: "10"
+        run: |
+          # Even a trusted member/collaborator account can be compromised or turn malicious,
+          # so cap how many times this workflow can fire per commenter in a rolling window --
+          # using GitHub's own run history as state, since Actions has no persistent store.
+          since=$(date -u -d "${WINDOW_MINUTES} minutes ago" +%Y-%m-%dT%H:%M:%SZ)
+          run_count=$(gh api "repos/${REPO}/actions/workflows/slack_mention_notify.yml/runs" \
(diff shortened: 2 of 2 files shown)


rajohnson90:
Two manual steps left to make this PR work:
1. Add the SLACK_WEBHOOK_URL repo secret.
    - In Slack: create an Incoming Webhook for whichever channel should receive these notifications (Slack App admin → Incoming Webhooks → Add
  New Webhook to Workspace, pick the channel, copy the URL).
    - In GitHub: Settings → Secrets and variables → Actions → New repository secret, name it SLACK_WEBHOOK_URL, paste the webhook URL.
  2. Double-check slack_user_map.json is accurate before merge. You already populated it with real Slack member IDs for the team — just
  confirm those are current and remove the leftover "octocat": "U0123456789" placeholder entry if you don't want it lingering in the map.

weimiao67 (approved):

.github/slack_user_map.json
```diff
@@ -0,0 +1,12 @@
+{
+    "_comment": "Map GitHub usernames (lowercase) to Slack member IDs, used by slack_mention_notify.yml to turn @github-mentions into real Slack pings. Find a Slack member ID via a user's profile > More (kebab menu) > Copy member ID.",
+    "octocat": "U0123456789",
```
Santi-3rd: Pointing out that octocat is still here since there was a comment about removing it.

Santi-3rd (approved):

josbell (changes requested):
Reviewed the two files across security, shell correctness, and CI best-practices. Overall this is a well-built workflow — correct env-var indirection for untrusted input, SHA-pinned checkout, least-privilege permissions, sparse checkout, jq-built payload, and complete mrkdwn escaping of the title. **No security vulnerabilities found.** Requesting changes for two correctness bugs that will break normal operation, plus a placeholder to remove.

## Must-fix

**1. Workflow fails ❌ on every comment without an @mention** — `.github/workflows/slack_mention_notify.yml:62-69`
GitHub's default shell is `bash -eo pipefail`. When a comment has no mention, `grep` exits 1 → `pipefail` propagates → `set -e` aborts the step at the `mentions=$(...)` assignment, *before* the graceful `exit 0` at line 66. Since the workflow fires on every comment, the common case (comments with no mention) will show a red failure. Fix — stop the empty grep from propagating:
```sh
mentions=$(printf '%s' "$body_no_code" \
  | { grep -oP '(?<![\w.@-])@[a-zA-Z0-9](?:-?[a-zA-Z0-9]){0,38}\b' || true; } \
  | sed 's/^@//' | sort -u)
```

**2. Mixed-case map key never resolves** — `.github/slack_user_map.json:7` + workflow lines 75-76
The lookup lowercases the username via `tr`, but the JSON key is `"Santi-3rd"` (capital S). jq object indexing is case-sensitive, so it never matches and falls back to plain `@Santi-3rd` text — no real Slack ping, ever. The file's own `_comment` says keys must be lowercase. Fix: change the key to `"santi-3rd"`.

**3. Placeholder Slack ID** — `.github/slack_user_map.json:3`
`"octocat": "U0123456789"` looks like leftover sample data; a mention of `@octocat` would post an unresolvable `<@U0123456789>`. Remove the entry.

## Nice-to-have (non-blocking)

- **Rate-limit race (defense-in-depth only):** no `concurrency:` group, so a burst of comments can each read `total_count` before siblings register, bypassing the 5/10-min cap. Consider `concurrency: { group: slack-notify-${{ github.event.comment.user.login }}, cancel-in-progress: false }` (matches the `deploy_reusable.yml` pattern).
- **Hardcoded workflow filename** in the rate-limit `gh api` path (line 29): a future rename 404s the call → `run_count` empty → limiter silently disabled. Add a comment tying the two together, or derive from `github.workflow_ref`.
- **No `timeout-minutes`** on the job (inherits 6h default); add `timeout-minutes: 5` to match `security_snyk.yml`/`storybook_build.yml`.
- **`curl -sf`** hides the Slack error body — `--fail-with-body` would make failures like `invalid_payload` diagnosable (line 96).
- **Fenced-code stripping** (line 60): an unclosed ``` fence deletes to EOF (dropping later mentions); indented/`~~~` fences aren't stripped. Detection-accuracy only — worth documenting the limitation.
- **Public-repo exposure:** `slack_user_map.json` commits internal Slack member IDs + org roster in plaintext to a public repo. Not a code defect, but worth a conscious team decision.

josbell (approved):

rajohnson90:
@rajohnson90 Test

Issue #6094, opened 2026-08-13
Skip Change Request review for Draft→Planned and in-Planned budget edits


jonnalley:
### User Story
As an OPS user, I want my change to a budget line's status from Draft to Planned, or my edit to budget details on a budget line that stays Planned, to take effect as soon as it meets all required data and validation rules, so that I don't have to wait on another user's review and approval.

### Acceptance Criteria
- [ ] When the capability is enabled, changing a budget line's status from Draft to Planned applies immediately once all existing data requirements for that status are met — it does not require another user's review and approval first.
- [ ] When the capability is enabled, editing budget details (e.g. amount, funding source, need-by date) on a budget line that is already Planned and remains Planned applies immediately — it does not require another user's review and approval first.
- [ ] A budget line changed under this capability does not appear as In Review once the change is applied.
- [ ] No approval request is generated or sent to a reviewer for a change made under this capability and conditions.
- [ ] All existing data validation and completeness requirements for moving a budget line to Planned status still apply and still block the change when unmet, regardless of whether the capability is enabled.
- [ ] A user cannot combine a status change with other budget-detail edits in a single change — this existing constraint is unaffected by this capability.
- [ ] Budget line types and agreement types that support Draft and Planned status behave consistently under this capability — none are unintentionally excluded or handled differently.
- [ ] Status changes other than Draft→Planned (e.g. Planned to a later status) are unaffected by this capability and continue to require review and approval exactly as they do today.
- [ ] Users and workflows that already bypass review and approval today continue to work as they do today; this capability does not remove or duplicate their existing behavior.
- [ ] The capability can be turned on or off independently per environment, and each environment's setting takes effect without requiring an application code change.
- [ ] When the capability is turned off, Draft→Planned status changes and in-Planned budget edits behave exactly as they do today — requiring review and approval — with no observable difference to users.
- [ ] A user cannot tell, from the change itself, that review and approval was skipped versus never having been required — the resulting budget line looks and behaves the same either way.

### Testing

**Automated — positive cases**
- Backend: a Draft budget line with all required fields present transitions to Planned and its new values are readable immediately after the request completes, with no pending/approval record created for that change, when the capability is on.
- Backend: an existing Planned budget line has a budget-detail field (amount, funding source, need-by date) changed, the new value is readable immediately, and no pending/approval record is created, when the capability is on.
[...]

PR #6105, opened 2026-08-17
feat: skip Change Request for Draft→Planned and in-Planned budget edits

weimiao67:
## What changed

Adds a per-environment capability that lets two budget-line changes apply immediately instead of routing through Change Request (CR) review and Division Director approval:

1. **Draft → Planned** status changes
2. **In-Planned budget-detail edits** (amount, funding source, need-by date) on a budget line that stays Planned

The behavior is gated by a new `SKIP_CR_FOR_DRAFT_PLANNED` setting (env var, default **OFF** everywhere), so each environment can enable or disable it without a code change.

**Backend**
- New `SKIP_CR_FOR_DRAFT_PLANNED` setting in `default_settings.py`, read from the `SKIP_CR_FOR_DRAFT_PLANNED` env var.
- `budget_line_items` service applies the two in-scope changes directly when the flag is on, and creates no pending/approval record for them. All existing validation and completeness rules for Draft→Planned still apply and still block incomplete changes.
- Out-of-scope transitions (e.g. Planned → later status) and the "can't combine a status change with other edits" rule are unchanged.
- `/version` now surfaces `skip_cr_for_draft_planned` (read live from config each request) for verifying the effective per-environment setting.

**Frontend**
- Review/approval flow (`ReviewAgreement`, `CreateBLIsAndSCs`) updated so review copy and the in-review indicators reflect when a change is applied directly rather than sent for approval.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6094

## How to test

1. **Flag OFF (default / regression):** With `SKIP_CR_FOR_DRAFT_PLANNED` unset or `false`, change a Draft BLI to Planned and edit a Planned BLI's budget details — both should still create a Change Request requiring review/approval, exactly as today.
2. **Flag ON:** Set `SKIP_CR_FOR_DRAFT_PLANNED=true`, then:
   - Change a fully-qualified Draft BLI to Planned → applies immediately, no CR created, BLI does not show as In Review.
   - Edit budget details on a Planned BLI → applies immediately, no CR, no pending/in-review indicator anywhere (including reviewer queue).
3. **Validation still enforced:** With the flag ON, try to move an incomplete Draft BLI (missing funding source / amount / need-by date) to Planned → still blocked with the existing validation message.
4. **Scope check:** With the flag ON, attempt a Planned → later-status change → still requires review and approval.
5. **Combined-edit rule:** Confirm a status change + other budget-detail edits in one request is still rejected with the flag on or off.
6. Hit `GET /version` and confirm `skip_cr_for_draft_planned` reflects the environment's effective setting.

## A11y impact

## Storybook

## Screenshots

_N/A_

## Links

_N/A_


Santi-3rd:
Flag-OFF regression on the Draft→Planned review flow

  The button label and success alert are correctly gated on skipCrForDraftPlanned, but three sibling changes in
  ReviewAgreement.jsx are gated only on action === CHANGE_DRAFT_TO_PLANNED, not the flag. So they also apply in production, where the flag is OFF and a Draft→Planned still creates a Change Request for DD review:

  - "Review Changes" copy now reads "status changes you are making" (was "requesting approval for") — misleading when a CR is
  still created.
  - "Select Budget Lines" instructions no longer tell the user the agreement "will be sent to your Division Director to
  review and approve."
  - Notes accordion removed for all Draft→Planned actions (:325) — this is a real functional loss, not just copy:
    - The accordion's <TextArea> is the only caller of setNotes, so with it gone notes is permanently "" and requestor_notes
  is always empty.
    - Those notes are wired end-to-end into the CR (budget_line_items.py:914 → change_requests.py:370) and shown to the
  approver under "Submitter's Notes" (ApproveAgreement.jsx:190-198).

  This also contradicts the plan doc, which specifies the relabel be conditional and revert to the approval flow when OFF.

  Fix: gate all three on appliesImmediately (i.e. skipCrForDraftPlanned && isDraftToPlannedAction), matching the button label and success alert.

frontend/src/pages/agreements/review/ReviewAgreement.jsx
```diff
@@ -322,19 +322,22 @@ export const ReviewAgreement = () => {
...
-                    maxLength={150}
-                    value={notes}
-                    onChange={(name, value) => setNotes(value)}
-                />
-            </Accordion>
+            {action != actionOptions.CHANGE_DRAFT_TO_PLANNED && (
+                <Accordion
+                    heading="Notes"
+                    level={2}
+                >
```
josbell, line 330: **Bug: Notes accordion is hidden for Draft→Planned even when the flag is OFF**

The condition `action != actionOptions.CHANGE_DRAFT_TO_PLANNED` hides the Notes accordion unconditionally — it does not check `skipCrForDraftPlanned` or `appliesImmediately`. Neither of those values is referenced anywhere in this JSX file (they stay inside the hook and only drive `submitButtonText` and the success alert copy).

When `SKIP_CR_FOR_DRAFT_PLANNED` is **OFF**, a Draft→Planned submission still routes through the Change Request workflow and a Division Director reviews it. The backend (`change_requests.py:370`) stores and surfaces `requestor_notes` to that reviewer — but the submitter can never enter notes because this accordion is always hidden for this action.

**Suggested fix:** export `appliesImmediately` from the hook and gate on that instead of the raw action:
```diff
-{action != actionOptions.CHANGE_DRAFT_TO_PLANNED && (
+{!appliesImmediately && (
```

frontend/src/pages/agreements/review/ReviewAgreement.hooks.js
```diff
@@ -287,6 +299,16 @@ const useReviewAgreement = (agreementId) => {
```
josbell, line 301: **Bug: `handleActionChange` never resets `notes`; stale notes leak into the Draft→Planned payload**

`handleActionChange` (further down in this file) calls `setAction`, `setToggleStates`, and `setBudgetLines` — but never `setNotes('')`. The `handleSendToApproval` function unconditionally includes `requestor_notes: notes` for both action types.

Failure scenario: user selects **Planned→Executing**, types something in the visible Notes textarea, then switches to **Draft→Planned** (Notes hidden, state not cleared). On submit, every selected BLI is patched with the stale notes string as `requestor_notes` — notes the user never intended for this submission. No existing test covers the action-switch scenario.

**Suggested fix:** reset notes when the action changes:
```diff
 const handleActionChange = (action) => {
     setAction(action);
+    setNotes('');
     setToggleStates({});
```

josbell (commented):
Two confirmed correctness bugs found in the `ReviewAgreement` frontend flow (verified by tracing the local checkout). Both are in unchanged behaviour that this PR should have addressed.

josbell (approved):
## Review follow-up — issues to fix before merge

The core backend logic (flag gating, field restriction, audit history) looks solid and the backend unit/integration tests are thorough. Three issues still need to land before this merges:

---

### 1. Notes accordion still hidden flag-independently (not fixed from prior review)

`ReviewAgreement.jsx:325` — the condition is still `action != actionOptions.CHANGE_DRAFT_TO_PLANNED`, which hides Notes for all Draft→Planned submissions regardless of the flag. When the flag is **OFF**, a CR is still created and routes to a Division Director — the backend stores `requestor_notes` on the CR and the reviewer can see it, but the submitter can never enter notes.

`appliesImmediately` is computed inside the hook but not included in the hook's return object and never reaches the JSX.

**Fix:**
1. Add `appliesImmediately` to the hook's return object.
2. Destructure it in `ReviewAgreement.jsx`.
3. Change the gate: `{!appliesImmediately && (<Accordion heading="Notes">...)}`

---

### 2. `handleActionChange` still doesn't reset notes (not fixed from prior review)

`ReviewAgreement.hooks.js:351` — `handleActionChange` still has no `setNotes('')`. `handleSendToApproval` unconditionally sends `requestor_notes: notes` for both action types (line 263). If the user types notes on Planned→Executing then switches to Draft→Planned, those stale notes go into every BLI patch payload silently.

**Fix:** Add `setNotes('')` inside `handleActionChange`, and add a test covering the action-switch scenario.

---

### 3. Cypress tests assert stale h1 text — will fail once CI stack is healthy

`frontend/cypress/e2e/reviewAgreement.cy.js` — 11 assertions still check `cy.get("h1").contains("Request BL Status Change")` but the page now renders `"Change Budget Line Status"`. These will fail as soon as the E2E infrastructure issue is resolved (currently all ~50 E2E specs are failing with a uniform ~2min timeout, suggesting a Docker stack startup failure unrelated to this PR's changes).

**Fix:** Update the 11 occurrences in `reviewAgreement.cy.js` to match the new heading.

---

CI status: all unit tests, backend tests, linting, and security scans are green. The E2E failures appear to be a global infrastructure issue on this run (all 50 specs fail uniformly, including unrelated ones like `auth.cy.js` and `canList.cy.js`), not caused by this PR's changes — but the h1 assertions above will surface as real failures once that's resolved.

rajohnson90:
Josbell's identified issues have all been fixed

Santi-3rd (approved):

PR #6109, opened 2026-08-18
fix(agreements): validate BLI Obligate By date against SC PoP window

rajohnson90:
## What changed

Adds validation on the Review Agreement page to catch budget lines whose Obligate By date no longer falls within their services component's Period of Performance (PoP) — a state that can arise when a SC's PoP window is edited after a BLI's Obligate By date was already set. Also cleans up several existing notifications to consistently refer to "Period of Performance" instead of "agreement's start and end."

**`frontend/src/pages/agreements/review/suite.js`**
- Adds a new `budgetLineSuite` rule: for CONTRACT/IAA budget lines with both `date_needed` and a services component PoP range present, the Obligate By date must fall within `sc_period_start`/`sc_period_end` (inclusive). Skipped for GRANT BLIs (no SC) and when required data is missing.
- Exports `POP_RANGE_ERROR_KEY` so callers can special-case this rule's per-BL messages instead of deduping to a single generic message.

**`frontend/src/pages/agreements/review/ReviewAgreement.hooks.js`**
- Attaches each BLI's SC `period_start`/`period_end` onto the budget line before validation.
- Accumulates one message per violating BL under `POP_RANGE_ERROR_KEY` instead of deduping to the first.

**`frontend/src/pages/agreements/review/ReviewAgreement.jsx`**
- Renders one error-banner line per violating budget line for `POP_RANGE_ERROR_KEY`, instead of a single generic line.

**`frontend/src/components/BudgetLineItems/BudgetLinesTable/BLIRow.helpers.js`**
- Adds `isDateOutsidePopRange(budgetLine)` helper.

**`frontend/src/components/BudgetLineItems/BLIReviewTable/BLIReviewRow.jsx`**
- Applies error styling and a tooltip to the Obligate By date cell when the date falls outside the PoP range, while the row is in review mode.

**Wording cleanup** (`datePickerSuite.js`, `BLIReviewRow.jsx`, `ServicesComponents.hooks.js`, and their tests)
- Replaced "agreement's start and end" phrasing with "Period of Performance" in the Budget Lines form date validation message, the BLI review table tooltip, and the Services Component edit confirmation modal, for consistent terminology across notifications.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5772

## How to test

Automated:
```bash
cd frontend
bun run test --watch=false -- suite.test.js ReviewAgreement datePickerSuite BudgetLinesForm BLIReviewRow ServicesComponents
```

Manual:
[...]

PR #6127, opened 2026-08-22
fix(agreements): resolve UX feedback for Draft→Planned change request skip

weimiao67:
## What changed

Resolves UX feedback items on the "Skip Change Request review for Draft→Planned and in-Planned budget edits" work (#6094):

- Clarify the disabled **Change BL Status** button tooltip for non-team members — now reads "Only team members listed on this agreement can change a BL status" (previously "...can send to approval").
- Remove the send-to-review icon from agreements list rows.
- Skip the procurement-shop change request under `SKIP_CR_FOR_DRAFT_PLANNED`.
- Add Obligate By date validation against the SC PoP window (with tests).
- Update history title for budget line status changes.
- Resolve SC link before the dirty check so a newly-assigned SC saves.
- Add top margin to the Product Service Code field.
- Flag unassociated BLs on Edit/Review Agreement with an error border and message.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6094

## How to test

1. Open an agreement's Budget Lines tab as a user who is **not** a team member on that agreement.
2. Hover over the disabled **Change BL Status** button.
3. Confirm the tooltip reads: "Only team members listed on this agreement can change a BL status".
4. Verify the other UX changes above against their respective flows (agreements list icon, SC PoP validation, history title, SC link save, unassociated BL error borders).

## A11y impact

## Storybook

## Screenshots

_N/A_

## Links

_N/A_

### DIFF
--- a/backend/models/agreement_history.py
+++ b/backend/models/agreement_history.py
@@ -162,8 +162,12 @@ def agreement_history_trigger_func(event: OpsEvent, session: Session, system_use
                 )
             )
         case OpsEventType.UPDATE_AGREEMENT:
-            # Handle CAN Updates
-            change_dict = event.event_details["agreement_updates"]["changes"]
+            # `agreement_updates` is absent when a publisher of UPDATE_AGREEMENT does not
+            # compute a field diff (e.g. an edit-bundle save that changes no diffed agreement
+            # field). Default to empty so this handler records nothing rather than KeyError-ing
+            # (the error was previously swallowed by the subscriber, silently losing all history).
+            agreement_updates = event.event_details.get("agreement_updates") or {}
+            change_dict = agreement_updates.get("changes", {})
             for key in change_dict.keys():
                 history_item = create_agreement_update_history_event(
                     key,
@@ -171,22 +175,22 @@ def agreement_history_trigger_func(event: OpsEvent, session: Session, system_use
                     change_dict[key]["new_value"],
                     event_user,
                     event.created_on.strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
-                    event.event_details["agreement_updates"]["owner_id"],
+                    agreement_updates["owner_id"],
                     event.id,
                     session,
                     system_user,
                 )
                 if history_item:
                     history_events.append(history_item)
-            if event.event_details["agreement_updates"].get("team_member_changes"):
+            if agreement_updates.get("team_member_changes"):
                 # Handle team member changes
-                team_member_changes = event.event_details["agreement_updates"]["team_member_changes"]
+                team_member_changes = agreement_updates["team_member_changes"]
                 for item in team_member_changes.get("user_ids_added", []):
                     added_user_id = session.get(User, item)
                     history_events.append(
                         AgreementHistory(
-                            agreement_id=event.event_details["agreement_updates"]["owner_id"],
-                            agreement_id_record=event.event_details["agreement_updates"]["owner_id"],
+                            agreement_id=agreement_updates["owner_id"],
+                            agreement_id_record=agreement_updates["owner_id"],
                             ops_event_id=event.id,
                             history_title="Change to Team Members",
                             history_message=f"{event_user.full_name} added team member {added_user_id.full_name}.",
@@ -198,8 +202,8 @@ def agreement_history_trigger_func(event: OpsEvent, session: Session, system_use
                     removed_user_id = session.get(User, item)
                     history_events.append(
                         AgreementHistory(
-                            agreement_id=event.event_details["agreement_updates"]["owner_id"],
-                            agreement_id_record=event.event_details["agreement_updates"]["owner_id"],
+                            agreement_id=agreement_updates["owner_id"],
+                            agreement_id_record=agreement_updates["owner_id"],
                             ops_event_id=event.id,
                             history_title="Change to Team Members",
                             history_message=f"{event_user.full_name} removed team member {removed_user_id.full_name}.",
@@ -207,15 +211,15 @@ def agreement_history_trigger_func(event: OpsEvent, session: Session, system_use
                             history_type=AgreementHistoryType.AGREEMENT_UPDATED,
                         )
                     )
-            if event.event_details["agreement_updates"].get("special_topic_changes"):
+            if agreement_updates.get("special_topic_changes"):
                 # Handle special topic changes
-                special_topic_changes = event.event_details["agreement_updates"]["special_topic_changes"]
+                special_topic_changes = agreement_updates["special_topic_changes"]
                 for item in special_topic_changes.get("special_topics_ids_added", []):
                     added_special_topic = session.get(SpecialTopic, item)
                     history_events.append(
                         AgreementHistory(
-                            agreement_id=event.event_details["agreement_updates"]["owner_id"],
-                            agreement_id_record=event.event_details["agreement_updates"]["owner_id"],
+                            agreement_id=agreement_updates["owner_id"],
+                            agreement_id_record=agreement_updates["owner_id"],
                             ops_event_id=event.id,
                             history_title="Change to Special Topic/Population Studied",
                             history_message=f"{event_user.full_name} added Special Topic/Population Studied {added_special_topic.name}.",
@@ -227,8 +231,8 @@ def agreement_history_trigger_func(event: OpsEvent, session: Session, system_use
                     removed_special_topic = session.get(SpecialTopic, item)
                     history_events.append(
                         AgreementHistory(
-                            agreement_id=event.event_details["agreement_updates"]["owner_id"],
-                            agreement_id_record=event.event_details["agreement_updates"]["owner_id"],
+                            agreement_id=agreement_updates["owner_id"],
+                            agreement_id_record=agreement_updates["owner_id"],
                             ops_event_id=event.id,
                             history_title="Change to Special Topic/Population Studied",
                             history_message=f"{event_user.full_name} removed Special Topic/Population Studied {removed_special_topic.name}.",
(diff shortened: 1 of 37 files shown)


Santi-3rd (approved):

Santi-3rd:
There's this one finding that might be worth looking into: 

--

PoP-range error banner shows an incomplete, confusing message

  In frontend/src/pages/agreements/review/suite.js:41, the PoP-range test's failure message is just:

  test("Budget Line Obligate By Date must be within PoP", `Budget Line ${budgetLine.id} Obligate By date`, () => { ... })

  ReviewAgreement.jsx renders this message string verbatim in the error banner (it bypasses convertCodeForDisplay for POP_RANGE_ERROR_KEY). So when a BL's Obligate By date falls outside the PoP, the user sees:

  ▎ Budget Line 42 Obligate By date

  — a sentence fragment that never states what's wrong. This is confirmed by suite.test.js (...toBe("Budget Line 42 Obligate By date")). The row-level tooltip in BLIRow.jsx has the full text ("Obligate By date is outside the agreement's Period of Performance"), so the banner is the odd one out — notable given this PR is specifically UX-polish.

  Suggest making the message a complete sentence, e.g. `Budget Line ${budgetLine.id} Obligate By date is outside the Period of Performance` (and updating the two suite.test.js assertions). Note the ReviewAgreement.test.jsx cases mock pageErrors with hand-written messages, so they don't catch this — the real message only surfaces through the suite.

rajohnson90:
Guillermo your review comments look suspiciously like they were meant for my BLI and SC validation PR.

josbell (changes requested):
## Review: Request Changes

Four issues found after deep-dive analysis. Two are correctness bugs that should block merge.

---

### 🔴 Bug 1 — Wrong `updated_by` recorded in procurement-shop history (stale value)

**File:** `backend/ops_api/ops/services/agreement_edit_bundle.py` line ~219

`handle_create_update_by_attrs` is registered only on `before_commit`, not `before_flush`. The `commit=False` path calls `session.flush()`, which does **not** trigger that hook. When line 219 reads `agreement.updated_by` to pass to `generate_agreement_events_update`, the value is stale — either the previous editor's ID or `None` — not the current request user.

Result: the "Change to Procurement Shop" history entry is stamped with the wrong actor.

**Fix:** Replace `agreement.updated_by` with `get_current_user().id`, the same pattern used by the standard PATCH endpoint in `agreements.py:417`:
```python
from ops_api.ops.utils.core import get_current_user
current_user_obj = get_current_user()
full_updates = generate_agreement_events_update(
    old_serialized,
    agreement.to_dict(),
    agreement.id,
    current_user_obj.id if current_user_obj else None,
)
```

---

### 🔴 Bug 2 — Pre-existing `NameError` crash in `create_services_component_history_event`

**File:** `backend/models/agreement_history.py` lines 1047/1049

When `period_start` or `period_end` is `None`/empty, the `else` branch is skipped and `old_date`/`new_date` are never assigned. The f-strings on lines 1047 and 1049 reference them unconditionally → `NameError` crash, silently losing all history for that event.

This is a pre-existing bug, but the PR's fix to the `UPDATE_AGREEMENT` handler now routes more events through this subscriber (previously errors were swallowed), making it reachable in practice.

The sibling `create_grant_number_history_event` (lines 1095–1112) handles this correctly with a ternary that reassigns back to `old_value`/`new_value`. Apply the same pattern here:
```python
old_value = (
    "None" if old_value is None or old_value == ""
    else datetime.strftime(datetime.strptime(old_value, "%Y-%m-%d"), "%m/%d/%Y")
)
new_value = (
    "None" if new_value is None or new_value == ""
    else datetime.strftime(datetime.strptime(new_value, "%Y-%m-%d"), "%m/%d/%Y")
)
# f-strings reference {old_value} and {new_value} — always assigned
```

---

### 🟡 Issue 3 — `PLANNED_MOD` in SKIP_CR `any()` is dead code

**File:** `backend/ops_api/ops/services/agreements.py` line ~630

The blocked-status guard at lines 604-612 already raises `ValidationError` for `PLANNED_MOD`, so it can never reach the `any()` check at line 630. The comment at line 621 correctly documents this, but the code still includes `BudgetLineItemStatus.PLANNED_MOD` in the tuple.

This is dead code today, but if `PLANNED_MOD` is ever removed from `_blocked_statuses`, the SKIP_CR branch silently inherits it with no test to catch the regression.

**Fix:** Remove `BudgetLineItemStatus.PLANNED_MOD` from the `any()` at line 630, leaving only `BudgetLineItemStatus.PLANNED`.

---

### 🟡 Issue 4 — Missing `agreement_type !== "GRANT"` guard in `CreateBLIsAndSCs/suite.js`

**File:** `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/suite.js` line ~32

`pages/agreements/review/suite.js` guards the PoP-range rule with an explicit `budgetLine.agreement?.agreement_type !== "GRANT"` check. `CreateBLIsAndSCs/suite.js` does not — it relies implicitly on GRANT BLIs having `sc_period_start=null` from the data pipeline. Currently functionally equivalent, but if any future code path attaches SC period data to GRANT BLIs, the edit-mode suite would silently validate them while the review-mode suite correctly skips them.

**Fix:** Add the explicit type guard to match `review/suite.js`:
```js
if (item.agreement?.agreement_type !== "GRANT" && item.date_needed && item.sc_period_start && item.sc_period_end) {
```

rajohnson90 (approved):
Looks mostly good on my end. A few small things you can fix at the same time while working on Josbell's issues:

1. suite.js:107 — leftover template literal with no interpolation; now a plain string, backticks are misleading.
  2. ReviewAgreement.jsx:144 — the per-BL messages array in pageErrors is no longer read anywhere in the render path, but suite.js/both hooks
  files still build it and three test files assert on that now-dead text — worth trimming so tests don't mask future regressions in the
  actually-rendered label.

weimiao67:
@josbell Thanks for the deep-dive review. All four items are addressed on this branch. Two of them diverged from the suggested fix after tracing the code — details below.

### 🔴 Bug 1 — `updated_by` in edit-bundle proc-shop history diff
Fixed in `8940c19` — now passes the current request user instead of `agreement.updated_by`, which is stale on the `commit=False` path (the created/updated-by stamping fires on `before_commit`, and this path only flushes).

Two notes on the diagnosis:
- The suggested import path doesn't exist — `get_current_user` comes from `flask_jwt_extended` (as in `agreements.py`), not `ops_api.ops.utils.core`. Used the correct one.
- The user-visible history **actor** is actually derived from `OpsEvent.created_by` (`events.py` sets it to `current_user.id`) via `event_user = session.get(User, event.created_by)`, and `agreement_updates["updated_by"]` is never read for it. So there was no wrong-actor symptom in practice — the stale value only lived in the `event_details` audit blob. Fixed anyway as correctness hygiene.

### 🔴 Bug 2 — `NameError` in `create_services_component_history_event`
Fixed in `95a61b3` — mirrored the grant sibling's ternary so `old_value`/`new_value` are always assigned, plus a regression test (`test_services_component_period_change_with_none_value`) that reproduces the crash for empty/None period values.

Confirmed genuinely pre-existing, but the reachability note in the review is off: this runs on `UPDATE_SERVICES_COMPONENT`, not the `UPDATE_AGREEMENT` handler this PR changed, and the subscriber's `try/except` in `agreement_messages.py` still swallows the exception — so it silently drops history rather than surfacing a 500. Worth fixing, but not a merge blocker for this PR.

### 🟡 Issue 3 — `PLANNED_MOD` dead code
Fixed in `9ac45e8` — the `any(...)` now checks only `PLANNED`, since the blocked-status guard raises first for `PLANNED_MOD`.

### 🟡 Issue 4 — grant guard in `CreateBLIsAndSCs/suite.js`
Fixed in `7e40b336`, but **not** with the literal per-item guard suggested. In this editor, locally-created BLIs carry only a partial agreement stub (`{ procurement_shop }`) with no `agreement_type`, so `item.agreement?.agreement_type !== "GRANT"` would be `undefined !== "GRANT"` → always true — a no-op. Instead threaded the reliable `agreement_type` into the suite data and guard on `data.agreement_type !== "GRANT"`, with a test that fails without the guard.

weimiao67:
@rajohnson90 Thanks for the approval! Both of these were resolved by reverting `aca96a6` (the "complete sentence" PoP-banner change) — they were artifacts of that commit's state:

1. **suite.js:107** — after the revert this is back to `` `Budget Line ${budgetLine.id} Obligate By date` ``, which interpolates the BL id, so the backticks are correct now (not a plain string).
2. **ReviewAgreement.jsx per-BL messages** — the array is live end-to-end: built in `ReviewAgreement.hooks.js` (accumulates one message per violating BL), rendered in `ReviewAgreement.jsx` via the `key === POP_RANGE_ERROR_KEY` branch (`messages.map` → one `<li>` each, added in #6109), and covered by `ReviewAgreement.test.jsx` which asserts one `<li>` per message. So it's not dead — nothing to trim, and the test assertions match the current rendered text.

No changes needed on these two.

rajohnson90:
Appreciate the changes Wei. The text has been improved in a different direction after talking with Nate. The biggest issue before was that there was an error line for each BLI which was inconsistent with the rest of the validation in the error banner. Now it just simply states one time whether the Obligate By Date is an issue. Other fields listed are just listed by name so this is consistent with the rest of the banner.

josbell:
One gap to address before or after merge:

**`CreateBLIsAndSCs.jsx` — grant agreement unassociated bucket missing error treatment (line ~563)**

The PR adds `isUnassociatedError` + red border + "This is required information" to the `ServicesComponentAccordion` path when BLIs sit in the number-0 bucket in review mode. The parallel `GrantNumberAccordion` path (lines 562–578) has no equivalent — `groupByGrantNumber` routes unlinked grant BLIs into a `grantNumberNumber === 0` bucket the same way, so a grant agreement with BLIs not assigned to any grant number silently skips the error UI.

Fix mirrors the SC pattern:
```js
const isUnassociatedError =
    isReviewMode && group.grantNumberNumber === 0 && group.budgetLines.length > 0;
```
...then pass `isError={isUnassociatedError}` to `GrantNumberAccordion` and render the error message div above it.

Note: `ReviewAgreement.jsx` lines 210/216 already handle this correctly for the status-change flow — gap is only in the `CreateBLIsAndSCs` wizard/edit path.

josbell (approved):
LGTM — the four prior-round findings are all addressed. One small gap noted in a separate comment (grant unassociated bucket error treatment in the CreateBLIsAndSCs wizard).

PR #6132, opened 2026-08-24
fix: show a single PoP-range error for the Obligate By date regardless of BLI count

rajohnson90:
## What changed

Fixes the "Obligate By Date" validation error in the BLI status-change review flow (Review Agreement / Request Pre-Award Approval pages) so that any nonzero number of budget lines violating the "Obligate By within the services component's Period of Performance" rule surfaces as **one** line in the error banner, instead of one duplicated line per violating budget line.

**`frontend/src/pages/agreements/review/suite.js`**
- Simplified the PoP-range Vest test message to a generic string (no longer embeds the budget line id).
- Updated the `POP_RANGE_ERROR_KEY` comment: it now dedupes like every other BLI validation key instead of accumulating one message per violating BLI.

**`frontend/src/pages/agreements/review/ReviewAgreement.jsx`**
- Removed the special-case rendering branch that mapped one `<li>` per PoP-range message. The key now renders identically to every other error key via `convertCodeForDisplay`.

**`frontend/src/pages/agreements/review/ReviewAgreement.hooks.js`** and **`frontend/src/pages/agreements/pre-award-approval/RequestPreAwardApproval.hooks.js`**
- Removed the accumulate-instead-of-dedupe branch (and the now-unneeded ascending-BLI-id sort) for `POP_RANGE_ERROR_KEY`. It now dedupes to a single message across all selected budget lines, matching every other error key.

**`frontend/src/helpers/utils.js`**
- Updated a stale comment describing the old per-BLI bypass behavior.

**Tests**
- Updated `suite.test.js`, `ReviewAgreement.hooks.test.js`, and `ReviewAgreement.test.jsx` to assert the new single-message behavior (previously they explicitly asserted the old per-BLI duplicated-message behavior).

## Issue

https://github.com/HHS/OPRE-OPS/issues/5772

## How to test

1. `cd frontend && bun run test --watch=false -- src/pages/agreements/review/suite.test.js src/pages/agreements/review/ReviewAgreement.hooks.test.js src/pages/agreements/review/ReviewAgreement.test.jsx` — all pass.
2. Manually: open the "Change Budget Line Status" (Review Agreement) page for an agreement with 2+ budget lines whose Obligate By date falls outside their services component's period of performance. Select those budget lines and attempt to change their status. Confirm the error banner shows exactly one "Obligate By Date Outside Period of Performance" line, not one per budget line.

## A11y impact

## Storybook

## Screenshots

N/A — the visible change is that a previously duplicated error line now appears once; no new UI.

## Links

N/A

### DIFF
--- a/frontend/src/helpers/utils.js
+++ b/frontend/src/helpers/utils.js
@@ -152,9 +152,8 @@ export const codesToDisplayText = {
         services_component: "Services Component",
         grant_number: "Grant Number",
         date_needed: "Obligate By Date",
-        // POP_RANGE_ERROR_KEY (see suite.js) — ReviewAgreement.jsx bypasses this and renders
-        // the per-BL messages directly, but keep a label here so the key isn't undefined if
-        // ever displayed generically (e.g. a future caller, or a page without the bypass).
+        // POP_RANGE_ERROR_KEY (see suite.js) — shown as a single generic line regardless
+        // of how many budget lines violate the PoP window.
         date_needed_pop_range: "Obligate By Date Outside Period of Performance"
     },
     classNameLabels: {
--- a/frontend/src/pages/agreements/pre-award-approval/RequestPreAwardApproval.hooks.js
+++ b/frontend/src/pages/agreements/pre-award-approval/RequestPreAwardApproval.hooks.js
@@ -17,7 +17,7 @@ import {
 } from "../../../components/Agreements/Documents/Document";
 import { PROCUREMENT_STEP_STATUS } from "../../../components/Agreements/ProcurementTracker/ProcurementTracker.constants";
 import usePreAwardApprovalData from "./usePreAwardApprovalData";
-import agreementSuite, { POP_RANGE_ERROR_KEY, validateBudgetLineItems } from "./suite";
+import agreementSuite, { validateBudgetLineItems } from "./suite";
 import { VALIDATABLE_BLI_STATUSES } from "./constants";
 
 /**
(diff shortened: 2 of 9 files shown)


weimiao67 (approved):

josbell (approved):
The core fix is correct and well-scoped — removing the per-BLI accumulation branch and making `POP_RANGE_ERROR_KEY` behave like every other error key is the right architectural call. Data flow confirmed end-to-end: suite → hooks → render → `convertCodeForDisplay` all consistent.

Two test gaps worth addressing (not blocking, but easy wins):

**1. `RequestPreAwardApproval.hooks.test.js` — no PoP-range coverage**
No BLI fixture in this file sets `sc_period_start`/`sc_period_end`, so the `date_needed_pop_range` path never fires. The dedup change was applied to this hook too, but a regression reintroducing accumulation here would go undetected. Suggest adding a case with 2+ BLIs outside their PoP window asserting `pageErrors[POP_RANGE_ERROR_KEY]` has exactly one entry.

**2. `ReviewAgreement.test.jsx` line 92 — test description vs. intent mismatch**
The test feeds a 3-element messages array directly through the mocked hook and asserts `toHaveLength(1)`. That assertion passes because the render loop iterates *keys*, not message arrays — the array length is irrelevant. The test description ("no matter how many BLs violate") claims to verify dedup, but it only verifies JSX key-iteration. The actual dedup contract is correctly covered in `ReviewAgreement.hooks.test.js` line 264. Suggest either updating the description to reflect what's actually tested, or swapping the 3-element fixture for a single-element array to remove the misleading implication.

Minor: the `const errorKey = `${fieldName}`` on `ReviewAgreement.hooks.js` line 238 is a no-op identity coercion (`Object.entries` always yields strings) — `RequestPreAwardApproval.hooks.js` uses `fieldName` directly. The two aggregation loops are now functionally identical; a follow-up extraction to a shared helper would eliminate the drift risk the `POP_RANGE_ERROR_KEY` history already demonstrated.

PR #6137, opened 2026-08-25
feat: notify Slack on deploy job failure

jonnalley:
## What changed

Adds a Slack notification whenever a Dev/Staging deploy job fails, so failures surface immediately instead of going unnoticed. 

**New: `.github/actions/notify-slack-failure/action.yaml`**
A composite action that POSTs deploy-failure data to a Slack webhook via `curl`. It sends structured fields — `workflow_name`, `job_name`, `run_url`, `run_number`, `commit_sha` — rather than one pre-formatted text string, so the receiving Slack Workflow Builder automation (and whoever owns that channel) controls the message layout without needing a follow-up PR here to change formatting.

**Workflow edits** — wires the new action into the last step of every deploy job, gated on `if: failure()`, in all four Dev/Staging workflows:
- `dev_fe_build_and_deploy.yml` — `deploy-frontend`
- `dev_be_build_and_deploy.yml` — `deploy-data-tools`, `deploy-backend`
- `stg_fe_build_and_deploy.yml` — `deploy-frontend`
- `stg_be_build_and_deploy.yml` — `deploy-data-tools`, `deploy-backend`

No new `permissions:` needed anywhere — posting to a webhook needs no repo-token scopes.

**Design notes for reviewers**
- **No dedupe.** Each failure posts its own message rather than updating a prior one. Accepted tradeoff for a plain webhook instead of a Slack bot/app token (much simpler setup, no OAuth scopes to manage) — this exact incident (3+ consecutive failures) would produce 3+ separate messages under this design, not one updated thread.
- The `SLACK_WEBHOOK_URL` secret was added by the repo owner, pointing at a Slack Workflow Builder webhook trigger with 5 matching Text variables (`workflow_name`, `job_name`, `run_url`, `run_number`, `commit_sha`).

## Issue

No ticket — incident-driven, not part of pre-existing planned work. See [HHS/OPRE-OPS-Data#58](https://github.com/HHS/OPRE-OPS-Data/pull/58) for the root-cause fix this complements.

## How to test

1. On a throwaway branch, force a deploy job to fail (e.g. temporarily point `creds`/`webhook_url` at a bad value, or break a preceding step) and push, or trigger via `gh workflow run <name>`.
2. Confirm the `Notify Slack on failure` step actually runs (not skipped) and a message lands in the configured Slack channel with all 5 fields populated correctly.
3. Confirm a *successful* run shows the notify step as **skipped** (not absent) — `if: failure()` steps still appear in the job summary, just marked skipped, which is expected and not a bug.
4. Revert the throwaway breakage before merging that test branch.

No automated tests apply here — this is CI/CD configuration, not application code.

## A11y impact

## Storybook

## Screenshots

N/A — backend/CI configuration change only, no UI.

Most items don't apply to a CI-config-only change (no application code, no tests to add, no coverage/load/a11y/security surface):

## Links

Companion PR: [HHS/OPRE-OPS-Data#58](https://github.com/HHS/OPRE-OPS-Data/pull/58)

### DIFF
--- a/.github/actions/notify-slack-failure/action.yaml
+++ b/.github/actions/notify-slack-failure/action.yaml
@@ -1,10 +1,10 @@
 name: "notify-slack-failure"
 author: "@devops-chris"
-description: "Posts a Slack message via incoming webhook when a deploy job fails"
+description: "Posts deploy-failure data to a Slack Workflow Builder webhook"
 
 inputs:
   webhook_url:
-    description: "Slack incoming webhook URL"
+    description: "Slack Workflow Builder webhook URL"
     required: true
 
 runs:
@@ -20,8 +20,12 @@ runs:
         RUN_NUMBER: ${{ github.run_number }}
         COMMIT_SHA: ${{ github.sha }}
       run: |
-        MESSAGE=":red_circle: *${WORKFLOW_NAME}* failed — job \`${JOB_NAME}\` on <${RUN_URL}|run #${RUN_NUMBER}> (commit ${COMMIT_SHA})"
-
         curl -sf -X POST -H 'Content-type: application/json' \
-          --data "$(jq -n --arg text "$MESSAGE" '{text: $text}')" \
+          --data "$(jq -n \
+            --arg workflow_name "$WORKFLOW_NAME" \
+            --arg job_name "$JOB_NAME" \
+            --arg run_url "$RUN_URL" \
+            --arg run_number "$RUN_NUMBER" \
+            --arg commit_sha "$COMMIT_SHA" \
+            '{workflow_name: $workflow_name, job_name: $job_name, run_url: $run_url, run_number: $run_number, commit_sha: $commit_sha}')" \
           "$WEBHOOK_URL"
--- a/.github/actions/notify-slack-failure/action.yaml
+++ b/.github/actions/notify-slack-failure/action.yaml
@@ -1,0 +1,27 @@
+name: "notify-slack-failure"
+author: "@devops-chris"
+description: "Posts a Slack message via incoming webhook when a deploy job fails"
+
+inputs:
+  webhook_url:
+    description: "Slack incoming webhook URL"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Post failure notification to Slack
+      shell: bash
+      env:
+        WEBHOOK_URL: ${{ inputs.webhook_url }}
+        WORKFLOW_NAME: ${{ github.workflow }}
+        JOB_NAME: ${{ github.job }}
+        RUN_URL: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        RUN_NUMBER: ${{ github.run_number }}
+        COMMIT_SHA: ${{ github.sha }}
+      run: |
+        MESSAGE=":red_circle: *${WORKFLOW_NAME}* failed — job \`${JOB_NAME}\` on <${RUN_URL}|run #${RUN_NUMBER}> (commit ${COMMIT_SHA})"
+
+        curl -sf -X POST -H 'Content-type: application/json' \
+          --data "$(jq -n --arg text "$MESSAGE" '{text: $text}')" \
+          "$WEBHOOK_URL"
--- a/.github/actions/notify-slack-failure/action.yaml
+++ b/.github/actions/notify-slack-failure/action.yaml
@@ -1,5 +1,4 @@
 name: "notify-slack-failure"
-author: "@devops-chris"
 description: "Posts deploy-failure data to a Slack Workflow Builder webhook"
 
 inputs:
--- a/.github/workflows/dev_be_build_and_deploy.yml
+++ b/.github/workflows/dev_be_build_and_deploy.yml
@@ -72,6 +72,12 @@ jobs:
             # Start the data-tools job
             az containerapp job start -n opre-ops-${{ env.ENVIRONMENT }}-app-data-tools -g opre-ops-${{ env.ENVIRONMENT }}-app-rg
 
+      - name: Notify Slack on failure
+        if: failure()
+        uses: ./.github/actions/notify-slack-failure
+        with:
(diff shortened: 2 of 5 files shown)


Santi-3rd:
Everything went well except for this finding. The Slack notification came up empty. Might be worth investigating

--

Step 2 partially passes — the notification fires, but the fields don't populate. This is a Slack Workflow Builder configuration issue, not a code bug. Whoever owns that Slack automation needs to open the workflow in Slack, go to the message step, and replace the empty field values with the corresponding webhook variable references.

weimiao67 (approved):

josbell (changes requested):
## Review findings

Good idea and clean implementation. Five issues to address before merging:

---

### 1. Build job failures produce no Slack notification (confirmed)

The notify steps are only wired into deploy jobs. When a build job fails, GitHub Actions places all downstream `needs:` jobs into **skipped** state — no steps execute inside a skipped job, so the notify step never fires.

Failure chain: `build-backend` fails → `deploy-backend` **skipped** → notify step never runs → zero Slack message.

Fix: add the same notify step to `build-data-tools`, `build-backend`, and `build-frontend`.

---

### 2. `if: failure()` misses manual cancellations (confirmed — all 5 call sites)

`failure()` only returns true when a prior step exited non-zero. A manually-cancelled job has status `cancelled`, not `failure`, so `failure()` returns false and the step is silently skipped.

Fix: change all 5 instances to:
```yaml
if: failure() || cancelled()
```

---

### 3. `curl -sf` adds a spurious second failure when the Slack webhook errors (confirmed)

`-f` / `--fail` causes curl to exit 22 on HTTP 4xx/5xx. There is no `continue-on-error: true` anywhere. If Slack is down, the URL is wrong, or the secret is unset, the notify step itself exits non-zero — GitHub Actions marks it as a **second failed step**, adding a confusing extra red-X annotation on top of the real deploy failure.

Fix: add `continue-on-error: true` to each calling step (preferred — keeps the composite action clean), or drop `-f` from the curl invocation.

---

### 4. Notify step placed before a commented-out deploy block in `deploy-data-tools` (dev_be) (confirmed)

In `dev_be_build_and_deploy.yml`, the notify step is inserted before the commented-out `azure/container-apps-deploy-action` block (lines 75–81). The other job (`deploy-backend`) correctly places notify at the absolute end. When that commented block is eventually uncommented, failures from the newly-active deploy step will occur after the notify step has already been evaluated — and won't trigger a notification.

Fix: move the notify step to after the commented block so it remains last regardless of what gets uncommented.

---

### 5. `webhook_url` input should be marked `secret: true` (plausible)

GitHub's value-based masking protects the URL in normal logs since callers pass `secrets.SLACK_WEBHOOK_URL`. However, without `secret: true`, there is a gap: with `ACTIONS_STEP_DEBUG=true` enabled, env var assignment trace lines can be emitted before the masking filter applies. It also protects future callers who don't source the value from `secrets.*`.

Fix: add `secret: true` to the input declaration:
```yaml
inputs:
  webhook_url:
    description: "Slack Workflow Builder webhook URL"
    required: true
    secret: true
```

jonnalley:
@Santi-3rd  Turned out to be a stale-publish issue on the Slack Workflow Builder side. Verified fixed now and tagged you in the slack message. 

> Everything went well except for this finding. The Slack notification came up empty. Might be worth investigating
> 
> --
> 
> Step 2 partially passes — the notification fires, but the fields don't populate. This is a Slack Workflow Builder configuration issue, not a code bug. Whoever owns that Slack automation needs to open the workflow in Slack, go to the message step, and replace the empty field values with the corresponding webhook variable references.

jonnalley:
@josbell 
## Response to review

  Went through each finding against the actual code and GitHub's docs before responding, rather than taking them at face value. Fixes for 1–4 are in `1993f1427`.

  ---

  ### 1. Build job failures produce no Slack notification (confirmed → fixed)

  Verified against GitHub's docs on `needs:` — a failed build job causes its dependent deploy job to be skipped entirely, so the notify step (previously only in deploy jobs)
  never ran for build failures.

  Fixed: added the notify step to `build-data-tools`, `build-backend`, and `build-frontend` in all four workflows.

  ---

  ### 2. `if: failure()` misses manual cancellations (confirmed → fixed)
  
  `failure()` and `cancelled()` are documented as distinct, non-overlapping conditions.

  Fixed: all 12 notify call-sites now use:
  ```yaml
  if: failure() || cancelled()
  ```

  ---
  
  ### 3. `curl -sf` adds a spurious second failure when the Slack webhook errors (confirmed → fixed)
  
  `curl -sf` exits non-zero on an HTTP error, which would fail the notify step itself and stack a second red-X on top of the real failure it's reporting on.

  Fixed: added `continue-on-error: true` to every calling step — kept out of the composite action itself, per your suggestion.

  ---

  ### 4. Notify step placed before a commented-out deploy block in `deploy-data-tools` (dev_be) (confirmed → fixed)

  Verified directly against the file — `deploy-data-tools` in `dev_be_build_and_deploy.yml` had its notify step before the commented-out deploy block, unlike `deploy-backend`
  where it was correctly last.

  Fixed: moved it so it stays last regardless of what gets uncommented later.

  ---

  ### 5. `webhook_url` input should be marked `secret: true` (pushing back)
  
  Checked GitHub's [Actions metadata syntax docs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions) directly — `inputs.<id>` only
  supports `description`, `required`, `default`, and `deprecationMessage`. There's no `secret` key for composite action inputs, so `secret: true` wouldn't have any effect
  here.

  The protection you're after already exists independent of that: since every caller passes `${{ secrets.SLACK_WEBHOOK_URL }}`, GitHub registers that value for masking the
  moment it's read, regardless of which input it flows through — masking is value-based, not declaration-based. I didn't find documentation of `ACTIONS_STEP_DEBUG` bypassing
  that.

  If you have a specific case in mind where it doesn't hold, let's discuss

Santi-3rd (approved):

jonnalley:
@josbell can you please see my comment above and request for re-review

josbell (approved):

Issue #6013, opened 2026-07-29
Edit a Grant


rajohnson90:
### User Story
As an OPRE staff member, in order to keep a Grant's information accurate as circumstances change after its creation, I want to update an existing Grant's details, grant numbers, and budget lines.

### Acceptance Criteria
  - [x] An user can update a Grant's attributes, including attributes unique to a grant, and see any changes successfully persisted
  - [x] User can add, update, or remove grant numbers on an existing Grant, and the change is reflected everywhere grant numbers are displayed
  - [x] User can add, update, or remove budget lines on an existing Grant, and the change is reflected in the Grant's totals and budget line displays
  - [x] The changes a user is permitted to make are constrained by the Grant's current lifecycle stage, consistent with how editability is constrained for other agreement types
  - [x] User receives a clear indication that a change was saved successfully, or a clear explanation when a change could not be saved
  - [x] User can remove a grant number that still has budget lines and the blis will be disassociated with grant automatically on save, exactly how service component deletion works. Header of no-grant group will be "Budget lines not associated with a grant number"
  - [x] Only users authorized to edit the Grant can make these changes; unauthorized users cannot
  - [x] Users will be prompted with modals if they decide to navigate away (different modals depending on wether approvals are required or not as seen in figma)
  - [x] UI is updated to reflect figma design

  **Accessibility considerations:**
  - [x] All editing capability is reachable and operable via keyboard navigation
-----------------------------------------

#### Manual Test Cases
 **Positive cases:**
  - Edit an existing Grant's metadata (e.g., FOA (NOFO), total funding, nickname) and confirm the changes are persisted and retrievable
  - Add, edit, and delete a grant number on an existing Grant and confirm each change is reflected in grant number displays
  - Add, edit, and delete a budget line on an existing Grant and confirm the Grant's totals and budget line list update accordingly
  - Confirm edit history reflects who changed what and when for each of the above

  **Negative / edge cases:**
  - Attempting to save an edit with a required field cleared shows a validation error and does not persist the change
  - Attempting to delete a grant number that still has budget lines attached is prevented or clearly disallowed
  - A user without edit authorization on the Grant cannot edit its metadata, grant numbers, or budget lines (role/authorization check)
  - Attempting to edit a Grant (or its budget lines) while in a lifecycle stage that other agreement types treat as locked (e.g., in review, post-award-locked) is prevented
  consistent with existing rules

  #### Automated Test Coverage
  
[...]

PR #6160, opened 2026-08-26
fix: editing tag alignment and post-save redirect for budget lines tab

Santi-3rd:
## What changed

Three related UI/UX fixes for the Grants & Budget Lines and SCs & Budget Lines tabs on the agreement detail page.

**`DetailsTabs.jsx`**
- Corrected tab label from "Grant & Budget Lines" to "Grants & Budget Lines" (pluralized to match the SC tab convention).

**`AgreementBudgetLinesHeader.jsx`**
- Added `margin-top-6` to the budget lines header container for consistent vertical spacing.

**`GrantNumberForm.jsx` / `ServicesComponentForm.jsx`**
- The "Editing…" tag (pen icon + text) was vertically centered (`flex-align-center`) relative to the full `FormHeader` block (heading + details text), causing it to float below the heading. Changed to `flex-align-start` so the tag aligns with the heading row instead. Added small bottom padding so the tag doesn't crowd the content below.

**`CreateBLIsAndSCs.hooks.js`**
- After a successful save from the Grants & Budget Lines or SCs & Budget Lines tab, the success alert was redirecting to `/agreements/${id}` (Agreement Details tab). Changed both redirect paths to `/agreements/${id}/budget-lines` so the user lands back on the tab they were editing. Blocker-modal navigation (`savedViaModal`) is unchanged.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6013

## How to test

1. Open an existing grant agreement in edit mode and navigate to the **Grants & Budget Lines** tab.
   - Confirm the tab label reads "Grants & Budget Lines" (plural).
2. Click **Edit** on a grant number row to open the edit form.
   - Confirm the "Editing…" tag (pen icon) appears inline with the "Edit Grant Numbers" heading, not centered below it.
3. Make a change and click **Save Changes**.
   - Confirm the success banner appears and the page redirects to the **Grants & Budget Lines** tab (not the Agreement Details tab).
4. Repeat steps 2–3 for a **contract agreement** on the **SCs & Budget Lines** tab to verify the same behavior for `ServicesComponentForm`.

Frontend unit tests:
```bash
cd frontend
bun run test --watch=false
```

## A11y impact

## Storybook

## Screenshots

Screenshots should be added by the author to show the before/after of the "Editing…" tag alignment and the post-save redirect destination.

## Links

N/A

### DIFF
--- a/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
+++ b/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
@@ -68,6 +68,12 @@ export const isDeletionRoutedToApproval = (budgetLine, isSuperUser) => {
 };
 
 /**
+ * @param {number} agreementId - The id of the agreement.
+ * @returns {string} The path to the agreement's budget lines page.
+ */
+const getBudgetLinesUrl = (agreementId) => `/agreements/${agreementId}/budget-lines`;
+
+/**
  * Custom hook to manage the creation and manipulation of Budget Line Items and Service Components.
  *
  * @param {Function} setIsEditMode - Function to set the edit mode.
@@ -583,7 +589,7 @@ const useCreateBLIsAndSCs = (
                                     heading: "Changes Sent to Approval",
                                     message:
                                         "Your changes have been successfully sent to your Division Director to review. Once approved, they will update on the agreement.",
-                                    redirectUrl: `/agreements/${selectedAgreement?.id}/budget-lines`
+                                    redirectUrl: getBudgetLinesUrl(selectedAgreement?.id)
                                 });
                                 resolve();
                             }
@@ -651,7 +657,7 @@ const useCreateBLIsAndSCs = (
                         ` ${pendingChanges}`,
                     redirectUrl: savedViaModal
                         ? blocker.location?.pathname
-                        : `/agreements/${selectedAgreement?.id}/budget-lines`
+                        : getBudgetLinesUrl(selectedAgreement?.id)
                 });
             } else {
                 setAlert({
@@ -660,7 +666,7 @@ const useCreateBLIsAndSCs = (
                     message: `The agreement ${selectedAgreement?.display_name} has been successfully updated.`,
                     redirectUrl: savedViaModal
                         ? blocker.location?.pathname
-                        : `/agreements/${selectedAgreement?.id}/budget-lines`
+                        : getBudgetLinesUrl(selectedAgreement?.id)
                 });
             }
         },
@@ -1064,7 +1070,7 @@ const useCreateBLIsAndSCs = (
                     resetForm();
                     dispatch({ type: "RESEED_BUDGET_LINE_ITEMS", payload: [] });
                     setIsEditMode(false);
-                    navigate(`/agreements/${selectedAgreement?.id}/budget-lines`);
+                    navigate(getBudgetLinesUrl(selectedAgreement?.id));
                     scrollToTop();
                 }
             }
--- a/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
+++ b/frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
@@ -583,7 +583,7 @@ const useCreateBLIsAndSCs = (
                                     heading: "Changes Sent to Approval",
                                     message:
                                         "Your changes have been successfully sent to your Division Director to review. Once approved, they will update on the agreement.",
-                                    redirectUrl: `/agreements/${selectedAgreement?.id}`
+                                    redirectUrl: `/agreements/${selectedAgreement?.id}/budget-lines`
                                 });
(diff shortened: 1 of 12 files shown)


frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.hooks.js
```diff
@@ -649,14 +649,18 @@ const useCreateBLIsAndSCs = (
                         `Your changes have been successfully sent to your Division Director to review. Once approved, they will update on the agreement.\n\n` +
                         `<strong>Pending Changes:</strong>\n` +
                         ` ${pendingChanges}`,
-                    redirectUrl: savedViaModal ? blocker.nextLocation : `/agreements/${selectedAgreement?.id}`
+                    redirectUrl: savedViaModal
+                        ? blocker.nextLocation
```
weimiao67: **Finding 2 — `blocker.nextLocation` (full Location object) instead of `.pathname`**

Both `savedViaModal` ternary branches here pass `blocker.nextLocation` — the full React Router Location object — as `redirectUrl`. Line 527 in this same file correctly uses `blocker.nextLocation?.pathname`. Same issue on line 662.

Redux Toolkit's serialization check will warn (or error in strict mode) if `blocker.nextLocation.state` is non-null, since a Location object is not a plain serializable value.

Suggested fix:
```js
redirectUrl: savedViaModal
    ? blocker.nextLocation?.pathname
    : `/agreements/${selectedAgreement?.id}/budget-lines`
```

weimiao67 (commented):
Found and resolved 2 issues:

**Finding 1 — `handleFinancialSnapshotChanges` redirect not updated (line 586)**

The `showSuccessMessage` redirect was correctly updated to `/budget-lines`, but `handleFinancialSnapshotChanges` at line 586 still uses the old path:

```js
redirectUrl: `/agreements/${selectedAgreement?.id}`  // missing /budget-lines
```

This is masked in the normal save flow because `showSuccessMessage` runs afterward and overwrites the alert. But on the `suppressSuccessAlert=true` path (batch/saveTrigger save), `showSuccessMessage` is skipped and this stale redirect fires, sending the user to the Agreement Details tab instead of Budget Lines.

rajohnson90:
Below is what my AI review found. I will start fixing these myself, starting from issue 1 but feel free to mention if you want to take any others.

1. blocker.nextLocation doesn't exist (CreateBLIsAndSCs.hooks.js:652) — the "fix" reads a property react-router's Blocker never had (should
  be blocker.location, used correctly elsewhere in the same file). The redirect-after-save path only works today because of a redundant
  navigate() call elsewhere.
  2. Alignment fix in GrantNumberForm.jsx:77 targets the wrong container and papers over it with a magic padding-bottom-6 that will break
  again with different content length.
  3. AgreementBudgetLinesHeader.jsx:46 — unconditional margin-top-6 added to a header shared by two call sites, risking doubled spacing in
  one of them.
  4. Same "Editing..." tag markup duplicated across 3 components, fixed inconsistently in 2 and untouched in the 3rd.
  5. Redirect URL string literal now hardcoded 3x in one file — the same duplication pattern that caused this bug.
  6. The test mock's useBlocker shape doesn't match react-router's real API, so it can't catch #1.

weimiao67 (approved):

rajohnson90:
Follow-up fixes from a code review of PR #6160, covering two areas:

  ### Budget lines redirect bug
  - Replaced `blocker.nextLocation` — a property that doesn't exist on react-router's `Blocker` object (the real one is `.location`) — with
  `blocker.location?.pathname` at all 3 call sites plus a `useCallback` dependency in `CreateBLIsAndSCs.hooks.js`. This was a dead reference that silently
  broke the post-save redirect when a user saved via the "unsaved changes" modal.
  - Extracted the repeated `/agreements/${id}/budget-lines` string into a `getBudgetLinesUrl(agreementId)` helper, used at all 4 call sites in that file.
  - Fixed the `useBlocker` test mock to match react-router's real shape (`location: { pathname }` instead of a bare `nextLocation` string), and added a
  regression test asserting the redirect actually targets the blocker's pending destination when saved via the modal.

  ### Duplicated "Editing…" indicator
  - The pen-icon + "Editing…" tag was copy-pasted with inconsistent alignment fixes (magic `padding-bottom-6`/`padding-bottom-4` hacks) across
  `GrantNumberForm.jsx`, `ServicesComponentForm.jsx`, and `AgreementDetailHeader.jsx`.
  - Extracted a shared `EditingIndicator` component and gave `FormHeader` a new `actions` slot so it renders alongside the heading (rather than against the
  whole heading+details block, which is what caused the alignment bugs).
  - All three components now render the same indicator; the magic padding is gone.
  - Added test coverage for `FormHeader`'s new `actions` prop (including a guard against a falsy `actions` leaking `"false"` into the DOM).

  ### Self-review
  Ran two adversarial reviews against these changes. Caught and fixed one real regression (a `margin-top-6` layout fix I made earlier was reverted after
  verifying it broke spacing at the wizard call site) and one design gap (`FormHeader`'s `actions` slot was pushing wrapper-div duplication into callers —
  now owned internally by `FormHeader`).

rajohnson90 (approved):

Issue #6178, opened 2026-08-31
Edit button disappeared instead of just disabled


kimschulke:
The edit button on agreements has disappeared from certain situations, when it should not.

## Expected Behavior
The edit button an agreements (any editable tabs on all agreement types) should **always** display on agreements, **even when the user cannot edit the agreement**. In situations where the user cannot edit the agreement, we never hide the edit button, we simply disable it with a tooltip explaining why.

## Current Behavior
The edit button has disappeared from situations where the user cannot edit.

## Possible Cause
Possible regression because this previously was functioning.

## Steps to Reproduce
1. Click on an agreement you are not listed as a team member on or that is in pre-award
2. View no edit button and no tooltip explaining why
3. Be confused

## Context
This is critical to ensure users have visibility of the system status; so they can understand how the system works and whats happening without relying on underlying knowledge or guessing how things work.

## To do

- [x] for scenarios when the user cannot edit because they are not a team member, disable the edit button and display the following tooltip "Only team members can edit this agreement."
- [x] for scenarios when the user is a team member, but they cannot edit because the agreement is in pre-award, disable the edit button and display the following tooltip "This agreement is In Review for Pre-Award Approval. Edits or changes cannot be made at this time."
- [x] for scenarios when the user is a team member, but they cannot edit because the agreement is in award, disable the edit button and display the following tooltip "This agreement is In Review for Award Approval. Edits or changes cannot be made at this time."
- [x] please check any other tooltips we display on the status change button for other disabled scenarios and match the tooltip.
- [x] insure relevant unit or BDD tests are in place that keeps checking for this

npage123:
Would it be possible to not include the 'Edit' buttons at all for read only users please?

kimschulke:
### UX/UI Review Feedback 

- I think the only remaining item here was to remove edit mode for Read Only... Testing now and I am not seeing the edit mode for Read Only role :white_check_mark:
- The only thing I am noticing is that on the list pages/tables, when in hover state, the last column data gets hidden. See below :warning:

### Table data - default state, no hover
<img width="1018" height="223" alt="Image" src="https://github.com/user-attachments/assets/7ab41017-16bf-4eb2-a645-f2e563267e44" />


### Table data - hover state

<img width="1013" height="229" alt="Image" src="https://github.com/user-attachments/assets/71950e68-884e-44d7-90e5-21116df824da" />

npage123:
[...]

PR #6188, opened 2026-09-02
fix: keep agreement Edit button visible and disabled with a tooltip

Santi-3rd:
## What changed

Fixes a bug where the agreement **Edit** button disappeared entirely when the current user couldn't edit, instead of remaining visible-but-disabled with an explanatory tooltip. Both header components had a disabled-with-tooltip branch that was effectively dead code, because the enabled *and* disabled branches were both gated on the single `isEditable` prop (which the parents had already folded review/lock state into).

The fix keeps the parent's `isEditable` as the single source of truth for *enabled vs. disabled*, and always renders the button — enabled when `isEditable`, otherwise disabled with a tooltip. The individual *reason* flags are now passed down purely to select the correct tooltip text.

- **`agreement.helpers.js`** — new `getEditDisabledTooltip(...)` helper + `EDIT_DISABLED_TOOLTIPS` map so both headers share tooltip strings. Precedence (first match wins): not-a-team-member → not-developed → pre-award review → award review → post-pre-award lock → all-budget-lines-in-review.
- **`AgreementDetailHeader.jsx`** — always render the button; enabled when `isEditable`, else disabled `<span role="button">` + `Tooltip` via the helper. Accepts new `canUserEdit` / `isAgreementNotDeveloped` props.
- **`AgreementBudgetLinesHeader.jsx`** — same change; also accepts `allBudgetLinesInReview`; aligned disabled-span a11y attributes (`role="button"`, `tabIndex={0}`, `aria-disabled`) with the Details header.
- **`AgreementDetails.jsx` / `AgreementBudgetLines.jsx`** — editable computations left untouched; pass the new reason props down.
- **Tests** — new `AgreementDetailHeader.test.jsx` and `AgreementBudgetLinesHeader.test.jsx` (net new coverage); helper + page tests updated (e.g. the undeveloped-agreement case now asserts the disabled button + "coming soon" tooltip instead of button absence).

## Issue

https://github.com/HHS/OPRE-OPS/issues/6178

## How to test

1. `cd frontend && bun run test --watch=false` — all unit tests pass (4392 passing locally).
2. Manual (`docker compose up`), verifying the exact tooltip text per scenario:
   - Non-team-member on a normal agreement → disabled Edit button + tooltip: **"Only team members can edit this agreement."**
   - Team member on a **pre-award** agreement → disabled button + tooltip: **"This agreement is In Review for Pre-Award Approval. Edits or changes cannot be made at this time."**
   - Team member on an **award** agreement → disabled button + tooltip: **"This agreement is In Review for Award Approval. Edits or changes cannot be made at this time."**
   - Team member on a **post-pre-award-locked** agreement → disabled button + tooltip: **"This agreement has completed Pre-Award Approval and is locked from further edits."**
   - Regular user on an **undeveloped** agreement (grant / IAA / IPA / IDDA / direct obligation) → disabled button + tooltip: **"Agreements that are grants, other partner agreements (IAAs, IPAs, IDDAs), or direct obligations have not been developed yet, but are coming soon."**
   - Budget Lines tab, all budget lines in review → disabled button + tooltip: **"Budget lines In Review Status cannot be sent for status changes"**
   - Superuser on an undeveloped agreement → **enabled** button, edit form works (unchanged).
   - Verify the tooltip appears on both hover and keyboard focus.

## A11y impact

## Storybook

## Screenshots

_To be added — disabled Edit button + tooltip across scenarios._

## Links

N/A


frontend/src/components/Agreements/AgreementBudgetLinesHeader.jsx
```diff
@@ -86,17 +91,19 @@ export const AgreementBudgetLinesHeader = ({
                             <span className="text-primary">Edit</span>
                         </button>
                     )}
-                    {/* DISABLED EDIT BUTTON */}
-                    {!isEditMode && isEditable && isEditDisabled && (
+                    {/* DISABLED EDIT BUTTON - always shown (with tooltip) when editing is not allowed */}
+                    {!isEditMode && !isEditable && (
```
weimiao67: 🐛 **Possible regression: spurious disabled Edit button in the “Edit Budget Lines” screen.**

This gate changed from `isEditable && isEditDisabled` to `!isEditMode && !isEditable`. But `AgreementBudgetLinesHeader` has a **second** consumer that this change also affects — `CreateBLIsAndSCs.jsx:498` (the budget-line **edit** flow, `workflow === "none"`):

```jsx
<AgreementBudgetLinesHeader
    heading="Edit Budget Lines"
    includeDrafts={includeDrafts}
    setIncludeDrafts={setIncludeDrafts}
    isEditable={false}
/>
```

It passes `isEditable={false}` and no `isEditMode` (defaults to `false`), so this branch is now `!false && !false` → `true`. Result: a disabled **Edit** pen button with the tooltip *“Only team members can edit this agreement.”* now renders in the **Edit Budget Lines** header while the user is actively editing. Before this PR nothing rendered there (old gate was `isEditable && isEditDisabled` → `false`).

This call site is mounted only under `{isEditMode && …}` in `AgreementBudgetLines.jsx`, so it always hits the edit flow. No test covers it, so CI stays green.

**Suggested fix** at `CreateBLIsAndSCs.jsx:498` — pass `isEditMode` (in scope, `true` here) so both buttons are suppressed:

```jsx
    isEditable={false}
    isEditMode={isEditMode}
```

Worth a regression test for that header instance too.

rajohnson90:
Feedback from my AI review:

1. Regression on a third caller (CreateBLIsAndSCs.jsx:498): This PR's new render rule !isEditMode && !isEditable causes a spurious disabled Edit button to
     appear on the Budget Lines tab screen — a caller that previously never rendered any Edit button under the old logic. No test covers it.
  2. Duplicated tooltip logic drifting (AgreementBudgetLines.jsx:106): The "Change BL Status" button has its own hand-rolled tooltip text for the same "not a team
     member" reason, with wording that's already inconsistent with the newly centralized getEditDisabledTooltip — undermining the PR's stated goal of preventing
     tooltip-string drift.
  3. Missing default on canUserEdit (agreement.helpers.js:430): Every other param in getEditDisabledTooltip defaults to false; canUserEdit doesn't, so a future
     caller that omits it silently falls into the wrong tooltip branch.
  4. Duplicated markup across two header components (AgreementDetailHeader.jsx:79): Both header components independently render nearly identical
     disabled-Edit-button markup and call the helper separately, so a11y/markup fixes must be applied twice.

josbell (changes requested):
## Review findings

**Requesting changes for one confirmed regression.**

### 🔴 Finding 1 — Wrong tooltip in CreateBLIsAndSCs (confirmed regression)

**File:** `frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.jsx:498`

`AgreementBudgetLinesHeader` is rendered here with `isEditable={false}` hardcoded, but `canUserEdit` is never passed (defaults to `false`). The new `!isEditMode && !isEditable` condition means the disabled button renders on page load (when `isEditMode=false`, its initial state). Because `canUserEdit=false`, `getEditDisabledTooltip` immediately matches the `!canUserEdit` case and returns "Only team members can edit this agreement." — shown to **all** users, including team members.

The `isEditMode={isEditMode}` prop added in the follow-up commit only prevents the button from persisting *during* edit mode; it does not fix the wrong tooltip or suppress the spurious button on page load.

**Fix options:**
- Pass `canUserEdit` at this call site (and suppress the disabled button when it's not contextually meaningful — this is already an edit surface), or
- Suppress the disabled Edit button entirely in this wizard context (`isEditable={false}` here means "no inline Edit button", not "you are blocked from editing")

---

### 🟡 Finding 2 — Fragile default fallback in `getEditDisabledTooltip` (not a current regression, but worth noting)

**File:** `frontend/src/helpers/agreement.helpers.js:451`

The `default` branch returns `notTeamMember` even when `canUserEdit=true`. This is unreachable with the current parent wiring (every code path that makes `isEditable=false` while `canUserEdit=true` also sets a covered flag), but the test at line 720 asserts this wrong fallback as correct, which will silently mislead whoever adds the next blocking reason. Consider a generic fallback message or a guard comment explaining why this branch should never be reached.

weimiao67 (approved):

rajohnson90 (approved):

josbell (approved):

PR #6189, opened 2026-09-02
feat: make CLIN column sortable in budget line review table

weimiao67:
## What changed

Makes the **CLIN** column in the budget line review table (`BLIReviewTable`) sortable. Previously the CLIN header rendered as plain, non-clickable text. Clicking the header now sorts budget lines by CLIN number.

Details:
- Added a `CLIN` sort code to `tableSortCodes.budgetLineCodes` and wired it as the CLIN header's sort `value`, so the shared `Table` component renders it as a sort button.
- Sorting uses the **displayed** CLIN value: it prefers a locally-assigned (possibly unsaved) CLIN over the persisted `budgetLine.clin.number`, so the order matches what the user sees during the award-approval flow.
- CLIN values sort **numerically** (CLIN 2 before CLIN 10).
- Rows without a CLIN (Draft "N/A", unassigned "TBD"/"—") always sort to the **bottom**, regardless of ascending/descending direction, via a `SORT_TO_END` sentinel scoped to the CLIN case only (no behavior change to other columns).
- No backend change — `budgetLine.clin.number` is already in the API payload; sorting is entirely client-side.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6138

## How to test

1. Open an awarded contract agreement so the budget line review table shows the CLIN column.
2. Click the **CLIN** column header — rows should sort by CLIN number ascending; click again for descending.
3. Verify numeric ordering (e.g. CLIN 2 appears before CLIN 10, not lexicographically).
4. Verify rows without a CLIN (Draft "N/A", unassigned "TBD"/"—") stay at the bottom in both sort directions.
5. In the award-approval flow, assign a CLIN locally (without saving) and confirm sorting reflects the assigned value.

## A11y impact

- [x] No accessibility-impacting changes in this PR

The CLIN header now uses the same existing sortable-header pattern (button + `aria-sort`) already used by the other sortable columns.

## Storybook

- [x] N/A — change is page-specific or non-visual

## Screenshots

_N/A_

## Definition of Done Checklist
- [x] OESA: Code refactored for clarity
- [x] OESA: Dependency rules followed
- [x] Automated unit tests updated and passed
- [ ] Automated integration tests updated and passed
- [ ] Automated quality tests updated and passed
- [ ] Automated load tests updated and passed
- [ ] Automated a11y tests updated and passed
- [ ] Automated security tests updated and passed
- [x] 90%+ Code coverage achieved
- [ ] Form validations updated

## Links

_N/A_

Issue #6138, opened 2026-08-25
View CLINs on Awarded Agreements


kimschulke:
### User Story
As an OPS user, I want to view CLINS on an awarded agreement, so that I know what CLIN each budget line is assigned to in the award document.

### Acceptance Criteria
- [ ] TBD

-----------------------------------------

#### UX
- [x] Create designs/wireframes that separate awarded agreements from mods
- [ ] Add all annotations, biz rules and interaction notes
- [ ] Update OPS business rules document

#### Dev
- [ ] TBD

### Definition of Done Checklist
- [ ] UI works as designed (UX team)
- [ ] Usability validated (UX team)
- [ ] PR(s) have been merged to main
- [ ] Design/tech debt eliminated
- [ ] New design/tech debt documented (if applicable)
- [ ] Build process updated
- [ ] Documentation updated or added
- [ ] Feature flags/toggles created

### Additional Context & Resources
* _Any additional information that would be useful to share about this work_
* _Dependencies, constraints?_
* _Info already known or found, previous work that's been done?_

rajohnson90:
[Figma Link](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=15133-60380&t=oPlfiJCv9cJYSUDg-4)

kimschulke:
I assume we are starting with displaying CLINs when they are entered into the procurement tracker / step 6 inside of OPS.... but we need to also discuss if this GH issue will include displaying CLINs when agreements are migrated into OPS in an already awarded state?

jonnalley:
@kimschulke @npage123 @andressacurvelo this is awaiting your review in dev and staging

npage123:
@weimiao67 Would it be possible to make the CLIN column sortable in the budget lines tables please?

npage123:
@weimiao67 @jonnalley @kimschulke @andressacurvelo Other than the comment I left above, after reviewing in staging, things look good from a UX Review standpoint. However, did we ever answer the question Kim had about if this GH issue will include displaying CLINs when agreements are migrated into OPS in an already awarded state?

weimiao67:
@kimschulke @npage123  — the answer includes 2 parts:

[...]

PR #6210, opened 2026-09-07
fix: make CLIN column sortable in awarded contract budget lines table

weimiao67:
## What changed

Two related changes to the **agreement details → Budget Lines** experience for awarded contracts:

1. Makes the **CLIN** column sortable in the `BudgetLinesTable` and refines when/how it appears.
2. **Releases the post-pre-award BLI editing lock** once the Award request is approved, so budget lines can be edited during the award-completion window.

### Sortable CLIN column

PR #6189 made CLIN sortable in `BLIReviewTable` (the review / award-approval flows), but the awarded-contract details view uses a different table — `BudgetLinesTable` — whose `AWARDED_CONTRACT_BUDGET_LINE_TABLE_HEADERS` still hard-coded the CLIN header as `value: ""`. In the shared `Table` component, an empty `value` means `isSortable` is `false`, so the header rendered as plain text: no hover, no click, no sort. This is the case a user hits when they "open an awarded contract agreement" and click the CLIN header.

- Changed the CLIN header `value` from `""` to `tableSortCodes.budgetLineCodes.CLIN`, so the shared `Table` renders it as a sort button.
- No hook change needed: `BudgetLinesTable` already passes `SORT_TYPES.BUDGET_LINES`, which routes to `getBLIDiffComparableValue` — the function that already carries the CLIN sort case added in #6189.
- Sorting is scoped per rendered table (the existing per-group behavior), which is acceptable for this view.
- `budgetLine.clin.number` is already in the payload — no backend change for this part.

### CLIN column now shows in edit mode
- Previously the CLIN column only rendered in the read-only Budget Lines view. Clicking **Edit** swaps in `CreateBLIsAndSCs`, whose `BudgetLinesTable` was not receiving `showClinColumn`, so the column disappeared while editing.
- `CreateBLIsAndSCs` now accepts an `isAgreementAwarded` prop and derives `showClinColumn = isContract && isAgreementAwarded` — mirroring the logic already used by the read-only view — and passes it to the contract-branch `BudgetLinesTable`.
- The prop defaults to `false`, so the create wizard, review-flow edit, and award-approval paths (none of which are awarded contracts) are unchanged.

### CLIN display rules (awarded contracts)
Centralized in `getClinDisplayValue`, so the rules apply to the read-only view, the edit-mode table, and the XLSX export:
- **Draft** budget lines always show **`N/A`**.
- **Planned / Executing / Obligated** budget lines show the **CLIN number** if one exists, otherwise **`TBD`** (previously an em dash `—`). A CLIN numbered `0` still displays as `0`.

### Release post-pre-award BLI lock on award approval (backend)
- `is_post_pre_award_locked` (`backend/ops_api/ops/utils/budget_line_items_helpers.py`) previously returned `True` **permanently** once pre-award was fully approved (DD approved + Budget Team submitted requisition), locking BLI editing from that point on.
- It now **releases the lock** as soon as the Budget Team **approves the Award request** (AWARD step `award_approval_status == "APPROVED"`), even if the AWARD step itself has not been completed — reopening BLI editing for the award-completion window.
- A **pending / not-yet-decided** Award request, or a **DECLINED** one, leaves the lock in place.
- Added unit tests covering the approved (unlocked), pending (locked), and declined (locked) cases.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6138

## How to test

1. Open an awarded contract agreement and go to the **Budget Lines** view (details page), where the CLIN column is shown.
2. Hover the **CLIN** column header — it now shows the sortable-header affordance (cursor/arrow), like the other sortable columns.
3. Click the **CLIN** header — rows sort by CLIN number. First click sorts descending; click again for ascending.
4. Verify numeric ordering (e.g. CLIN 2 appears before CLIN 10, not lexicographically).
5. Verify rows without a CLIN (Draft `N/A`, non-draft `TBD`) stay at the bottom in both sort directions.
6. Click **Edit** on the Budget Lines tab — the CLIN column is still shown while editing.
7. Confirm Draft rows show `N/A`, and Planned/Executing/Obligated rows show their CLIN number or `TBD` when none is assigned.
8. On an agreement whose pre-award is fully approved but whose Award request is still pending, confirm BLIs remain locked; once the Budget Team **approves the Award request**, confirm BLIs become editable again.

## A11y impact

The CLIN header now uses the same existing sortable-header pattern (button + `aria-sort`) already used by the other sortable columns in this table.

## Storybook

## Screenshots

_N/A_

## Links

_N/A_


Santi-3rd (approved):

josbell (changes requested):
## Code Review

Overall the feature logic is solid — the sortable CLIN column, TBD display rules, and `is_post_pre_award_locked` unlock-on-award-approval are all correct and well-tested. One confirmed bug needs fixing before merge.

---

### 🔴 Must fix — CLIN column missing in review-route edit flow

**`frontend/src/pages/agreements/review/EditAgreementAndBudgetLines.jsx`, line 363**

`isAgreementAwarded` is computed at line 315 but never forwarded to `<CreateBLIsAndSCs>`:

```jsx
// line 315
const isAgreementAwarded = agreement?.is_awarded;

// line 363 — missing isAgreementAwarded={isAgreementAwarded}
<CreateBLIsAndSCs
    workflow="agreement"
    budgetLines={agreement?.budget_line_items ?? []}
    selectedAgreement={agreement}
    ...
/>
```

`CreateBLIsAndSCs` defaults `isAgreementAwarded` to `false`, so `showClinColumn` is always `false` here — the CLIN column is permanently hidden when editing an awarded contract via `/agreements/review/:id/edit`. This route is directly reachable via the "Edit" button in `ReviewAgreement.jsx`.

**Fix:** Add `isAgreementAwarded={isAgreementAwarded}` to the `<CreateBLIsAndSCs>` call at line 363.

---

### 🟡 Suggestion — Add a DB uniqueness constraint on `(procurement_tracker_id, step_type)`

`ProcurementTrackerStep` has no `UniqueConstraint` on `(procurement_tracker_id, step_type)`. The `is_post_pre_award_locked` function uses `next()` to find the first AWARD step — if a second AWARD step were ever inserted (e.g., by a future reset flow or manual migration), a stale APPROVED step could silently release the lock while a second PENDING step is still under review.

No current code path creates duplicate steps, so this is not an active bug. A defensive migration adding `UniqueConstraint("procurement_tracker_id", "step_type", name="uq_tracker_step_type")` would future-proof the lock logic.

---

### 🔵 Cleanup (non-blocking)

**Duplicate `showClinColumn` derivation**

`AgreementBudgetLines.jsx:91` and `CreateBLIsAndSCs.jsx:156` contain byte-for-byte identical logic:
```js
const showClinColumn = isContract && isAgreementAwarded;
```
The confirmed bug above is a direct symptom — callers of `CreateBLIsAndSCs` must remember to pass `isAgreementAwarded` themselves. Consider computing `showClinColumn` only in the parent and passing it as a single boolean, or extracting to a shared helper.

**`isAgreementAwarded` is a dead prop in `BudgetLinesTable`/`BLIRow`**

`BudgetLinesTable` accepts `isAgreementAwarded` and forwards it to `BLIRow`, but `BLIRow` never destructures or uses it — `showClinColumn` is the sole gate for both header selection and cell rendering. The prop can be removed from `BudgetLinesTable` and `BLIRow`'s interface to avoid misleading future readers.

weimiao67:
@josbell Thanks for the thorough review. Addressed in fb7b424df:

**🔴 #1 — CLIN missing in review-route edit — fixed.** You're right that the prop was computed but never forwarded, and that editability is BLI-status-based (not gated on `is_awarded`), so an awarded contract with a DRAFT/PLANNED BLI can reach `/agreements/review/:id/edit`. Added `isAgreementAwarded={isAgreementAwarded}` to the `<CreateBLIsAndSCs>` call in `EditAgreementAndBudgetLines.jsx`, plus two regression tests asserting the prop forwards for an awarded contract and defaults to `false` otherwise. (Note: this surface never rendered CLIN before, so it was a pre-existing gap rather than a regression from this PR — but the one-liner is fully in the spirit of the change, so happy to close it here.)

**🔵 #3 — dead prop + duplicate derivation — fixed (dead prop).** Removed the dead `isAgreementAwarded` prop from `BudgetLinesTable` and `BLIRow` (both gate solely on `showClinColumn`), and dropped the now-dead pass-throughs in `AgreementBudgetLines.jsx`. I left the `showClinColumn` derivation where it is: `CreateBLIsAndSCs` derives `isContract` internally from `selectedAgreement.agreement_type`, so hoisting `showClinColumn` into the parents would force every caller (including `EditAgreementAndBudgetLines`, which doesn't currently compute `isContract`) to derive it — net-neutral. Removing the dead prop is the clean win; happy to extract a shared helper in a follow-up if you feel strongly.

**🟡 #2 — DB uniqueness constraint — respectfully deferring.** A couple of reasons I'd rather not include this here:
- As noted, it's not an active bug — no code path creates duplicate steps.
- The `next()`-picks-first concern isn't introduced by this PR; it's the established codebase pattern. `ProcurementTracker.get_step()` is literally `next((s for s in self.steps if s.step_type == step_type), None)`, and `is_pre_award_in_review` / `is_award_approval_requested` already rely on the same "one step per type per tracker" invariant. This PR doesn't add a new risk class.
- A `UniqueConstraint` is a schema migration well outside this PR's scope (CLIN sorting + lock release) and needs its own Alembic migration + review.

If the team wants that invariant enforced at the DB level, I think it deserves its own ticket. If anything, the in-scope nit would be to have the new AWARD lookup call `tracker.get_step(...)` instead of re-implementing the `next()` inline — glad to do that if you'd like.

backend/ops_api/ops/utils/budget_line_items_helpers.py
```diff
@@ -299,10 +303,20 @@ def is_post_pre_award_locked(agreement) -> bool:
...
+    pre_award_fully_approved = (
         pre_award_step.pre_award_approval_status == "APPROVED"
         and pre_award_step.pre_award_requisition_approved_by is not None
     )
+    if not pre_award_fully_approved:
+        return False
+
+    # Release the lock once the Budget Team approves the Award request. A DECLINED (or not-yet-
+    # decided) Award request leaves the lock in place.
+    award_step = next((step for step in tracker.steps if step.step_type == ProcurementTrackerStepType.AWARD), None)
```
josbell, line 315: Nit: every other callsite in this file (lines 215, 257, 302) and in `agreement_award_history.py` uses `tracker.get_step(ProcurementTrackerStepType.AWARD)`. The two are functionally identical (`get_step` is a thin wrapper around the same expression), but staying consistent means this site would inherit any future defensive logic added to the helper.

```python
award_step = tracker.get_step(ProcurementTrackerStepType.AWARD)
```

backend/ops_api/ops/utils/budget_line_items_helpers.py
```diff
@@ -299,10 +303,20 @@ def is_post_pre_award_locked(agreement) -> bool:
...
+    if award_step and award_step.award_approval_status == "APPROVED":
```
josbell, line 316: Nit: `"APPROVED"` is a raw string literal — same pattern as line 307 for `pre_award_approval_status`. `agreement_award_history.py` has a module-private `_AWARD_APPROVED_STATUS` but it is intentionally scoped there. No `AwardApprovalStatus` enum exists yet; worth introducing one (or a shared public constant) to consolidate all these status comparisons and satisfy the CLAUDE.md convention of using enum classes over raw strings.

frontend/src/components/BudgetLineItems/CreateBLIsAndSCs/CreateBLIsAndSCs.jsx
```diff
@@ -149,6 +151,9 @@ export const CreateBLIsAndSCs = ({
 
     const isAgreementWorkflowOrCanEditBudgetLines = workflow === "agreement" || canUserEditBudgetLines;
     const isGrant = selectedAgreement.agreement_type === AGREEMENT_TYPES.GRANT;
+    const isContract = selectedAgreement.agreement_type === AGREEMENT_TYPES.CONTRACT;
```
josbell, line 154: Nit: `isContract` is declared solely for the next line and never referenced again in the component. Can be inlined:

```js
const showClinColumn = selectedAgreement.agreement_type === AGREEMENT_TYPES.CONTRACT && isAgreementAwarded;
```

josbell (approved):
Changes look good — clean prop-threading cleanup and the backend unlock logic is correct. A few minor nits inline.

Issue #6191, opened 2026-09-02
Grant Agreement History


rajohnson90:
### Goals
We would like to add the additional Agreement History messages necessary for users to track the audit log of changes made to the Grant. This is necessary because a Grant has some fields that are unique and not shared with the Contract Agreement

### Tasks
- [x] Design a set of Agreement History messages to represent changes to Grant specific fields
- [x] Add Agreement History triggers for each of the new fields to generate new messages
- [x] Add tests to verify that these history messages are being properly populated when an event fires with the relevant fields updated.

andressacurvelo:
@rajohnson90 here is the history messages in [Figma](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=35866-41773&t=KlSzOZOfkLRbMieO-4) let me know if you have any question!

npage123:
Feedback on 9/9 was to try and include in the history messages the detail in parenthesis with the ALN numbers in the dropdown in edit mode and on agreement details.

PR #6215, opened 2026-09-08
feat: add missing Grant agreement history messages

rajohnson90:
## What changed

Grant agreements were missing several Agreement History log entries that a design review identified as required, and two existing entries were mislabeled for Grants. This PR closes those gaps on the backend and adds E2E coverage proving the messages actually render.

- **`backend/models/agreement_history.py`**:
  - Adds `nofo_number` to the generic scalar-field history path (`Change to NOFO Number`).
  - Adds a `funding_period_months` case that formats values as "X months" (`Change to Grant Funding Period`).
  - Adds an `aln_number_changes` handler that emits one history row per added/removed ALN Number (`Change to ALN Numbers`) — there's no ALN lookup table, so the raw number is used directly.
  - Makes the Project Officer / Alternate Project Officer history labels agreement-type-aware via two new helpers (`get_project_officer_display_name`, `get_alternate_project_officer_display_name`), so Grants now correctly show **"Change to FPO"** / **"Change to Project Specialist"** instead of the hardcoded "Change to COR" / "Change to Alternate COR".
- **`backend/models/utils/__init__.py`**: adds the `aln_numbers` set-diff block in `generate_agreement_events_update`, mirroring the existing `special_topics` pattern, to produce the `aln_number_changes` payload consumed above.
- **`backend/data_tools/data/ops_event.json5`**: seeds 3 new `UPDATE_AGREEMENT` events against the fixture Grant agreement, covering all four new/relabeled message types.
- **`backend/ops_api/tests/ops/agreement/test_agreement_history.py`**: 3 new tests asserting the exact history title/message text for NOFO Number + Funding Period, ALN Numbers added/removed, and FPO/Project Specialist relabeling.
- **`frontend/cypress/e2e/editGrantAgreement.cy.js`**: extends the existing "edits grant metadata" E2E test to also edit Grant Funding Period, ALN Numbers, FPO, and Project Specialist, then reloads the agreement and asserts all five history messages (including the pre-existing NOFO Number one, which previously had no history assertion at all) appear in the History panel.

Grant Numbers already had full history support (create/update/delete) prior to this PR — no changes were needed there.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6191

## How to test

1. Backend: `cd backend/ops_api && pipenv run pytest tests/ops/agreement/test_agreement_history.py`
2. E2E: rebuild the seed data so the new `ops_event.json5` fixtures load, then run the spec:
   ```
   docker compose up db data-import --build
   docker compose up --build -d
   cd frontend && bunx cypress run --spec "cypress/e2e/editGrantAgreement.cy.js"
   ```
3. Manual check: create or open a Grant agreement, edit its NOFO Number, Grant Funding Period, ALN Numbers, FPO, and Project Specialist, save, and confirm each change appears in the Agreement History panel with the correct Grant-specific wording (FPO / Project Specialist, not COR / Alternate COR).

## A11y impact

## Storybook

## Screenshots

N/A — backend history-message changes, plus test-only frontend changes (no UI/visual changes).

## Links

N/A


Santi-3rd (approved):

Santi-3rd:
One finding from the AI agent that might be worth looking into before merging: 

 waitForAgreementHistory doesn't actually wait for the entries the test asserts on (frontend/cypress/e2e/editGrantAgreement.cy.js:19).

  The poll returns as soon as data.length > 0. But the grant is seeded via POST /agreements, which already produces an "Agreement Created" history entry. So by the time the PATCH fires, ≥1 entry almost always exists and the poll short circuits immediately — it never waits for the five update rows (NOFO / Funding Period / ALN / FPO / Project Specialist), which are written asynchronously by the MessageBus subscriber.

  That would be fine if the later text check retried, but it runs inside .invoke("text").then(...), and .then() does not retry — so if the async rows haven't landed yet, the expectedEntries loop fails. This is a real flake risk (and this repo runs flaky-E2E detection).

  Suggested fix: have the poll gate on the specific update content rather than mere presence, e.g. keep polling until the response text contains NOFO-UPDATED (or data.length >= expectedCount), so it waits for the actual new rows.

weimiao67 (approved):

rajohnson90:
Incorporated Guillermo's suggestion.

PR #6217, opened 2026-09-08
fix: fix disabled Edit button tooltip not showing on text hover

rajohnson90:
## What changed

Root-caused why hovering the "Edit" text (as opposed to the pen icon) in the disabled agreement Edit button didn't show its explanatory tooltip. USWDS's `usa-tooltip` JS module resolves the tooltip trigger from the raw `event.target` rather than the delegate-matched ancestor element — so when the hovered node is a child of the trigger (the "Edit" text span) instead of the trigger itself, `getTooltipElements` looks for the tooltip body in the wrong DOM location and silently fails (the pen icon "worked" only because its surrounding margin happened to hit the trigger's own box).

Fixed by making the icon and text non-hit-testable (`pointer-events: none`) so the browser always resolves the hovered element to the outer trigger span, regardless of where inside the button the cursor lands — the actual root cause, rather than working around it with a duplicate nested tooltip.

**`frontend/src/components/Agreements/DisabledEditButton.jsx`**
- Added `pointerEvents: "none"` to the icon's `style` and to the "Edit" text `<span>`.
- Removed a stray, inert `data-position="top"` attribute that was on the icon instead of the actual tooltip trigger.

## Issue

#6178

## How to test

1. `cd frontend && bun run test --watch=false` — full unit test suite should pass.
2. Manually: open an agreement's Details page or Budget Lines page where editing is disallowed (disabled Edit button visible). Hover directly over the "Edit" text (not the icon) — the tooltip explaining why editing is disabled should now appear immediately, the same as hovering the pen icon.
3. Confirm hovering the icon and tabbing to the button via keyboard still show the tooltip as before (no regression).

## A11y impact

This is an accessibility bug fix (tooltip content was previously unreachable via hover on part of the control) with no new violations introduced — leaving unchecked for reviewer/author sign-off.

## Storybook

## Screenshots

N/A — no visual change (tooltip behavior fix only; appearance is unchanged).

## Links

N/A

### DIFF
--- a/frontend/src/components/Agreements/DisabledEditButton.jsx
+++ b/frontend/src/components/Agreements/DisabledEditButton.jsx
@@ -35,11 +35,13 @@ const DisabledEditButton = ({ label, variant = "detail" }) => {
                     icon={faPen}
                     size={isBudgetLines ? "2x" : undefined}
                     className="height-2 width-2 margin-right-1"
-                    style={isBudgetLines ? { position: "relative", top: "2px" } : undefined}
+                    style={{
+                        pointerEvents: "none",
+                        ...(isBudgetLines ? { position: "relative", top: "2px" } : {})
+                    }}
                     aria-hidden="true"
-                    data-position="top"
                 />
-                <span>Edit</span>
+                <span style={{ pointerEvents: "none" }}>Edit</span>
             </span>
         </Tooltip>
     );


weimiao67 (approved):

rajohnson90:
@rajohnson90 Testing something out about the tag. Not really relevant to this story so dont worry.

Santi-3rd (approved):

PR #6219, opened 2026-09-09
fix: fix and harden the Slack mention-notify workflow

rajohnson90:
## What changed

The `notify-mention` workflow's "Rate limit check" step has been failing 100% of the time (confirmed live against the real repo's run history — every recent run shows `conclusion: failure`), which meant **no Slack mention notifications have been going out at all**.

**Root cause (`.github/workflows/slack_mention_notify.yml`):** the `gh api` call used `-f actor=... -f created=...` without `-X GET`. `gh api` defaults to POST whenever `-f`/`-F` flags are present unless `-X` is explicitly given, and POST isn't valid against this GET-only "list workflow runs" endpoint — hence the 404 every time.

After fixing that, three parallel adversarial reviews (security, correctness, reliability) were run against the change and the file as a whole, since the workflow had never actually run to completion in production and several latent issues were about to be exercised for the first time. This PR also fixes what those reviews turned up:

- **`slack_user_map.json` failure guard** — a missing/malformed map file previously killed the whole step mid-loop under `set -e` before ever reaching the Slack POST. Now falls back to plain `@username` text instead.
- **Concurrency group** — added `concurrency: group: notify-mention-${{ github.event.comment.user.login }}` to close a TOCTOU race: the run-history API this limiter reads is cached up to 60s and includes in-progress runs, so a comment burst from one actor could have multiple concurrent runs all read the same stale, undercounted total and all slip past `MAX_RUNS`.
- **Retry + fail-safe instead of hard job-kill** — the `gh api` call now retries 3x with backoff; if it's still unverifiable, the workflow skips the Slack notification (fail-safe) instead of hard-aborting the entire job on any transient error.
- **Off-by-one fix** — `total_count` includes the currently-running invocation itself; the check now compares prior triggers only, so exactly `MAX_RUNS` (not `MAX_RUNS - 1`) invocations are allowed per window. This also closes a narrow silent-bypass path where a non-numeric API result could no-op the limiter.
- **Team mentions excluded from the notify regex** — `@org/team-name` previously matched as a mangled `@org` fallback; now skipped entirely since it's not an individual user mention.
- **`curl` timeout added** — `--connect-timeout 5 --max-time 10` so a hung Slack endpoint can't eat the full 5-minute job timeout.

Failed-run counting in the rate window was deliberately left unchanged — this limiter caps invocation frequency for abuse prevention, not notification success, so counting failures against quota is intentional.

## Issue

N/A — no associated ticket (CI/tooling fix, branch not tied to a story).

## How to test

There's no automated test suite for this GitHub Actions workflow, so verification is manual:

1. **Reproduce the original bug vs. the fix** (already done live against `HHS/OPRE-OPS` during development):
   ```bash
   # Old behavior (404): gh api "repos/HHS/OPRE-OPS/actions/workflows/slack_mention_notify.yml/runs" -f actor=<user> -f created=">=<ts>" --jq '.total_count'
   # Fixed behavior (200): gh api -X GET "repos/HHS/OPRE-OPS/actions/workflows/slack_mention_notify.yml/runs" -f actor=<user> -f created=">=<ts>" --jq '.total_count'
   ```
2. After merge, as an org member/collaborator, comment `@some-mapped-user hey` on an issue or PR and confirm a Slack message posts with the mapped mention.
3. Comment mentioning a team, e.g. `@HHS/some-team`, and confirm it's dropped from the notification instead of posting a mangled `@HHS`.
4. Post 6+ comments with mentions from the same account within 10 minutes and confirm the 6th is rate-limited (check the run's `::warning::` annotation).
5. Check `gh run list --workflow=slack_mention_notify.yml` after merge to confirm runs show `conclusion: success` instead of the prior 100%-failure streak.

## A11y impact

## Storybook

## Screenshots

N/A — backend/CI workflow change only, no UI.

No automated test suite covers GitHub Actions workflows in this repo; verification was manual/live as described above.

## Links

N/A

### DIFF
--- a/.github/workflows/slack_mention_notify.yml
+++ b/.github/workflows/slack_mention_notify.yml
@@ -8,6 +8,12 @@ permissions:
   contents: read
   actions: read
 
+# Serialize runs per commenter so the rate-limit check (which reads this workflow's own,
+# eventually-consistent run history) can't be raced by a burst of comments from one actor.
+concurrency:
+  group: notify-mention-${{ github.event.comment.user.login }}
+  cancel-in-progress: false
+
 jobs:
   notify-mention:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jonnalley (approved):
lgtm

weimiao67 (approved):

Issue #5892, opened 2026-07-03
Procurement tracker step 6 updates


npage123:
### Goals
* Findings from demoing step 6 of the procurement tracker to Sheila on 6/30/26.

### UX Tasks
- [x] Prompt the user to update the agreement title so it matches the signed award exactly
- [x] Add mod # (for a new award, it will prep-populate be Base, but for future mod versions, it will provide up to 20 options. Options should be labeled as P00001, P00002, etc.)
- [x] Add Purchase Order #
- [x] Add Task Order #
- [x] Add annotations
- [ ] ~~Update the business rule documentation~~

### Dev Tasks
- [x] Create new accordion section for Update Agreement Title, including instructional text and a component to enter the agreement title
- [x] Under current award information, add a modification # component
- [x] Under current award information, add a purchase order # component
- [x] Under current award information, add a task order # component

### Additional Context
* Please note that the purchase order # for a contract is known as the ODN to the budget team. For grants, its a different #. For AAs, its an order #. ODN is just a general term that encompasses all agreement types, even though each agreement type has its own unique # used to the obligating document number.

### Resources
* [Designs in Figma](https://www.figma.com/design/dsQFe6G2KjcTWNq7NNZujm/OPRE-OPS?node-id=24912-21301&t=forzCioZuYu6Qlpm-1)

kimschulke:
> Update designs for the COR view to include a mod number dropdown where on initial award, it will always be Base, but for future mod versions, up to 20 options. Options should be labeled as P00001, P00002, etc.

~~This one relates to modifying the current procurement tracker, which was designed for contracts/new awards, to also accommodate contract modifications for FY26. This would be a temporary work-around for users to utilize the procurement tracker for contract modifications until the mod feature is developed. There are some other changes that will need to take place in addition to being able to enter the modification number, so I suggest we create a new GH issue focused solely on modifying the existing procurement tracker for mods.~~

jonnalley:
above note from Kim is out of date now. we're not doing any temporary work-arounds now to incorporate mods into the PT but still need to do some updates for the award step.

kimschulke:
### UX/UI Review

- agreement title showing up
- mod #, PO# and task order # showing up

future stories
- keep an eye on update agreement title. we originally intended this component to default blank so users are encouraged to copy/paste from the award doc and ensure it matches the award, rather than leaving the existing title in place. however, its currently pre-populated with the current. we can leave as-is and monitor for future improvement
[...]

PR #6225, opened 2026-09-09
feat: add agreement title and award numbers to request award approval

Santi-3rd:
## What changed

Implements Procurement Tracker Step 6 (Award) updates for the **Request Award Approval** flow (OPS-5892), per the Figma design:

- **New "Update Agreement Title" accordion** (between *Add CLINs* and *Vendor Information*) with a text input pre-filled with the current agreement title. The entered title is stored on step 6 as a proposed title and applied to `agreement.name` only when the Budget Team approves.
- **Current Award Information** now also collects:
  - **Modification #** dropdown (`Base` + `P00001…P00020`, default `Base`)
  - **Purchase Order #** (ODN) text input
  - **Task Order #** text input
- All four new fields are required to submit for approval, persist on step 6, and re-display on the Edit page.
- Field order under *Current Award Information* reordered to match the design: **Contract # · Award Amount · Award Date** on the first row, **Modification # · Purchase Order # · Task Order #** on the second.
- **Length caps** consistent with the rest of the app: Agreement Title is capped at 200 characters at all three layers (input `maxLength`, Vest, backend schema) — the same cap `AgreementEditForm` puts on the `name` field this value overwrites. Purchase Order # / Task Order # are capped at 100.
- **Budget Team approval screen** now shows the submitted Modification # / Purchase Order # / Task Order # alongside the existing award fields, plus a new **Review Agreement Title** accordion contrasting the current agreement title with the proposed one before approving.
- **Award & Modification History** now reads the step 6 values: `modification_number`, `purchase_order_number`, and `task_order_number` prefer the AWARD step's per-cycle value and fall back to the previous sources (`AgreementMod.number`, and the agreement-level `po_number` / `task_order_number`) only when the step has none. This closes the divergence where a MAPS-imported contract kept showing its stale imported PO # / Task Order # after a COR entered new values at step 6, and where an OPS-native agreement showed blanks.
  - Two deliberate non-changes: this flow still does not write the agreement-level `po_number` / `task_order_number` columns (only the legacy MAPS import does), and the accordion header label still comes from `AgreementMod.number` (the prose cycle label, "Mod 1") rather than the award's modification number ("P00001") — those are allowed to differ.

Backend adds the new columns to the step 6 procurement tracker model (with Alembic migration), schema, and service handling, and threads them through the award history service.

## Issue

#5892

## How to test

1. Go to an agreement with Executing budget lines and open **Request Award Approval** (`/agreements/:id/award-approval`).
2. Confirm the **Update Agreement Title** accordion appears between *Add CLINs* and *Vendor Information*, pre-filled with the current title and showing the design copy.
3. Under **Current Award Information**, confirm the field layout matches the design:
   - Row 1: Contract #, Award Amount, Award Date
   - Row 2: Modification # (dropdown, default `Base`, options `P00001`–`P00020`), Purchase Order #, Task Order #
4. Verify all four new fields are required to submit for approval.
5. Submit, then reopen the **Edit** page and confirm all values re-display.
6. As Budget Team, open the review screen and confirm the **Review Agreement Title** accordion shows current vs. updated title and that the three new award numbers display; approve and confirm the agreement title is updated to the entered value.
7. Open the agreement's **Award & Modification History** accordion and confirm the Modification #, Purchase Order #, and Task Order # rows show what was entered at step 6. On a MAPS-imported contract, confirm they replace the imported values rather than showing the stale ones.

## A11y impact

Step 6 adds four new required form controls (three text inputs plus a Modification # `<select>`). Each has an associated `<label htmlFor>`, `required` / `aria-required="true"`, and error messages rendered in a `role="alert"` region before the control inside a `usa-form-group--error` wrapper, matching the existing fields in this form.

## Storybook

- [x] N/A — change is page-specific or non-visual (`AwardRequestForm` is a feature component under `src/components/Agreements/`, not `src/components/UI/`, and has no existing story)

## Screenshots

_See design: `Step 6 - Award - COR Request.pdf`_

Notes on the unchecked items:

- **Quality / load / security tests** — none exist for this flow and none were added; the CodeQL, Semgrep, Snyk and TruffleHog gates pass unchanged.
- **A11y tests** — no automated a11y test was added or updated, so this stays unchecked; the `A11y Regression Gate` and the existing Cypress a11y spec pass. The four new controls are covered by unit assertions on labels, `aria-required`, and the `role="alert"` error region (see *A11y impact* above).
- **Coverage** — every new branch has a test (see below), but `bun run test:coverage` cannot produce a repo-wide number locally: it fails on a pre-existing dead import in `CANBudgetSummary.jsx` (`api/getCanFiscalYear`, deleted in `2f44a8a1cc`) that only surfaces when coverage instruments untested files. Unrelated to this PR; leaving the box for whoever can run the gate.

Tests added by this PR:

- Frontend — new files: `EditAwardApproval.hooks.test.js` (17 cases), `RequestAwardApproval.suite.test.js` (20), `awardForm.helpers.test.js` (12). Extended: `RequestAwardApproval.hooks.test.js` (+8), `AwardRequestForm.test.jsx` (+8), and one tightened assertion in `ApproveAwardApproval.test.jsx`.
- Backend — new file: `test_step_6_proposed_agreement_title.py` (11 collected cases, API + service level). Extended: `test_agreement_award_history_service.py` / `test_agreement_award_history.py` (+3) and `test_procurement_tracker_steps_schema.py` for the four new fields.


rajohnson90:
Did a quick AI review and got this feedback for us to maybe address today:
 Most severe:
  1. backend/ops_api/ops/schemas/procurement_tracker_steps.py:488 — modification_number has no OneOf validation despite being a fixed
     21-value dropdown, inconsistent with how this same file validates other fixed-choice fields like approval_status. An out-of-list
     value via direct API PATCH would leave the frontend <select> unselected.
  2. RequestAwardApproval.hooks.js:136 — seeds agreementTitle from the agreement but never seeds
     modificationNumber/purchaseOrderNumber/taskOrderNumber from step6, unlike the Edit page. After a decline→resubmit, a COR's
     previously-entered PO#/TO# get wiped back to blank/defaults.
  3. EditAwardApproval.hooks.js:169 — agreementTitle seeds via ?? (handles empty string) while the other three fields seed only under
     truthy if guards, so an empty-string step6.modification_number would make hasChanged fire immediately on a pristine, untouched
     page.
  4. backend/openapi.yml:9018 — the 4 new fields were added to Marshmallow schemas but not synced to ProcurementTrackerAwardStep in
     the OpenAPI spec, violating the project's own documented /sync-openapi rule.

  Lower severity:
  5. Whitespace-trim asymmetry between hasChanged's comparison and the trimmed value actually submitted for agreementTitle.
  6. _apply_proposed_agreement_title mutates agreement.name directly via ORM, bypassing AgreementsService.update() and its
  ImmutableAwardedFieldsRule that every other field in this PR goes through.
  7. getModificationOptions() rebuilds a static 21-element array every render instead of being a module-level constant.
  8. The seed/hasChanged pattern for the 4 new fields is duplicated near-identically across RequestAwardApproval.hooks.js and
  EditAwardApproval.hooks.js.

weimiao67 (commented):
Reviewed the full diff and ran the suites locally: **61 frontend tests, 23 backend schema tests, 32 tracker-model tests all pass**, `black --check` is clean, and the migration chain is correct (`c4e7a9b1d2f3` revises `c9d0e1f2a3b4`, the current head, with the `_version` history table properly mirrored). The one duplicate `down_revision` in the tree (`b9c8d7e6f5a4`) pre-exists on `main`.

Nice work on the `hasChanged` logic — comparing pre-filled fields against their *seeded* values rather than `""`, with a regression test for the pristine-form case, is exactly the bug that pattern invites. `_remove_award_fields` being updated so the new fields don't leak onto non-AWARD steps is also good, and covered.

Inline comments below, roughly by priority. **The first one is the only one I'd consider blocking** — it's a question about intended scope rather than a defect. The rest are mergeable cleanups plus some test-coverage gaps.

One caveat on my own analysis, for transparency: my first pass concluded that nothing in the codebase ever writes `agreement.po_number`. That was wrong — I'd grepped only for literal `.po_number =` assignments and missed both the generic `db_session.merge()` path and the legacy ETL. The corrected version is in the inline comment; the user-visible consequence still stands, but the mechanism is different than I first thought.

backend/ops_api/ops/services/procurement_tracker_steps.py
```diff
@@ -1032,6 +1048,10 @@ def _handle_award_approval(self, step, approval_status, obligated_date, current_
         # AwardApprovalObligatedDateRequiredRule) and must never be assumed to be today —
         # it is generally first documented in another system.
         agreement = step.procurement_tracker.agreement
+
+        # OPS-5892: apply the proposed agreement title exactly at approval.
+        self._apply_proposed_agreement_title(step, agreement)
```
weimiao67, line 1076: **Question on intended scope — do the new PO # / Task Order # values need to reach the agreement?**

This propagates `name` to the agreement, but not the other three new step-6 fields. That matters because the **Award & Modification history** reads the *agreement-level* columns, not the step:

- `ops_api/ops/services/agreement_award_history.py:106-107` reads `agreement.po_number` and `agreement.task_order_number`, renders them at `:210-211`
- frontend consumes that at `helpers/awardModificationHistory.helpers.js:65` ("Task Order #")
- the accordion is **not** feature-gated — rendered unconditionally at `pages/agreements/details/Agreement.jsx:358`

Those agreement columns *are* writable (via `ContractAgreementData` in `schemas/agreements.py:107-108` → `agreement_cls(**updated_fields)` → `db_session.merge()` in `services/agreements.py:448-451`, and the legacy `load_contracts` ETL populates them), but nothing in the award flow and no frontend UI ever sets them.

So after a COR fills in Purchase Order # / Task Order # at step 6 and the Budget Team approves:
- **MAPS-imported contracts** → history keeps showing the stale *imported* value, silently diverging from what was just entered
- **OPS-native agreements** → history shows blank, even though step 6 collected the values

Same divergence for Modification #: the history uses `AgreementMod.number` (`:202`), not `award_modification_number`.

If propagation is a deliberate follow-up story, that's completely reasonable — could you note it in the PR description? The matching field names read as though this PR closes that loop, so it'd be easy for the next person to assume it's wired up.

backend/ops_api/ops/services/procurement_tracker_steps.py
```diff
@@ -1009,6 +1014,17 @@ def get_pending_requisitions_for_user(self, user_id: int) -> list[ProcurementTra
 
         return list(self.db_session.scalars(stmt).all())
 
+    @staticmethod
+    def _apply_proposed_agreement_title(step, agreement):
```
weimiao67, line 1018: **This is the riskiest behavior in the PR and it has no test coverage.**

All 23 new backend tests are schema serialization/validation only — I grepped `ops_api/tests/` for `award_agreement_title` / `proposed_title` / "agreement title" and the only hits are a pre-existing, unrelated assertion in `test_agreement_history.py`.

Worth pinning three cases in a service test:
1. `approval_status == "APPROVED"` → `agreement.name` overwritten with the proposed title
2. **declined / still-pending → `agreement.name` unchanged** (the early return at `:1044-1045`)
3. blank / whitespace-only title → name left alone (the `if proposed_title and proposed_title.strip()` guard)

There's also a subtlety worth locking down with a test: `ContractAgreement.immutable_awarded_fields` **includes `name`** (`models/agreements.py:721`), but `ImmutableAwardedFieldsRule` is only registered on the *agreement service* path (`services/agreements.py:418-423`), not in the step-6 validator chain (`validation/procurement_tracker_steps_validator.py:155-168`). So this write succeeds today purely because the two validator stacks are separate. Note that `_advance_active_step_if_needed` (`:294`, before this call at `:312`) can already have set the procurement action to `AWARDED`, meaning `agreement.is_awarded` may be `True` at the moment the name is assigned. Nothing blocks it now, but if this is ever refactored to route through `AgreementService.update()` it would start getting rejected — a test would catch that regression.

frontend/src/components/Agreements/AwardRequestForm/AwardRequestForm.jsx
```diff
@@ -159,6 +175,58 @@ const AwardRequestForm = ({
...
+                                {validationResult.getErrors("agreementTitle")?.length > 0 && (
+                                    <div
+                                        className="usa-error-message"
+                                        role="alert"
+                                    >
+                                        {validationResult.getErrors("agreementTitle")[0]}
+                                    </div>
+                                )}
+                                <input
+                                    id="agreementTitle"
```
weimiao67, line 211: **This field bypasses the 200-character cap that's enforced everywhere else on agreement titles.**

Since this value is written straight into `agreement.name` on approval, it should carry the same guardrail the agreement editor has: `AgreementEditForm.jsx:317` sets `maxLength={200}` on the `name` input (also labeled "Agreement Title").

Currently there's no cap at any layer:
- no `maxLength` on this input
- `RequestAwardApproval.suite.js:35-37` has only `isNotEmpty()` — note `purchaseOrderNumber` / `taskOrderNumber` both correctly got `shorterThanOrEquals(100)`
- `schemas/procurement_tracker_steps.py:487` `agreement_title` has no `validate.Length` (the other three do)
- both DB columns are unbounded `String`

Net effect: step 6 can silently produce an agreement name that the agreement edit form itself could never have created. Suggest adding the 200 cap at all three layers for consistency.

frontend/src/pages/agreements/award-approval/RequestAwardApproval.suite.js
```diff
@@ -31,6 +31,36 @@ const suite = create((data = {}, fieldName) => {
         enforce(data.vendor).isNotNullish().greaterThan(0);
     });
 
+    // OPS-5892: Agreement Title is required (overwrites agreement.name on approval)
+    test("agreementTitle", "Agreement Title is required", () => {
```
weimiao67, line 35: **These six new Vest rules aren't asserted by any test.**

`RequestAwardApproval.hooks.test.js:34-44` mocks this suite wholesale (`getErrors: () => []`, `hasErrors: () => false`), and there's no `RequestAwardApproval.suite.test.js` — only `ApproveAwardApproval.suite.test.js`, which covers just `obligatedDate`. So the four "is required" rules and both `shorterThanOrEquals(100)` caps have zero coverage. The green suite is a bit misleading here.

A small `RequestAwardApproval.suite.test.js` exercising each key would cover it cheaply.

One good-news note: I traced the consumer chain and these keys are **not** read by any error-summary that maps through `convertCodeForDisplay("validation", key)` — the form renders each key directly. So no entries are needed in the mapping table at `helpers/utils.js:134-158`. Correct as-is.

frontend/src/pages/agreements/award-approval/EditAwardApproval.hooks.js
```diff
@@ -159,6 +165,12 @@ export default function useEditAwardApproval(agreementId) {
         if (step6.award_date) setAwardDate(formatApiDateForDisplay(step6.award_date));
         if (step6.requestor_notes) setNotes(step6.requestor_notes);
 
+        // OPS-5892: seed additional award fields (fall back to agreement.name for legacy rows)
+        setAgreementTitle(step6.agreement_title ?? agreement.name ?? "");
+        if (step6.modification_number) setModificationNumber(step6.modification_number);
+        if (step6.purchase_order_number) setPurchaseOrderNumber(step6.purchase_order_number);
+        if (step6.task_order_number) setTaskOrderNumber(step6.task_order_number);
```
weimiao67: **The entire Edit path is untested** — there's no `EditAwardApproval.hooks.test.js` and no `EditAwardApproval.test.jsx` in this directory.

The Request path got solid new coverage (seeding, blocker, PATCH payload), but the equivalent Edit logic added here is unverified:
- seeding the four new fields from `step6`, including the `?? agreement.name` fallback for legacy rows (`:169`)
- the four new `hasChanged` comparisons against seeded values (`:200-210`)
- the new fields in the PATCH payload (`:334-338`)

The re-display-on-edit behavior is step 5 of your own "How to test" list, so it seems worth a test. Mirroring the existing Request-path tests should be mostly copy-adapt.

frontend/src/components/Agreements/AwardRequestForm/AwardRequestForm.jsx
```diff
@@ -313,6 +381,117 @@ const AwardRequestForm = ({
...
+                    <div className="grid-row grid-gap flex-align-end">
+                        <div className="grid-col-4">
+                            <label
+                                className="usa-label"
+                                htmlFor="modificationNumber"
+                            >
+                                Modification #
+                            </label>
+                            <select
+                                id="modificationNumber"
```
weimiao67: **This select breaks the error-display pattern used by every other field in this form, and its error state is unreachable.**

Compare with `contractNumber` / `purchaseOrderNumber` / `taskOrderNumber` in this same file, which all:
- wrap in `usa-form-group` + conditional `usa-form-group--error`
- apply `usa-label--error` to the label and `usa-input--error` to the control
- render the error message **before** the control

This block does none of those and renders its error **after** the select.

Secondly, that error block looks like dead UI: the select has no `<option value="">` placeholder (contrast the Vendor select at `:267`, which does), and `modificationNumber` defaults to `"Base"` in both hooks and is never cleared — so `enforce(data.modificationNumber).isNotEmpty()` can't fail from the UI. Either wire the field into the standard form-group pattern or drop the error rendering.

Related process note: the PR checks **"No accessibility-impacting changes"** while adding four new required form controls including a select. Worth unchecking. (I'd originally flagged the missing `aria-describedby` here too, but the existing `contractNumber` field doesn't use it either — that's this file's established convention, not a regression from this PR.)

frontend/src/components/Agreements/AwardRequestForm/awardForm.helpers.js
```diff
@@ -22,3 +22,22 @@ export const formatVendorType = (vendorType) => {
 
     return typeMap[cleanType] || vendorType;
 };
+
+/**
+ * Default Modification # value for a new award.
+ * @type {string}
+ */
+export const DEFAULT_MODIFICATION_NUMBER = "Base";
```
weimiao67, line 30: `DEFAULT_MODIFICATION_NUMBER` is exported but only ever referenced inside this file (by `getModificationOptions` below). Meanwhile `"Base"` is hardcoded in four places in the hooks:

- `RequestAwardApproval.hooks.js:53` (`useState`) and `:155` (`hasChanged`)
- `EditAwardApproval.hooks.js:75` (`useState`) and `:200` (`seededMod`)

Importing the constant in both hooks would keep the default in one place — the `hasChanged` comparisons in particular silently break if the default ever changes.

frontend/src/components/Agreements/AwardRequestForm/AwardRequestForm.jsx
```diff
@@ -159,6 +175,58 @@ const AwardRequestForm = ({
...
+            {/* Update Agreement Title (OPS-5892) */}
+            <Accordion
+                heading="Update Agreement Title"
+                level={3}
+                isClosed={false}
+                dataCy="update-agreement-title-accordion"
+            >
+                <fieldset className="usa-fieldset">
+                    <p className="margin-top-1 margin-bottom-3">
+                        Enter the Agreement Title to match the signed award exactly. This will over write the agreement
```
weimiao67: Typo in user-facing copy: **"over write"** → **"overwrite"**.

Worth noting the sibling copy on the approval screen spells it correctly ("this title will overwrite the current agreement title" — `ApproveAwardApproval.jsx:247`), so the two are inconsistent as written. If this was transcribed faithfully from the Figma and the design has the typo, might be worth fixing in the design too so it doesn't come back.

frontend/src/components/Agreements/AwardRequestForm/AwardRequestForm.jsx
```diff
@@ -313,6 +381,117 @@ const AwardRequestForm = ({
...
+                                value={modificationNumber}
+                                onChange={(e) => {
+                                    onModificationNumberChange(e.target.value);
+                                    runValidate("modificationNumber", e.target.value);
+                                }}
+                                required
+                                aria-required="true"
+                                data-cy="modification-number-select"
+                            >
+                                {getModificationOptions().map((option) => (
```
weimiao67: Minor: `getModificationOptions()` rebuilds the same 21-element array on every render of this form. Since the list is static, a module-level constant in `awardForm.helpers.js` (e.g. `export const MODIFICATION_OPTIONS = [...]`, with the function kept only if something else needs it) would avoid the churn and read a bit cleaner at the call site.

frontend/src/pages/agreements/award-approval/ApproveAwardApproval.test.jsx
```diff
@@ -173,7 +173,8 @@ describe("ApproveAwardApproval", () => {
 
     it("renders the agreement name as subtitle", () => {
         renderPage();
-        expect(screen.getByText("Agreement Name B")).toBeInTheDocument();
+        // The name appears in the PageHeader subtitle and in the "Current Agreement Title" review row (OPS-5892).
+        expect(screen.getAllByText("Agreement Name B").length).toBeGreaterThan(0);
```
weimiao67: Nit: changing `getByText` → `getAllByText(...).length > 0` fixes the ambiguity but weakens the assertion — it no longer verifies the name renders *in the PageHeader subtitle* specifically, and would still pass if the subtitle disappeared entirely as long as the new "Current Agreement Title" row rendered.

Scoping to the intended container keeps the original meaning, e.g. `within(screen.getByTestId(...)).getByText("Agreement Name B")`, or assert the exact expected count.

frontend/src/pages/agreements/award-approval/EditAwardApproval.hooks.js
weimiao67: Added `EditAwardApproval.hooks.test.js` (17 tests) in abd69cb — covers seeding all four fields from step 6, the `agreement.name` / `Base` fallbacks for legacy rows, `hasChanged` false on a pristine seeded page (including the fallback-seeded case) and true per edited field, whitespace-only edits as no-ops, the four new fields in the PATCH payload, that the payload carries no `approval_requested`/`approval_status`, the missing-step-6 guard, and save-failure handling. Also added Budget Team / non-Budget-Team permission cases.

frontend/src/components/Agreements/AwardRequestForm/AwardRequestForm.jsx
weimiao67: Wired the Modification # select into the standard pattern: `usa-form-group` + conditional `usa-form-group--error`, `usa-label--error` on the label, `usa-input--error` on the select, and the `role="alert"` message rendered *before* the control. Kept the error rendering rather than dropping it so the field behaves like its siblings if the default ever stops being pre-filled; added a test asserting all four properties including DOM order. Also unchecked "No accessibility-impacting changes" in the PR description and described the a11y treatment of the four new controls.

frontend/src/components/Agreements/AwardRequestForm/awardForm.helpers.js
weimiao67: Both hooks now import `DEFAULT_MODIFICATION_NUMBER` from `awardForm.helpers` — no more hardcoded `"Base"`. The `useState` initializers and the `hasChanged`/seeding comparisons all go through `getSeededAwardFields` / `hasAwardFieldChanges`, so the default lives in one place.

frontend/src/components/Agreements/AwardRequestForm/AwardRequestForm.jsx
weimiao67: Fixed — now reads "overwrite", matching the approval screen copy. Flagged the typo for the design file too.

frontend/src/components/Agreements/AwardRequestForm/AwardRequestForm.jsx
weimiao67: Done — the list is now a frozen module-level `MODIFICATION_NUMBER_OPTIONS` constant in `awardForm.helpers.js`; `getModificationOptions()` is gone.

frontend/src/pages/agreements/award-approval/ApproveAwardApproval.test.jsx
weimiao67: Scoped to the PageHeader: `within(screen.getByTestId("page-header")).getByText("Agreement Name B")`. It now fails if the subtitle disappears, regardless of the new "Current Agreement Title" row.

rajohnson90:
Looks like Wei fixed all the things I identified so I'm going to approve

rajohnson90 (approved):

weimiao67 (approved):

PR #6232, opened 2026-09-11
fix: hide Edit button entirely for read-only users on agreement pages

Santi-3rd:
## What changed

Read-only users no longer see an Edit button on the agreement Details or Budget Lines tabs. Previously they saw it rendered disabled with a "Only team members can edit this agreement." tooltip — misleading for a role that can never edit anything. All other non-editable cases keep the existing visible-but-disabled button.

- **`AgreementDetailHeader.jsx`** — new `showEditButton` prop (default `true`) gating both the enabled and disabled render branches, mirroring the seam `AgreementBudgetLinesHeader` already had.
- **`AgreementDetails.jsx`**, **`AgreementBudgetLines.jsx`** — pass `showEditButton={!isReadOnly}` via `useIsUserReadOnly()`.

The role needs its own gate because `_meta.isEditable` comes from team-member association, not role. It can be `true` for a read-only user on the agreement's team, in which case the old code rendered a **fully clickable** Edit button — a second, previously unreported hole this closes. The role lookup sits in the page components rather than the headers because both header test files render without a Redux `Provider`.

Also fixed a pre-existing false-passing test in `AgreementBudgetLines.test.jsx`: it queried `{ name: /^edit$/i }`, but the enabled button's icon carries `title="edit"`, making the accessible name `"editEdit"` — the anchored regex matched nothing, so the assertion passed even with the button on screen.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6178

## How to test

Automated:

```bash
cd frontend
bunx vitest run \
  src/components/Agreements/AgreementDetailHeader.test.jsx \
  src/pages/agreements/details/AgreementBudgetLines.test.jsx \
  src/pages/agreements/details/AgreementDetails.test.js
```

All 56 tests should pass.

Manual — the fix (button absent for read-only):

1. `docker compose up --build`
2. Log in via FakeAuth as **Read Only User** (Randy, user 529).
3. Visit `/agreements/9` (Details tab). Confirm **no** Edit button and **no** "Only team members can edit this agreement." tooltip next to the "Agreement Details" heading.
4. Visit `/agreements/9/budget-lines`. Confirm the same — no Edit button in the header row alongside the Include Drafts toggle.

Manual — no regression (button still visible-but-disabled for everyone else):

5. Log in as **System Owner** and visit an agreement they are not a team member of. The Edit button should still render, disabled, with its tooltip.
6. Visit an agreement they *can* edit and confirm the Edit button is present and clickable.

Note: agreement 9's team members are users 511 and 520 with project officer 503, so it exercises the read-only *non-team-member* path. To exercise the read-only *team-member* path (the clickable-button hole), the unit tests cover it directly via `_meta.isEditable: true`.

## A11y impact

## Storybook

## Screenshots

_TODO: add before/after of the Details and Budget Lines headers as a read-only user._

## Links

No E2E coverage is included in this PR — verification is unit-test-only.

Two related gaps were found while investigating and are **deliberately out of scope** here; worth separate tickets:

- `frontend/src/components/Agreements/AgreementProcurementTracker.jsx:73` gates on `isSuperUser || _meta.isEditable` with no read-only role check.
- `frontend/src/pages/agreements/details/AgreementBudgetLines.jsx:296` renders the edit wizard on `isEditMode` alone, which is reachable via `?mode=edit` (set by `Agreement.jsx:59`) regardless of role.

### DIFF
--- a/frontend/src/pages/agreements/details/AgreementDetails.jsx
+++ b/frontend/src/pages/agreements/details/AgreementDetails.jsx
@@ -1,5 +1,5 @@
 import AgreementDetailHeader from "../../../components/Agreements/AgreementDetailHeader";
-import { useIsUserSuperUser } from "../../../hooks/user.hooks";
+import { useIsUserSuperUser, useIsUserReadOnly } from "../../../hooks/user.hooks";
 import { AgreementType } from "../agreements.constants";
 import AgreementDetailsEdit from "./AgreementDetailsEdit";
 import AgreementDetailsView from "./AgreementDetailsView";
@@ -36,6 +36,10 @@ const AgreementDetails = ({
     isPostPreAwardLocked = false
 }) => {
     const isSuperUser = useIsUserSuperUser();
+    // Read-only users can never edit any agreement, so the Edit button is hidden outright rather
+    // than shown disabled. `_meta.isEditable` reflects team-member association, not role, so it can
+    // be true for a read-only user who happens to be on the team — the role check has to be its own gate.
+    const isReadOnly = useIsUserReadOnly();
     const isGrant = agreement?.agreement_type === AgreementType.GRANT;
     const grantNumbers = isGrant ? (agreement?.grant_numbers ?? []) : [];
 
@@ -79,6 +83,7 @@ const AgreementDetails = ({
                 isPreAwardInReview={isPreAwardInReview}
                 isAwardInReview={isAwardInReview}
                 isPostPreAwardLocked={isPostPreAwardLocked}
+                showEditButton={!isReadOnly}
             />
 
             {isEditMode && isEditable ? (
--- a/frontend/src/components/Agreements/AgreementDetailHeader.jsx
+++ b/frontend/src/components/Agreements/AgreementDetailHeader.jsx
@@ -18,6 +18,9 @@ import DisabledEditButton from "./DisabledEditButton";
  * @param {boolean} [props.isPreAwardInReview] - Whether pre-award approval is in review.
  * @param {boolean} [props.isAwardInReview] - Whether award approval is in review.
  * @param {boolean} [props.isPostPreAwardLocked] - Whether the agreement is permanently locked after full pre-award approval.
+ * @param {boolean} [props.showEditButton] - Whether to render the Edit button at all. Set to false for
+ *   read-only users, who can never edit any agreement — for them neither the enabled button nor the
+ *   disabled button-plus-tooltip is meaningful, since there is no state in which they could edit.
  * @returns {JSX.Element} - The rendered component.
  */
 export const AgreementDetailHeader = ({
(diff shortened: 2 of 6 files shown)


frontend/src/pages/agreements/details/AgreementDetails.jsx
```diff
@@ -79,6 +83,7 @@ const AgreementDetails = ({
                 isPreAwardInReview={isPreAwardInReview}
                 isAwardInReview={isAwardInReview}
                 isPostPreAwardLocked={isPostPreAwardLocked}
+                showEditButton={!isReadOnly}
             />
 
             {isEditMode && isEditable ? (
```
weimiao67, line 96: `isEditMode && isEditable` still renders the live `AgreementDetailsEdit` form for a read-only team member, since `isEditable` (lines 66-70) never checks `isReadOnly`.

A read-only user on an agreement's team (`_meta.isEditable === true`, granted purely by team association per `check_user_association` on the backend — role-independent) can visit `/agreements/9?mode=edit`. `Agreement.jsx` sets `isEditMode=true` from the URL param with no role check, `isEditable" here evaluates true, and the live edit form renders — even though this PR just hid the Edit button for this exact user. This PR's new tests only exercise `isReadOnly` with `isEditMode={false}`, so this path is untested. Suggest gating `isEditable` (or the `isEditMode && isEditable` render) on `!isReadOnly` too (with the `isSuperUser` override preserved).

frontend/src/pages/agreements/details/AgreementDetails.jsx
```diff
@@ -79,6 +83,7 @@ const AgreementDetails = ({
                 isPreAwardInReview={isPreAwardInReview}
                 isAwardInReview={isAwardInReview}
                 isPostPreAwardLocked={isPostPreAwardLocked}
+                showEditButton={!isReadOnly}
```
weimiao67: `showEditButton={!isReadOnly}` omits the `isSuperUser` override that every other editability check in this PR's touched files includes (e.g. `isEditable` above at line 70, and `isAgreementEditable` in `AgreementBudgetLines.jsx:98`, both `isSuperUser || ...`).

`is_superuser` is derived from `any(role.name == "SUPER_USER" for role in self.roles)` (`backend/models/users.py:142-143`), and `roles" is a many-to-many list with no constraint preventing a user from also holding `READ_ONLY` in the same list. For such a user, `isEditable`/`isAgreementEditable` would correctly resolve `true` via the superuser override, but `showEditButton` would resolve `false`, hiding the button inconsistently with the rest of the logic. Suggest `showEditButton={isSuperUser || !isReadOnly}` here and in `AgreementBudgetLines.jsx:224`.

weimiao67 (approved):

rajohnson90 (approved):

Issue #5658, opened 2026-05-12
Error when attempting to delete an agreement


npage123:
When signed in as a viewer/editor, I get an error when attempting to delete an agreement I created even after adding myself as a team member.

## Expected Behavior
If user is a team member on an agreement with all draft BLs, or no BLs, they should be able to delete the agreement.

## Current Behavior
When signed in as a viewer/editor, I get an error when attempting to delete an agreement I created even after adding myself as a team member (see attached recording for more info).

## Steps to Reproduce
1. Log in to staging as viewer/editor role
2. Create an agreement and add 'User Demo' as a team member
3. Attempt to delete the agreement just created
4. View error

https://github.com/user-attachments/assets/39071b47-a212-4a5f-9b55-866ddb1f9842

npage123:
Questions to answer...Is the current ability to delete based on user role following business rules that are currently in place? If so, what are they?

npage123:
Moving to blocked as we are needing to get feedback from Sheila and team on business rules below for when deleting requires an approval and when not. Our current understanding of the rules should be:

1. Agreements with all draft BLs or no BLs can be deleted by team members
2. Agreements with planned or executing BLs need approval
3. Agreements with obligated BLs cannot be deleted

kimschulke:
[Updated OPS business rules](https://docs.google.com/document/d/165NZTGL3yreqhDzzhAwcr0Hg-eWf1fN3mUfSI0zG0Uw/edit?tab=t.0#heading=h.srfro3m3ouj) to review with OPRE

<img width="769" height="510" alt="Image" src="https://github.com/user-attachments/assets/7874fd00-22bc-4bdb-bfe5-14332f7344a6" />

jonnalley:
~I think in the specific example mentioned, this has to do with the SCs getting deleted.  that's my suspicion based on what I saw earlier at least.~

jonnalley:
@weimiao67 #5832 may resolve this. we'll see!

jonnalley:
issue still present and seen in dev - seen on 09/09 at 0128 UTC

jonnalley:
@npage123 the fix for this is deployed out to dev and staging, feel free to review!

npage123:
@rajohnson90 demoed during collab time today and we verified that user demo was able to successfully delete a created agreement.

PR #6237, opened 2026-09-14
Close agreement delete permission gap for non-admin roles

jonnalley:
## What changed

Fixes a route-level permission gap where VIEWER_EDITOR, REVIEWER_APPROVER, and BUDGET_TEAM users got a bare 403 deleting agreements they were legitimately authorized to delete via team membership — the service-layer authorization was already correct, only the decorator's static role→permission table was missing the grant. Also adds a new server-side guard so opening that gate doesn't let more roles delete agreements with non-draft budget lines, and makes the frontend defer entirely to the backend for this decision instead of re-deriving it client-side. Includes a follow-up remediation pass addressing review feedback (a tooltip-precedence bug, redundant computation, code duplication, test gaps).

**Permissions** (`backend/alembic/versions/…add_delete_agreement_permission.py`, `003-role.sql`, `user_data.json5`)
New migration grants `DELETE_AGREEMENT` to `VIEWER_EDITOR`/`REVIEWER_APPROVER`/`BUDGET_TEAM`, idempotent and downgradable, mirroring the existing `add_perms_for_reviewer_approver` migration. `003-role.sql` and `user_data.json5` updated to match — the latter is what actually seeds the pytest/local-dev database (discovered mid-implementation that role permissions live in three independent seed sources that can drift out of sync).

**Server-side deletability guard** (`backend/ops_api/ops/services/agreements.py`, `backend/models/agreements.py`)
`AgreementsService.delete()` now blocks deleting an agreement with any non-Draft budget line unless the caller is a super user — closing the gap that opening the permission to more roles would otherwise create. `_is_deletable`/`_get_locked_message` mirror the existing `_is_editable`, computed once per request (previously called 3x per agreement) and backed by a new `Agreement.has_non_draft_budget_lines` model property (previously duplicated inline in `award_type` and the service).

**API response** (`backend/ops_api/ops/resources/agreements.py`, `backend/openapi.yml`)
`_meta.isDeletable`/`lockedMessage` are now populated for agreements (schema fields already existed, unpopulated). Corrected the OpenAPI description text, which was copy-pasted from the budget-line-item version.

**Frontend** (`AgreementTableRow.jsx`, `AgreementsTable.helpers.js`, `AgreementTypes.d.ts`)
The delete button now reads `_meta.isDeletable` from the backend instead of re-deriving it client-side from budget-line statuses — this also fixes a latent bug where a super user's delete button never respected the awarded-agreement block. The locked-tooltip logic was extracted into a testable `getAgreementLockedMessage` helper with corrected precedence (frontend-owned "not a team member"/"not developed yet" reasons now take priority over the backend's delete-specific message, since both icons share one tooltip). The `Agreement` type now declares `isDeletable`/`lockedMessage`.

**Tests**
New/extended coverage in `test_agreements.py` (service-layer `_is_deletable`/`_get_locked_message`, including a NULL-status budget-line case), `test_agreement.py` (real-endpoint regression tests for the basic-role success path, the non-draft-BLI block, and the super-user bypass), `test_roles.py` (migration upgrade/downgrade/idempotency, mirroring the existing pattern used for the reviewer-approver migration), and frontend helper/component tests.

## Issue

https://github.com/HHS/OPRE-OPS/issues/5658

## How to test

1. Backend: `cd backend/ops_api && pipenv run pytest` — full suite passes (2285 passed, 9 skipped as of this branch).
2. Frontend: `cd frontend && bun run test --watch=false` — full suite passes (4460 passed).
3. Manual (Docker stack via `docker compose up --build`):
   - Log in via FakeAuth as a Viewer/Editor user, create a Contract agreement, add yourself as a team member, leave budget lines empty or Draft.
   - Confirm the delete icon is enabled and deleting succeeds (previously 403'd).
   - Add a Planned budget line to a similar agreement — confirm the delete icon is now disabled, with a tooltip reading "Cannot delete an agreement with budget lines that are not in Draft status" (not a generic "Disabled").
   - As a super user, confirm the delete icon on an *awarded* agreement is disabled (previously always enabled regardless of award status).

## A11y impact

## Storybook

- [x] N/A — change is page-specific (Agreements table row) and non-visual (button enabled/disabled state + tooltip text only)

## Screenshots

N/A — changes are behavioral (icon disabled state, tooltip text) rather than visual; no layout changes to capture.

## Links

N/A


weimiao67 (commented):
Reviewed the permission/guard changes and the frontend delete-state logic. One correctness issue that looks like it'll break an existing E2E spec (posted as a separate PR comment since that file isn't touched in this diff), plus two minor code-quality notes inline below.

backend/ops_api/ops/services/agreements.py
```diff
@@ -859,6 +865,39 @@ def _is_editable(self, agreement: Agreement, user: User) -> bool:
         """
         return user.is_superuser or associated_with_agreement(agreement.id)
 
+    def _is_deletable(self, agreement: Agreement, user: User, is_editable: bool | None = None) -> bool:
```
weimiao67: `_is_deletable` doesn't appear to have a production call site — `isDeletable` in the actual API response is derived from `_get_locked_message`'s result (`locked_message is None`, in `resources/agreements.py`), not from calling `_is_deletable`.

The docstring says this method "mirrors" `_get_locked_message` so the two "can never drift apart," but since only `_get_locked_message` is wired to the real endpoint, a future edit to one predicate's logic/ordering could silently diverge from the other without a test against the real response catching it. Worth either wiring `_is_deletable` into the resource layer or dropping it and keeping just `_get_locked_message`.

frontend/src/components/Agreements/AgreementsTable/AgreementTableRow.jsx
```diff
@@ -53,38 +51,17 @@ export const AgreementTableRow = ({ agreement }) => {
     const awardType = AWARD_TYPE_LABELS[agreement?.award_type] ?? NO_DATA;
     const vendor = isSuccess ? (agreement?.vendor ?? NO_DATA) : NO_DATA;
 
-    const areAllBudgetLinesInDraftStatus = isSuccess ? areAllBudgetLinesInStatus(agreement, BLI_STATUS.DRAFT) : false;
     const isSuperUser = useSelector((state) => state.auth?.activeUser?.is_superuser) ?? false;
     const isReadOnly = useIsUserReadOnly();
 
     const canUserEditAgreement = isSuccess && agreement?._meta.isEditable;
-    const areThereAnyBudgetLines = isSuccess ? isThereAnyBudgetLines(agreement) : false;
     const isAgreementTypeNotDeveloped = isSuccess && isNotDevelopedYet(agreement?.agreement_type ?? "");
```
weimiao67, line 58: Minor: `isNotDevelopedYet(agreement?.agreement_type ?? "")` is computed here for `isEditable`, and again independently inside `getAgreementLockedMessage` (AgreementsTable.helpers.js) for the tooltip. Not a bug today since both use the same input/function, but if one call site gets special-cased later without the other, the disabled-icon state and its tooltip could disagree. Might be worth computing once and passing it in.

weimiao67:
Possible test regression: `frontend/cypress/e2e/approveChangeRequestsAtAgreementLevel.cy.js` (line ~250) isn't touched in this diff, but it looks like it hits the same failure mode this PR already fixed elsewhere.

The spec runs as `budget-team` (roles `USER_ADMIN` + `BUDGET_TEAM`, not `SUPER_USER`). It approves a Draft→Planned CR for `bliId`, then DELETEs that BLI (which only creates a deletion change request — 202, BLI stays `PLANNED`), then DELETEs the agreement itself expecting `200`.

`AgreementsService.delete` (`backend/ops_api/ops/services/agreements.py:508-512`) now raises a 400 for non-superusers when `agreement.has_non_draft_budget_lines` is true, so that final assertion should fail here.

This PR already fixed the identical pattern in `reviewChangeRequestResponse.cy.js` and `changeRequestsPagination.cy.js` (switching cleanup to `power-user`, with a comment referencing #5658) — this file looks like it was missed. `budgetChangeRequest.cy.js` (~lines 144, 305), `declineChangeRequestsAtAgreementLevel.cy.js`, and `reviewChangeRequestsAtCardLevel.cy.js` look like they may have the same issue and might be worth a check too.

backend/ops_api/ops/services/agreements.py
jonnalley: I'll address but it is a bit of scope creep.

jonnalley:
Thanks for the thorough catch, @weimiao67 — confirmed and fixed.

**Cleanup-delete regression:** traced the pattern across the whole Cypress suite and found it in 7 files total (your 4 flagged files were all correct, plus `approveCrossDivisionCRs.cy.js` which CI caught independently). All now switch their cleanup agreement-DELETE to a `power-user` (superuser) token, mirroring the fix already applied to `reviewChangeRequestResponse.cy.js`/`changeRequestsPagination.cy.js`:
- `approveChangeRequestsAtAgreementLevel.cy.js`
- `approveCrossDivisionCRs.cy.js`
- `budgetChangeRequest.cy.js`
- `declineChangeRequestsAtAgreementLevel.cy.js`
- `reviewChangeRequestsAtCardLevel.cy.js`
- `procurementShopChangeRequest.cy.js`
- `saveChangesToEdits.cy.js`

**Inline nits:** both addressed.
- Dropped `_is_deletable` (dead code, no production call site) and rewrote its tests against `_get_locked_message` directly, so there's only one predicate left to drift.
- `isNotDevelopedYet` is now computed once in `AgreementTableRow.jsx` and passed into `getAgreementLockedMessage` instead of being recomputed internally.

**One thing I did *not* fix here:** `approveChangeRequestsAtAgreementLevel.cy.js`'s "review Budget Change change" test still fails locally on an unrelated assertion (a pending-CR count in the `_tabsList` nav element). I confirmed this is a pre-existing flake — the same spec is independently failing on `main` right now with a different assertion in that same element — so it predates this branch and I left it out of scope.

Full backend suite (2285 passed) and frontend suite (4460 passed) are green, and all 9 touched Cypress specs were run locally to confirm (8/9 fully green; the 1 failure is the pre-existing `main` flake above).

weimiao67 (approved):

Santi-3rd:
During manual testing, when a BLI was "In Review" because it's going from Draft to Planned status, the delete icon was still enabled and I was able to delete the agreement using the Viewer/editor role. If this is the intended functionality let me know and I'll approve.

jonnalley:
@Santi-3rd that's definitely a corner case, but would technically be allowed with the business rules that we have. Until it is approved, it's still in draft. So presuming that the agreement in question only had draft budget lines, including the in review that you're talking about, and the user was that was logged in was authorized to be doing that, then that would have been the expected behavior. Thank you, great testing.

Santi-3rd (approved):

josbell (changes requested):
Solid implementation overall — the three-source sync, the service-layer guard, and the meta-driven frontend are all well-structured. One confirmed functional bug needs a fix before merge, plus two lower-priority hardening items.

**Confirmed bug (blocks merge):**
`ChangeIcons.jsx` ignores `isItemDeletable` in the `!isItemEditable` branch — the delete button is unconditionally `disabled={true}` there regardless of the prop. A non-superuser on a "not-developed-yet" agreement type gets `isEditable=false` but `isDeletable=true` from the backend, so the UI silently re-locks the button this PR was meant to enable.

**Hardening (non-blocking but worth fixing):**
- `delete()` and `_get_locked_message()` are independently maintained copies of the same two guards. The docstring claims they "can never drift apart" but nothing enforces it. Consider having `delete()` call `_get_locked_message()` and raise if it returns a non-None value.
- `agreement?._meta.isEditable` / `agreement?._meta.isDeletable` crash if `_meta` itself is null — the optional chain only guards `agreement`. Under the current API this cannot happen, but `agreement?._meta?.isEditable` / `agreement?._meta?.isDeletable ?? false` would be safer.

frontend/src/components/Agreements/AgreementsTable/AgreementTableRow.jsx
```diff
@@ -53,38 +51,19 @@ export const AgreementTableRow = ({ agreement }) => {
...
     const isEditable = canUserEditAgreement && (!isAgreementTypeNotDeveloped || isSuperUser);
-    const canUserDeleteAgreement =
-        isSuperUser || (canUserEditAgreement && (areAllBudgetLinesInDraftStatus || !areThereAnyBudgetLines));
+    const canUserDeleteAgreement = isSuccess && (agreement?._meta.isDeletable ?? false);
```
josbell: The optional chain only guards `agreement`, not `_meta`. If `_meta` is ever null/undefined (stale cache, partial serialization failure), both lines 57 and 60 throw a `TypeError` on the `.isEditable`/`.isDeletable` access.

Under the current API `_meta` is always populated, so this cannot happen today — but the defensive form is cheap:

```suggestion
    const canUserDeleteAgreement = isSuccess && (agreement?._meta?.isDeletable ?? false);
```

And line 57 similarly: `agreement?._meta?.isEditable`.

frontend/src/components/Agreements/AgreementsTable/AgreementsTable.helpers.js
```diff
@@ -52,3 +40,32 @@ export const getProcurementShopDisplay = (agreement) => {
...
+    if (!canUserEditAgreement) {
+        return "Only team members on this agreement can edit or delete";
+    }
+    if (isAgreementTypeNotDeveloped && !isSuperUser) {
+        return "This agreement cannot be edited because it is not developed yet, \nplease contact the Budget Team.";
+    }
+    if (agreement._meta?.lockedMessage) {
+        return agreement._meta.lockedMessage;
+    }
+    return isSuperUser ? "" : "Disabled";
```
josbell, line 70: This helper is correct, but there's a bug one level up in `ChangeIcons.jsx` that will silently negate its effect in certain cases.

In `ChangeIcons.jsx`, the `!isItemEditable` branch (line 102+) renders the delete button as `disabled={true}` **unconditionally** — `isItemDeletable` is never read in that branch (contrast with the `isItemEditable` branch at line 81 which correctly uses `disabled={!isItemDeletable}`).

Concrete failure: a non-superuser team member on a `"not developed yet"` agreement type gets `isEditable=false` (frontend type guard) but `isDeletable=true` from `_meta`. `AgreementTableRow` passes both to `ChangeIcons`. `ChangeIcons` takes the `!isItemEditable` branch and locks the delete button with `disabled={true}`, regardless of the `isDeletable` prop. The user sees a locked icon; the backend would allow the `DELETE`.

Fix in `ChangeIcons.jsx` line 137:
```js
// before
disabled={true}
// after
disabled={!isItemDeletable}
```

backend/ops_api/ops/services/agreements.py
```diff
@@ -505,6 +505,12 @@ def delete(self, id: int) -> None:
         if agreement.is_awarded:
             raise ValidationError({"is_awarded": ["Cannot delete an awarded agreement."]})
 
+        user = get_current_user()
+        if not user.is_superuser and agreement.has_non_draft_budget_lines:
```
josbell: These two guards are manually kept in sync with `_get_locked_message()` (lines 879–882), which has the identical conditions. The `_get_locked_message` docstring says the two "can never drift apart" — but that's only true if someone remembers to update both places.

A future guard (e.g. "cannot delete while a change request is pending") added to only one side would either silently allow a delete the UI has blocked, or silently lock a delete the endpoint would allow. Consider delegating here:

```python
locked = self._get_locked_message(agreement, user)
if locked:
    raise ValidationError({"agreement": [locked]})
```

This makes the endpoint and the meta field provably identical by construction.

jonnalley:
@josbell Confirmed and fixed all three. Please re-review or remove requested change bit.

**ChangeIcons.jsx delete button** — real bug, verified reachable (non-superuser team member on an IAA/Direct-Obligation agreement with no non-draft BLIs gets `isEditable=false` from the frontend-only "not developed yet" type gate, but `isDeletable=true` from the backend). Fixed by mirroring the `isItemEditable` branch's delete-button pattern in full (conditional `disabled`, `onClick` guard, icon styling, tooltip label) rather than just flipping the `disabled` prop — the literal one-liner would've left a clickable-looking button with no `onClick` handler. Added a regression test covering this exact prop combination.

**Optional chaining** — fixed as suggested (`_meta?.isEditable`, `_meta?.isDeletable`).

**Backend guard duplication** — extracted a shared `_deletion_blocked_reason()` helper that both `delete()` and `_get_locked_message()` branch on, rather than delegating `delete()` to `_get_locked_message()`'s string output directly. Delegating to the string would have collapsed the "not a team member" case into a generic `ValidationError` (400), losing the existing `AuthorizationError` (403) distinction, and changed the field keys/text on the other two cases — a behavior/contract change, not just cleanup. The extracted helper keeps every existing exception type, HTTP status, field key, and message text identical while removing the actual drift risk (a future guard condition now only needs to be added in one place). Full backend suite (2285 passed) and frontend suite (4461 passed) confirm no behavior change.

josbell (approved):

Issue #6230, opened 2026-09-10
Bug on Service Requirement Type for Creating Agreements


kimschulke:
Bug related to the Service Requirement Type component on create-agreement workflows

## Expected Behavior
If Service Requirement Type is required for agreement type:contract, then the user should not be able to proceed into step 3 when the Service Requirement Type is left blank

## Current Behavior
Non-Severable is selected by default under Service Requirement Type, but only on page load… not after the user interacts with the Agreement Type component.

## Steps to Reproduce
1. click create agreement
2. choose project
3. select agreement type: grant
4. enter title
5. click continue
6. add a grant number
7. click back
8. change agreement type from grant to contract
9. click continue
10. see out-of-place validation.

## Context
- unexpected validation outside of our design system

kimschulke:
<img width="1080" height="898" alt="Image" src="https://github.com/user-attachments/assets/7e9cd788-0bcf-44dd-8d0b-bb353f72d509" />

kimschulke:
additionally, another odd scenario

1. click create agreement
2. choose project
3. select agreement type contract
4. enter title
5. select service req. type non several
6. click continue
7. add an SC
8. click back
9. change agreement type to grant, and then back to contract
10. click continue
11. see bug

<img width="1072" height="788" alt="Image" src="https://github.com/user-attachments/assets/0a17a180-028a-4a3a-8221-4acad865a55a" />

weimiao67:
@kimschulke the bug is fixed and ready for review in Staging.

npage123:
@weimiao67 @kimschulke Reviewed in staging and looks good to me.

https://github.com/user-attachments/assets/82c98391-d242-4640-b001-0f809a182158
https://github.com/user-attachments/assets/f744002d-9115-4254-96b5-cfaf522177eb

PR #6239, opened 2026-09-14
fix: restore Service Requirement Type default on agreement type change

weimiao67:
## What changed

Fixes the Service Requirement Type reset bug in the Create Agreement wizard (see Issue below):

- `handleAgreementFilterChange` now restores the Non-Severable default when switching to CONTRACT/DIRECT_OBLIGATION/PARTNER, but only when the value is actually missing (i.e. coming back from GRANT, the only branch that nulls it) — an explicit SEVERABLE choice survives toggling among the other types.
- Services Components are cleared only when switching to GRANT (the only transition that actually invalidates them; grants use `grant_numbers` instead and can't have Services Components added in the first place), via a new `CLEAR_SERVICES_COMPONENTS` reducer action that reconciles any budget line still linked to a cleared component — matching on `services_component_number` (not just `services_component_id`, which pre-save budget lines never carry).
- The Vest rule for `service_requirement_type` now rejects blank/null, but only while creating a new agreement (`isNewAgreement`), so it can't disable Save Changes on legacy agreements that already lack this field.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6230

## How to test

1. Go to Create Agreement, select a Project, continue.
2. Select Agreement Type = Grant, enter a title, click Continue.
3. Click Back, change Agreement Type to Contract.
4. Confirm Service Requirement Type shows "Non-Severable" (not blank) and Continue is enabled.
5. Click Continue — step 3 should show the normal empty Services Component state, not "Please add a Service Requirement Type to the Agreement."

Second repro:
1. Create Agreement, select Agreement Type = Contract, enter a title, click Continue.
2. Add a Services Component on step 3.
3. Click Back, toggle Agreement Type to Grant and then back to Contract.
4. Click Continue — the stale component from before the round trip should be gone, and no fallback error message should appear.

Also covered by the new Cypress spec `cypress/e2e/createAgreementTypeToggle.cy.js`, which reproduces both scenarios above.

## A11y impact

## Storybook

## Screenshots

N/A — bug fix, no visual/UI changes.

## Links

N/A


Santi-3rd (approved):

rajohnson90:
Two things identified by my AI PR review that I'm marking for posterity but plan to fix myself:

1. AgreementEditorContext.hooks.js:132 — CLEAR_SERVICES_COMPONENTS duplicates DELETE_SERVICE_COMPONENT's budget-line reconciliation
     logic (nulling services_component_id/services_component_number/serviceComponentGroupingLabel) instead of sharing it. Future
     changes to that reconciliation shape need to be made in both places.
  2. AgreementEditForm.hooks.js:234 — runValidate's useCallback deps include isAgreementCreated, which is a pure derivation of
     agreement (already a dep) — redundant, no behavior change.

rajohnson90 (approved):

frontend/src/components/Agreements/AgreementEditor/AgreementEditFormSuite.js
```diff
@@ -28,6 +28,12 @@ const suite = create((data = {}, fieldName) => {
     test("service_requirement_type", "This is required information", () => {
         if (isGrant) return;
         enforce(data.service_requirement_type).notEquals("-Select Service Requirement Type-");
+        // Only enforce presence while creating. Some existing non-grant agreements legitimately
+        // have no service_requirement_type; failing here would disable Save Changes on the edit
+        // screens for a field the user never touched. (issue #6230)
+        if (data.isNewAgreement) {
```
josbell, line 34: **Vest v6 `only()` retained-state fragility** (non-blocking, heads-up)

`isNotEmpty()` is correctly gated on `isNewAgreement`, and every caller today goes through `runValidate(name, value)` → `suite.run(data, name)`, so `only(name)` scopes each run safely.

The latent risk: `isNewAgreement` is an informal flag on the domain data bag. Any future `suite.run({ ...agreement, isNewAgreement: true })` that omits the second `fieldName` arg would trigger `isNotEmpty()` when `service_requirement_type` is null, permanently stick the test in a failed state, and block Save for the session — the classic Vest v6 retained-state bug the CLAUDE.md warns about. Fine as-is; just worth keeping in mind if this suite gets a new call site.

frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js
```diff
@@ -721,6 +740,15 @@ const useAgreementEditForm = (
         setFundingPeriodMonths(null);
     };
 
+    // Re-apply the Non-Severable default when switching to a non-grant type. The GRANT branch
+    // nulls service_requirement_type to shape the grant payload, but defaultState's default is
+    // only ever applied as useReducer's initial argument, so nothing restores it on the way
+    // back. Must not be left null: AaAgreementData rejects null (400) and formatServiceComponent
+    // returns undefined for any other value. (issue #6230)
+    const restoreServiceReqTypeDefault = () => {
```
josbell, line 748: **`restoreServiceReqTypeDefault` declared after its call site (TDZ, non-blocking today)**

`handleAgreementFilterChange` calls `restoreServiceReqTypeDefault()` at line 729, but this `const` arrow function isn't declared until line 748. Safe at runtime because the handler only fires as an event handler — all `const` declarations are initialized before any event fires. But if a future change ever invokes the handler synchronously during initialization, this would throw `ReferenceError: Cannot access 'restoreServiceReqTypeDefault' before initialization`. Moving the declaration above the call site is a low-effort fix.

frontend/src/components/Agreements/AgreementEditor/AgreementEditForm.hooks.js
```diff
@@ -220,13 +220,18 @@ const useAgreementEditForm = (
             suite.run(
                 {
                     ...agreement,
+                    // Only enforce service_requirement_type presence while creating (see suite).
+                    // Some existing non-grant agreements legitimately have no
+                    // service_requirement_type; failing here would disable Save Changes on the
+                    // edit screens for a field the user never touched. (issue #6230)
+                    isNewAgreement: !isAgreementCreated,
```
josbell, line 227: **`isNewAgreement` injected as an informal flag on the domain data bag**

This flag is only meaningful to the suite, but it travels via the same object as domain fields. Any future `suite.run` call site that spreads `...agreement` without explicitly adding `isNewAgreement: !isAgreementCreated` silently treats a new agreement as existing and the `isNotEmpty()` guard vanishes — no type error, no lint warning. Not an issue today since there's only one call path, but worth noting if the suite grows more create-vs-edit distinctions.

josbell (approved):
Logic looks correct and well-tested — the GRANT branch clears SCs, the default is restored on the way back, and the validation split between create and edit is sound. Three latent structural notes inline (none are blockers).

PR #6242, opened 2026-09-16
repo skills: add fail-fast sweep guidance and migration rebase mode

jonnalley:
## What changed

Two skill improvements, both learned from real gaps hit:

- **`.claude/skills/ci-log-triage/SKILL.md`**: adds a "Fail-Fast Hides the Rest of the Matrix" section explaining that the E2E workflow's `fail-fast: true` cancels the rest of the Cypress matrix the instant one spec fails — so a single CI run only ever surfaces one failure at a time, even when the same regression affects several spec files. Once a failure is confirmed real (not a flake), the skill now recommends sweeping the rest of the suite for the same pattern before handing off to a fix, instead of discovering siblings one push at a time.

- **`.claude/skills/db-migrations/SKILL.md`**: adds a new `rebase` mode for the case where another PR merges a new Alembic migration to `main` after this branch's own migration was written, and both point at the same `down_revision` — a divergent-heads conflict. The mode walks through: diffing `alembic/versions/` against `origin/main` to find the new file, confirming the shared-parent conflict, safely cherry-picking the upstream migration (if it's pure DDL with no application-code imports) or merging `main` instead, repointing this branch's migration onto the new head, and verifying with a real downgrade/upgrade round-trip — including a warning that Docker images used for seeding must be rebuilt first, since a stale image can produce a false "column doesn't exist" failure that looks like a migration bug but is actually just stale container state.

## Issue

N/A — docs/tooling branch, no ticket.

## How to test

Read through both updated `SKILL.md` files for clarity and correctness. No automated tests apply to skill documentation.

## A11y impact

## Screenshots

N/A — no visual changes.

## Links

N/A


rajohnson90 (approved):

.claude/skills/db-migrations/SKILL.md
```diff
@@ -162,7 +210,7 @@ cd backend
 alembic history -r -10:current --verbose
```
josbell, line 216: Missing `-d` — `docker compose up` without it holds the foreground indefinitely, so the `alembic downgrade`/`upgrade` commands in the next block are unreachable. The existing Upgrade section (line ~34) uses `-d` correctly.

```suggestion
docker compose --profile setup up db data-import --build -d
```

josbell (approved):
Good additions — the fail-fast explanation and the rebase mode fill real gaps. A few issues to address before merge:

**Confirmed bugs:**
- `docker compose up` in Rebase Step 5 is missing `-d`, so the foreground is held and the `alembic downgrade`/`upgrade` commands that follow can never execute.
- `--profile setup` silently activates `disable-users` in addition to `data-import` (both share the `setup` profile in `docker-compose.yml`). The canonical form in CLAUDE.md is `docker compose up db data-import --build` — explicit service names, no profile flag — which avoids that side effect.
- The output contract template in ci-log-triage has no field for sibling sweep results, so an agent following the template will silently drop them.

**Plausible issues worth addressing:**
- The old-parent revision id is never captured before Step 4 overwrites it — by Step 5 the value is gone.
- Steps 3–4 are singular throughout; no guidance for the case where `main` added two migrations since the branch cut.
- The sibling-sweep grep uses a flat glob (`*.cy.js`) rather than a directory argument, so `-r` is a no-op and any future subdirectory specs are silently missed.

.claude/skills/db-migrations/SKILL.md
```diff
@@ -102,7 +102,55 @@ Check for the same issues listed in Step 4 above. Also check:
...
+
+Update this branch's own migration's `down_revision` (and the `Revises:` line in its docstring) to the newly-added file's revision id. Confirm with `alembic heads` — it should report exactly one head, and `alembic history -r <old-parent>:<this-branch's-revision> --verbose` should show a single unbroken chain.
+
+**Step 5: Verify with a real downgrade/upgrade round-trip**
+
+Don't trust `alembic current` alone here — it only reports what the `alembic_version` table says, not what's actually in the schema. And if test data is seeded through a Docker image (e.g. `data-import`), that image has its own baked-in copy of the repo from whenever it was last built — if the new migration file from Step 3 didn't exist yet at build time, the container's own `alembic upgrade head` never touched it, even though the database ends up stamped at head. That produces a very convincing false failure on downgrade (a column reported as missing that should be there) that looks like a real migration bug but is actually just a stale image. Rebuild before testing:
+```bash
+docker compose down
+docker compose --profile setup up db data-import --build
```
josbell: Missing `-d` — without it `docker compose up` holds the foreground, so the `alembic downgrade`/`upgrade` commands in the next block are unreachable. The existing `docker compose up db -d` on line 34 uses the correct form.

```suggestion
docker compose up db data-import --build -d
```

(Also see finding #2 below about `--profile setup`.)

josbell: `--profile setup` also activates the `disable-users` service (it shares the `setup` profile in `docker-compose.yml`), which modifies DB state as a side-effect during what should be a pure migration verification step. CLAUDE.md documents the canonical form as `docker compose up db data-import --build` — explicit service names, no `--profile` flag — which avoids the side-effect.

.claude/skills/db-migrations/SKILL.md
```diff
@@ -102,7 +102,55 @@ Check for the same issues listed in Step 4 above. Also check:
...
+```
+Then walk the chain for real and check the actual schema/data at each step (e.g. `\d <table>` in psql), not just the reported revision:
+```bash
+cd backend
+alembic downgrade <old-parent-revision>
```
josbell, line 153: `<old-parent-revision>` is never explicitly captured. It appears as output from the `grep -m1 "^down_revision"` command in Step 2, but Step 4 then overwrites the branch migration's own `down_revision` — the only file where the value was recorded. By the time Step 5 runs, there's no instruction telling the agent where to find it. Worth adding a "note this value" cue at the end of Step 2 or Step 4.

.claude/skills/db-migrations/SKILL.md
```diff
@@ -102,7 +102,55 @@ Check for the same issues listed in Step 4 above. Also check:
...
+
+Read the new migration before touching anything. If it only calls `op.*`/`sa.*` (pure DDL or raw SQL, no imports of `models` or other application code), it's safe to cherry-pick on its own — a migration that only talks to the database doesn't care what else has or hasn't landed on this branch:
+```bash
+git show origin/main:backend/alembic/versions/<new-file>.py > backend/alembic/versions/<new-file>.py
+```
+If it imports application code, don't cherry-pick a single file into an inconsistent state — merge `main` into the branch instead so the code it depends on comes along with it.
+
+**Step 4: Repoint this branch's migration onto the new head**
+
+Update this branch's own migration's `down_revision` (and the `Revises:` line in its docstring) to the newly-added file's revision id. Confirm with `alembic heads` — it should report exactly one head, and `alembic history -r <old-parent>:<this-branch's-revision> --verbose` should show a single unbroken chain.
```
josbell: Steps 3 and 4 are singular throughout ("the new migration", "the newly-added file's revision id"). If `main` landed two migrations since this branch was cut — forming a chain A→B — there's no guidance on which revision id to use. Repointing to A instead of B leaves two heads and the conflict is unresolved. Worth a short note: "If more than one new file appeared on `main`, copy them in chain order and repoint to the final head."

.claude/skills/ci-log-triage/SKILL.md
```diff
@@ -144,9 +144,21 @@ Then scan for **real failure** markers (see taxonomy).
...
 - **Indeterminate** — neither side matched strongly; link to run and recommend manual log review.
 
+## Fail-Fast Hides the Rest of the Matrix
+
+The E2E Tests workflow runs `fail-fast: true` across its Cypress spec-file matrix: the instant one spec fails, every other in-progress job in that run gets cancelled. A single CI run can only ever surface its *first* failure — if the same regression exists in five spec files, you'll see them one at a time, each only after the previous one is fixed and pushed. That's a slow way to find out something affects more than one file.
+
+So once a failure gets a **Likely real** verdict, don't stop at the one file CI happened to hit first. Decompose the confirmed cause into a search pattern (the same API call, assertion shape, error string, or code pattern that caused it) and grep or read through the rest of the suite for the same shape before handing off to a fix:
+
+```bash
+grep -rl "<pattern that caused the confirmed failure>" frontend/cypress/e2e/*.cy.js
```
josbell: Two issues here:

1. **Flat glob silences `-r`** — `*.cy.js` expands to a list of files; `-r` only recurses when the argument is a directory. Any spec in a subdirectory is silently missed. Prefer `frontend/cypress/e2e/` as the argument so `-r` actually recurses.

2. **No output contract slot for sweep results** — the section tells the agent to "report every match," but the Output Contract template (around line 180) has no `Sibling sweep` or `Affected specs` field. An agent producing the structured report format will not know where to put these results and will likely drop them. Consider adding a `### Sibling specs affected` field to the template, or wiring the sweep into the `### Recommended next action` field explicitly.

weimiao67 (approved):

jonnalley:
@josbell Verified all six against the actual environment/code before touching anything.

**Confirmed and fixed:**
- Missing `-d` on the Step 5 `docker compose up` — confirmed empirically (I'd hit this exact hang earlier this session and always had to background it; with `-d` it returns cleanly). Fixed.
- Old-parent-revision capture gap — real. Added a "note this value" cue at the end of Step 2, since Step 4 overwrites the only place it appeared.
- Singular-migration gap — real, and the most consequential one: with two new upstream migrations forming a chain A→B, the old Step 2 check only matches A (B's `down_revision` points to A, not the old parent), so Step 3/4 would've cherry-picked and repointed onto A, leaving B out and the conflict still unresolved. Added guidance to follow the chain to its true end before cherry-picking/repointing.
- Flat-glob defeating `-r` in the sibling sweep — confirmed technically (verified with a throwaway test), though there are no subdirectories under `cypress/e2e/` today so it's not live-broken. Changed the grep target to the directory itself.
- No Output Contract field for sweep results — real, and I'd actually hit this myself earlier in the session (fell back to an ad hoc bullet list because there was nowhere in the rigid template to put multi-file sweep findings). Added an optional `### Sibling specs affected` section, omitted when no sweep was done.

**Pushed back on:**
- The `--profile setup` claim doesn't hold up. I ran the exact command and checked `podman ps -a`: `disable-users` does not start alongside `data-import` — Compose's explicit service-name arguments restrict scope regardless of profile activation; sharing a profile isn't the same as being a dependency. More importantly, the suggested "canonical form" (dropping `--profile setup`) fails outright here — `WARNING: missing services [data-import]`, exit 1 — the exact error I'd already hit and worked around earlier this session before adding the flag in the first place. Left the skill's `--profile setup` usage as-is; no code change for this item.

Full diff: two doc-only commits, no test suite implications.

PR #6247, opened 2026-09-16
fix: hide edit controls from read-only users

Santi-3rd:
## What changed

Read-only users were seeing edit affordances they shouldn't. This PR closes two gaps:

**`AgreementTableRow.jsx`** — On the Agreements list, hovering a row swapped the last column (FY Obligated amount) for the edit/delete action icons. For read-only users those icons are `null`, so the cell rendered empty and the amount disappeared on hover. Added a `!isReadOnly` guard to the last-cell ternary so read-only users keep seeing the FY Obligated amount on hover, mirroring the existing pattern in `BLIRow.jsx`.

**`ProjectDetail.jsx` / `ProjectDetailsView.jsx`** — The Project Details Edit button was gated only on the backend `_meta.isEditable` flag, with no client-side role check, so read-only users saw a (disabled) Edit button. Wired `useIsUserReadOnly()` through and gated both the enabled and disabled button variants on `!isReadOnly`, so read-only users see no Edit button at all. Non-read-only editors who aren't team members still get the disabled "You do not have permission" button.

**Tests** — Added a read-only case to `ProjectDetailsView.test.jsx` (button fully absent even when `canEdit` is true) and to `ProjectDetail.test.jsx` (no Edit button when the active user has the `READ_ONLY` role, even with `_meta.isEditable: true`). Existing `AgreementTableRow` read-only tests still pass.

## Issue

#6178 

## How to test

1. `cd frontend && bun run test --watch=false src/pages/projects/detail/ src/components/Agreements/AgreementsTable/AgreementTableRow.test.js`
2. Run the app and sign in as a read-only user:
   - **Agreements list:** hover over a row — the FY Obligated amount should stay visible (previously the last column went blank).
   - **Project Details:** open a project — there should be no Edit button (enabled or disabled).
3. Sign in as a non-read-only user who is *not* a team member on a project — the disabled "You do not have permission to edit this project" button should still appear.

## A11y impact

## Storybook

## Screenshots

_Recommend adding before/after of the Agreements list row on hover and the Project Details header for a read-only user._

## Links

N/A


rajohnson90:
AI Review identified the following potential issues. Fixing them now:
1. ProjectDetailsView.jsx:66 — Edit button gates on raw isReadOnly instead of the codebase's useCanEditByRole pattern, so a user
     with both SUPER_USER and READ_ONLY roles loses edit access (contradicts the "superuser wins" invariant used elsewhere).
  2. AgreementTableRow.jsx:121 (and changeIcons at line 90) — same issue: checks raw isReadOnly instead of also considering
     isSuperUser (already in scope), so a superuser+read-only user can't edit/delete from the table row.
  3. AgreementTableRow.test.js:293 — no test asserts the FY Obligated amount actually renders on hover for read-only users; existing
     tests only check that edit/delete icons are absent, so a regression of the line-121 fix would go undetected.

rajohnson90 (approved):

josbell (changes requested):
## Code Review

### Blocking

**1. `useIsUserReadOnly()` should be `useCanEditByRole()` — superuser with READ_ONLY role loses edit access**

`ProjectDetail.jsx:31` calls `useIsUserReadOnly()`, which has no superuser escape hatch. `useCanEditByRole()` — used by every sibling page (AgreementDetails, AgreementBudgetLines, AgreementProcurementTracker) — returns `isSuperUser || !isReadOnly`. Its JSDoc explicitly documents the dual-role case and states superuser should win.

A superuser who also holds READ_ONLY will have the project edit button hidden here, but not on any agreement page — a silent behavioral divergence.

The same issue exists pre-existing in `AgreementTableRow.jsx:58`, which uses `isSuperUser` + `useIsUserReadOnly()` independently instead of `useCanEditByRole()`.

**Recommended fix:** In `ProjectDetail.jsx`, replace `useIsUserReadOnly()` with `useCanEditByRole()` and derive:
```js
const canEdit = canEditByRole && (project?._meta?.isEditable ?? false);
```
Then remove the `isReadOnly` prop from `ProjectDetailsView` entirely — it shouldn't need to know about roles separately when `canEdit` already encodes that. In `AgreementTableRow.jsx`, replace the `isSuperUser` selector + `useIsUserReadOnly()` pair with a single `useCanEditByRole()` call.

---

### Non-blocking

**2. Read-only non-team-members silently lose the "no permission" tooltip** (`ProjectDetailsView.jsx:83`)

When `isReadOnly=true && canEdit=false`, both the edit button (`!isReadOnly && canEdit`) and the disabled tooltip (`!isReadOnly && !canEdit`) are gated on `!isReadOnly`, so neither renders. Previously, a read-only user who wasn't a team member saw a disabled button explaining why editing is blocked — now they see nothing. Was this intentional?

**3. Edit form branch unguarded by `!isReadOnly`** (`ProjectDetailsView.jsx` — wherever `isEditMode ? <ProjectDetailForm>` is)

The `isEditMode` ternary that renders the edit form doesn't check `!isReadOnly`. No current path can set `isEditMode=true` for a read-only user, but a mid-session role change would expose the full form. Worth adding `!isReadOnly &&` defensively on that branch.

weimiao67 (approved):
looks good to me as long as all Josbell's findings are resolved.

Santi-3rd:
@josbell I went through your findings 

1. Was fixed in a previous commit
2. Since the read-only user will not see the edit buttons, the "no permission" tooltip is unnecessary.   
3. I added the recommended safe guard

josbell (approved):

Issue #6144, opened 2026-08-25
Show agreement nickname instead of full name


jonnalley:
### User Story
As an OPS user, I want to see an agreement's nickname/acronym in place of its full name wherever agreement titles are listed, searched, or exported, so that I can recognize agreements by the shorter name my team actually uses.

### Acceptance Criteria
- [ ] When an agreement has a nickname, that nickname displays instead of the full agreement name everywhere a single agreement title is shown in a read-only context — including tables/lists, dropdowns and filters, and system notifications — with the exception of the agreement's own details page, which continues to show its full name.
- [ ] When an agreement has no nickname, the full agreement name continues to display exactly as it does today, in all of those same contexts.
- [ ] A user can find an agreement in a searchable/filterable list by typing either its nickname or its full name.
- [ ] The agreements list export includes both the full agreement name and the nickname (if any) as distinct values for every agreement, rather than one replacing the other.
- [ ] Exports other than the agreements list export (e.g. budget line exports) that reference an agreement show a single, nickname-preferred value, consistent with the general rule above.
- [ ] An authorized user can add, change, or remove an agreement's nickname at any point in the agreement's lifecycle, including after the agreement has been awarded.
- [ ] Once an agreement is awarded, its name/title remains locked from further edits exactly as it does today; this lock does not extend to the nickname field.
- [ ] Editing an agreement's nickname requires no permission beyond what's already required to edit that agreement today.

### Testing

**Automated — positive cases**
- Backend: an agreement with a non-blank nickname returns the nickname as its display value; an agreement with no nickname (null/blank) returns the full name as its display value.
- Backend: system notification/message content generated for an agreement (e.g. approval-request messages) reflects the nickname-preferred display value.
- Backend: an awarded agreement accepts a nickname change while its name/title change is still rejected, confirming the two fields are independently governed.
- Frontend: table rows, dropdown/filter option lists, and the agreements list export render the nickname-preferred value when present, and the full name when not.
- Frontend: filter/search input matches an agreement whether the typed text corresponds to its name or its nickname.
- Frontend: the agreements list export contains both the full name and the nickname as separate values, including for agreements with no nickname (blank nickname value, non-blank name).

**Automated — negative cases**
- Backend: attempting to change the name/title of an awarded agreement is still rejected, unchanged from current behavior.
- Backend: an agreement's display value never omits both name and nickname — a missing nickname always falls back to the full name rather than showing blank.
[...]

PR #6249, opened 2026-09-16
feat: show agreement nickname instead of full title

rajohnson90:
## What changed
Makes the agreement's nickname the primary display string in read-only contexts (tables, dropdowns, filters, notifications, exports) app-wide, while the agreement's own details page and the agreements-list export continue to show the full title. Implementation plan: `.claude/stories/6144-agreement-nickname-display.md`.

**Backend**
- `models/agreements.py`: `Agreement.display_name` now prefers `nick_name` (falls back to `name` for null/blank/whitespace-only); adds a `full_name` alias and a `display_name_expression()` SQL classmethod for ORDER BY/WHERE parity.
- `services/budget_line_items.py`: BLI agreement-name filter now matches either the full name or the nickname (fixes a round-trip bug that would silently return zero rows for nicknamed agreements); filter-options and the AGREEMENT_NAME sort use the nickname-preferred value.
- `services/agreements.py`: search matches nickname or name; the *partial*-match name filter also matches nickname (the *exact*-match branch used for title-uniqueness checks intentionally does not, to avoid a false "title already exists" collision); sort and filter-options use nickname-preferred display.
- `schemas/budget_line_items.py`: `SimpleAgreementSchema` gains `nick_name` and `display_name`.
- `openapi.yml`: documents the nickname-preferred semantics on affected schemas/params.

**Frontend**
- New `getAgreementDisplayName` helper backing every read-only agreement name render.
- `AgreementNameComboBox` + shared `ComboBox`: options carry both name and nickname; typing either resolves to the same option (dedup now keyed by agreement `id`, not the rendered label — fixes a bug where colliding name/nickname strings could silently drop an agreement from a picker).
- Filter tags/tables/sorts/exports (`BLIFilterTags`, `AllBLIRow`, `CANBudgetLineTable`, `AgreementSelect`, `useSortableData`, `ProcurementDashboardPage`, BLI export) render the nickname-preferred name.
- Agreements-list export gets a new "Agreement Nickname" column; `currencyColumns` indices now derive from the header array instead of a hard-coded list.
- The 8 review/approval page subtitles now show the nickname-preferred name — its own commit, easy to revert if UX wants the full title there instead.
- The agreement's own details page, the Name/Nickname side-by-side fields, and the nickname `Tag` are explicitly untouched — guarded by comments and negative tests.

**Not included** (flagged in the plan, deferred pending product confirmation): collapsing the Projects filter's duplicate name+nickname entries into one nickname-preferred option per agreement.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6144

## How to test

Automated:
```
cd backend/ops_api
DB_PORT=5433 pipenv run pytest tests/ops/agreement tests/ops/budget_line_items tests/ops/procurement_tracker

cd frontend
bun run test --watch=false
```

Manual (`docker compose up --build`):
1. Add a nickname to an agreement that lacks one. Confirm it now shows in the agreements table, the Agreement Title filter dropdown, the all-budget-lines table, and the CAN detail budget-line table.
2. That agreement's own page still shows the full title in the `<h1>`/breadcrumb, and the Name/Nickname fields still render side by side.
3. In the agreements-list filter, type the nickname → the option appears; clear, type a distinctive word from the full title → the same option appears. Select it, confirm the table filters and the chip reads the nickname, and the chip's X clears it. Repeat on the budget-lines list.
4. Export the agreements list — confirm "Agreement" (full title) and "Agreement Nickname" are separate columns, and Total/FY Obligated/Subtotal/Fees/Lifetime Obligated are still currency-formatted.
5. Export budget lines — the Agreement column shows the single nickname-preferred value.
6. Trigger a pre-award or award approval request notification — confirm it names the agreement by nickname.
7. On an awarded agreement, confirm the Title input is disabled while the Nickname input stays editable, and a nickname edit saves successfully.
8. Clear a nickname — confirm the full title reappears everywhere from step 1.

## A11y impact

(Left unchecked — this PR changes what text renders in many existing surfaces but doesn't add/remove interactive elements; flagging for reviewer judgment.)

## Storybook

## Screenshots

_Text/behavior changes only — no new visual design. Happy to add before/after screenshots of the agreements table, filter dropdown, and export if useful for review._

## Links

N/A


rajohnson90:
Correctness bugs found during AI review
  ~~1. budget_line_items.py:476 — BLI agreement_name filter ORs Agreement.name and Agreement.nick_name without scoping to the same
     agreement, so a nickname collision between two different agreements leaks one agreement's budget lines into another's filtered
     results.~~
  ~~2. Agreement.display_name vs display_name_expression() (backend/models/agreements.py:264) — Python .strip() vs SQL TRIM() diverge on
     whitespace like tabs/newlines, so DB sort order can disagree with the rendered name. The PR's own xfail(strict=True) test
     documents this without fixing it.~~
  3. ProjectSpendingAgreementRow.jsx:53 — getAgreementName(agreement) ?? NO_DATA no longer falls back correctly since the helper now
     returns "" instead of undefined, so malformed agreements render a blank cell instead of "No data available".
  4. ComboBox.hooks.js:66 — pre-existing numeric-label sort logic now triggers for short numeric nicknames (e.g. "24", "100"),
     silently overriding AgreementNameComboBox's alphabetical sort.

  Untested/undocumented behavior changes
  5. procurement_tracker_steps.py:27 — NestedAgreementSchema.display_name silently becomes nickname-preferred in notification
  payloads; no test or openapi.yml coverage.

  Duplication/drift risk
  6. agreements.py:745 — filter-options endpoint reimplements the nickname-fallback rule in raw Python instead of reusing
  display_name/display_name_expression(), so a future rule change could silently diverge here.
  7. opsAPI.js:129/408 and BLIFilterTags.jsx:70/134 — the same fallback chain (name ?? display_name ?? title) is duplicated in 4
  places; a future edit to one copy without its twin breaks tag removal.
  8. AgreementNameComboBox.jsx:51 — both option-building branches duplicate the identical option shape/searchText construction.

  Test quality
  9. test_budget_line_item.py:3076 — new sort test asserts Python casefold order against raw SQL collation order, which isn't actually
  guaranteed by the query.
  10. test_agreement.py (lines 48, 1204, 1230, 1256, 1306, 1335, 1355, 1382, 1719, 2035, 2071) — ~11 pre-existing display_name == name
  assertions pass only coincidentally because their fixtures lack a nick_name.

frontend/src/components/UI/Form/ComboBox/ComboBox.jsx
```diff
@@ -22,6 +22,13 @@ const LoadingMessage = (props) => (
     </components.LoadingMessage>
 );
 
+// Widens react-select's default filter to also match against a hidden `searchText` field
+// (e.g. an agreement's full name/nickname pair) so typing either resolves to the same option.
+// Using createFilter (rather than a hand-rolled predicate) preserves react-select's exact
+// default semantics (ignoreCase, ignoreAccents, trim, substring) — options without
+// `searchText` filter exactly as before. Ref: issue #6144 AC 3.
+const filterOption = createFilter({ stringify: (option) => `${option.label} ${option.data?.searchText ?? ""}` });
```
weimiao67: **Filter regression for every ComboBox in the app.**

react-select's `defaultStringify` is ``​`${option.label} ${option.value}`​`` (`node_modules/react-select/dist/Select-ef7c0426.esm.js:506`). Overriding `stringify` drops `option.value` from the searchable text, and `filterOption` is passed unconditionally to every `<Select>` this component renders — so the comment just above ("options without `searchText` filter exactly as before") doesn't hold.

`useComboBox` sets `value: item.id` (`ComboBox.hooks.js:44`). Concretely: `ProjectTypeComboBox` builds options as `{ id: "ADMINISTRATIVE_AND_SUPPORT", title: "Admin & Support" }`, so typing `administrative` used to match via the value and now returns "No options". Same loss for typing a numeric id in `ProjectTitleComboBox` / `PortfoliosComboBox` / `AgreementTypeComboBox`. The full frontend suite passes on this branch, so nothing currently covers value-based matching.

Suggested fix — keep the value in the stringify:

```js
const filterOption = createFilter({
    stringify: (option) => `${option.label} ${option.value} ${option.data?.searchText ?? ""}`
});
```

backend/ops_api/ops/services/budget_line_items.py
```diff
@@ -466,9 +466,13 @@ def _apply_agreement_type_filter(self, query, agreement_types):
...
     def _apply_agreement_name_filter(self, query, agreement_names):
-        """Apply agreement name filter if provided."""
+        """Apply agreement name filter if provided.
+
+        Matches against either the full name or the nick_name, since the filter
+        options endpoint (and the frontend) may round-trip either value. Ref: #6144.
+        """
         if agreement_names:
-            query = query.where(Agreement.name.in_(agreement_names))
+            query = query.where(or_(Agreement.name.in_(agreement_names), Agreement.nick_name.in_(agreement_names)))
```
weimiao67: **This can select agreements the user never picked.**

The `or_` matches either column across *all* agreements, not "this agreement by either of its names." Say agreement A has `name="ABC Study"` and no nickname, and agreement B has `name="Longer Title"` with `nick_name="ABC Study"`. The user selects A in the Agreement Title filter; `opsAPI.js:408` sends `agreement_name=ABC Study` (it now always sends the full `name`), and the response includes B's budget lines too — plus an inflated `count` and `totals`.

Since the frontend is guaranteed to send the full `name`, the `nick_name` leg isn't needed for the round trip and only adds false positives. I'd either drop it or filter by agreement `id` instead.

frontend/src/pages/budgetLines/list/BLIFilterTags.jsx
```diff
@@ -124,8 +129,9 @@ export const BLIFilterTags = ({ filters, setFilters, fyHelpers }) => {
 
     const agreementTitleTags = useMemo(() => {
         if (!Array.isArray(filters.agreementTitles)) return [];
+        // Nickname-preferred tag text — must match the removal predicate in removeFilter exactly.
         return filters.agreementTitles.map((title) => ({
-            tagText: title.name,
+            tagText: title.title ?? title.display_name ?? title.name,
```
weimiao67: **Tag text and the removal predicate at `:70` both key on a string that's no longer unique.**

The display value can now be a name *or* a nickname — which is exactly the collision `AgreementNameComboBox.jsx` calls out as "a legitimate (if confusing) UI state" when it switched its dedup Map to `id`. Select agreement A (`nick_name="ABC"`) and agreement B (`name="ABC"`, no nickname): two identical chips render, and clicking either X removes *both* entries from `filters.agreementTitles`, silently clearing a filter the user still wanted.

`AgreementsFilterTags.hooks.js:145` (`name.title !== tag.tagText`) has the same exposure via the same combobox change. Keying tag removal on the agreement `id` fixes both.

frontend/src/components/Agreements/AgreementsTable/AgreementsTable.helpers.js
```diff
@@ -1,14 +1,21 @@
...
+/**
+ * Nickname-preferred agreement label for the agreements table row. Delegates to the
+ * shared `getAgreementDisplayName` helper (see agreement.helpers.js) so all callers of
+ * this table (AgreementTableRow, ProjectSpendingAgreementRow, ProcurementDetailsTableRow)
+ * pick up the nickname-preference for free. Ref: issue #6144.
+ */
 export const getAgreementName = (agreement) => {
     handleAgreementProp(agreement);
-    return agreement.display_name;
+    return getAgreementDisplayName(agreement);
```
weimiao67, line 18: **This makes a downstream `NO_DATA` fallback unreachable.**

`getAgreementDisplayName` ends in `?? ""`, so it returns `""` — never `undefined` — when the agreement has no `display_name`, `nick_name`, or `name`. That makes `?? NO_DATA` at `ProjectSpendingAgreementRow.jsx:53` dead code, and an agreement with a null name (the DB allows it — `services/agreements.py:738` explicitly filters `Agreement.name.isnot(None)`) renders an empty Project Spending cell instead of the dash.

Switching that call site to `|| NO_DATA` matches what this PR already does in `CANBudgetLineTable.jsx` and `use-sortable-data.hooks.js`.

weimiao67 (commented):
Review pass over the nickname changes. Four findings below as inline comments — (1) is the one I'd fix before merge; (2)–(4) are lower severity. Also two notes not worth inlining: the `xfail(strict=True)` whitespace gap in `display_name_expression()` is cheaply closed with `func.btrim(cls.nick_name, " \t\n\r\f\v")` so the test can become a normal assertion instead of pinning a known bug; and the branch predates `b446f0a0c`, which edits the same `rowMapper` in `AgreementsList.jsx` and adds ~138 lines to `AgreementsList.test.jsx` — worth rebasing (the derived `currencyColumns` indices stay correct through it, but the test file will likely conflict).

Verified correct along the way: `currencyColumns` from `tableHeader.indexOf(...)` resolves to `[5, 6, 9, 10, 11]`, correctly shifted by the new Nickname column; the exact-match `name` branch deliberately excluding `nick_name` is right and `AgreementEditForm.hooks.js:272` does hit it with `exact_match` defaulting to `True`, so title-uniqueness is intact; keying the combobox dedup Map on `id` rather than the label is a real bug fix.

rajohnson90:
When the budget team member approves pre-award the toast notification still says the Agreement title instead of the agreement nickname
Screenshot 2026-09-16 at 3 47 55 PM

frontend/src/components/UI/Form/ComboBox/ComboBox.jsx
rajohnson90: Valid. Fixed this

backend/ops_api/ops/services/budget_line_items.py
rajohnson90: Fixed. Just went back to only name because agreement name is already passed in to this check.

frontend/src/components/Agreements/AgreementsTable/AgreementsTable.helpers.js
rajohnson90: Fixed

frontend/src/pages/budgetLines/list/BLIFilterTags.jsx
rajohnson90: Fixed

weimiao67 (approved):

rajohnson90:
Converting to draft because we actually still need more UX design before this task is ready to complete.

Issue #6140, opened 2026-08-25
Default Projects and Agreements lists (and their exports) to all fiscal years


jonnalley:
### User Story
As a user viewing the Projects or Agreements list, I want the list — and anything I export from it — to show all fiscal years by default, so that I see the complete picture across fiscal years without having to change the filter every time I visit.

### Acceptance Criteria

Core:
- [ ] When a user opens the Projects list without having chosen a fiscal year, projects from all fiscal years are shown.
- [ ] When a user opens the Agreements list without having chosen a fiscal year, agreements from all fiscal years are shown.
- [ ] A user can still narrow either list to one specific fiscal year, and the list updates to show only that year (existing behavior, unaffected by this change).
- [ ] Each time a user opens the Projects or Agreements list, it starts showing all fiscal years — regardless of a specific fiscal year chosen on a prior visit (existing behavior, no memory or saving of the selection).

Exports:
- [ ] When a user exports the Projects list without having narrowed to a specific fiscal year, the exported data covers all fiscal years.
- [ ] When a user exports the Agreements list without having narrowed to a specific fiscal year, the exported data covers all fiscal years.
- [ ] When a user has narrowed either list to one specific fiscal year, the export reflects only that fiscal year (existing behavior, unaffected by this change).
- [ ] Any fiscal-year-specific labeling shown within an export (e.g. a fiscal-year-obligated amount heading) accurately reflects "all fiscal years" when that is the active scope, rather than showing a single year's label.

Regressions:
- [ ] Showing all fiscal years by default does not surface any project or agreement data beyond what the user is already authorized to see. (should be a no-op)
- [ ] Viewing or exporting all fiscal years by default does not introduce a noticeable slowdown for typical data volumes compared to today's experience.

---

### Additional Requirements (PR #6253)

**Agreements list — All FYs column and card behavior**
- [ ] When All FYs is selected, the last table column header reads **"Lifetime Obligated"** (not "FY Obligated" / N/A) and each row shows the sum of obligated budget lines across all fiscal years for that agreement.
- [ ] When All FYs is selected, the Agreements summary card hides the "New" and "Continuing" sub-columns — new-vs-continuing cannot be determined across all fiscal years.
- [ ] When All FYs is selected, the export contains exactly one "Lifetime Obligated" column (the redundant static column is suppressed to avoid duplication).
[...]

PR #6263, opened 2026-09-21
fix: agreements list Lifetime Obligated — enable sorting and fix All FYs sort order

josbell:
## What changed

Follow-up fixes to the Agreements list after #6253 merged:

1. **Lifetime Obligated column is now sortable** — the column was incorrectly marked as disabled/non-interactive when All FYs was selected, showing a misleading tooltip and greyed-out cursor. The disabled state and sort guard have been removed; the column is sortable in all states.

2. **Sort by Lifetime Obligated actually uses lifetime values** — when All FYs is selected and the user sorts by the Lifetime Obligated column, the backend was secretly sorting by the *current fiscal year's* obligated amount (falling back to current FY when no FY filter was specified). The sort now correctly uses each agreement's `lifetime_obligated` — the sum of obligated BLIs across all fiscal years.

3. **Sort is preserved when switching FY scope** — switching the FY dropdown to "All" while the Lifetime Obligated column is active no longer resets the sort to the default. The backend handles the scope change correctly, so the sort arrow is preserved.

4. **Removed FY-specific disabled tooltip from shared `Table` component** — domain-specific copy ("Sorting by X is only meaningful when a specific fiscal year is selected") was added to a shared UI primitive. The general `"Click to sort by X"` tooltip is preserved; the conditional disabled-state variant is removed.

## Known limitation

When the **Compare Fiscal Years** filter is used (selecting 2+ specific years simultaneously) and the user sorts by Lifetime Obligated, the backend sorts by the all-time lifetime obligated rather than the sum of obligated amounts for only the selected years. This is a narrow edge case that only produces a misleading sort order when an agreement has significant obligations *outside* the selected comparison window. Tracked as a TODO in `agreements.py`; will be addressed if user-reported.

## Issue

https://github.com/HHS/OPRE-OPS/issues/6140

## How to test

1. Navigate to the Agreements list — FY defaults to **All**.
2. Click the **Lifetime Obligated** column header — confirm the column sorts (ascending/descending arrow appears, rows reorder by lifetime obligated sum).
3. Confirm no tooltip says "only meaningful when a specific fiscal year is selected."
4. Switch to a specific FY (e.g. 2044) — confirm the column sorts by that FY's obligated amount.
5. Switch back to All — confirm the sort arrow is preserved (not reset to default).
6. Confirm sort order reflects lifetime totals, not current-FY values.

## A11y impact

## Storybook

## Screenshots

_N/A_

## Links

_N/A_


josbell:
Addressed both Copilot review comments:

**Comment 1 (multi-year filter sorts by lifetime_obligated):** When multiple FYs are selected via Compare Fiscal Years, `resolve_fiscal_year` returns `None` and the sort falls through to `lifetime_obligated` rather than the sum of `fy_obligated` for only the selected years. A TODO comment has been added at the branch point in `_sort_agreements` documenting exactly when this is an issue. The limitation is also noted in the PR description under 'Known limitation'. The edge case only produces misleading order when an agreement has significant obligations *outside* the selected comparison window — narrow enough to defer.

**Comment 2 (stale sort reset):** Fixed. The `handleChangeFiscalYear` guard that reset sort to AGREEMENT when switching to All has been removed. The backend now handles the scope change correctly (sorts by `lifetime_obligated`), so the active sort arrow is preserved. Regression test updated to assert `setSortConditions` is NOT called on the switch.

rajohnson90:
Two issues identified by Claude that I'll have it fix:
 1. backend/ops_api/tests/ops/agreement/test_agreement.py:3174 — test_sort_by_fy_obligated_specific_fy_returns_200 doesn't assert sort order, unlike the sibling
     All-FYs test right above it, so a future sort regression on specific-FY agreements would still pass this test.
  2. backend/ops_api/ops/services/agreements.py:1401 — the new lifetime_obligated sort uses an inline lambda instead of a named function, breaking the pattern the
     other sort branches follow (agreement_total_sort, project_sort, etc.).

rajohnson90 (approved):

Santi-3rd (approved):

PR #6262, opened 2026-09-19
feat: add shared fiscalYearFilter helpers for filter standardization

josbell:
## What changed

Adds `src/helpers/fiscalYearFilter.helpers.js` — 6 pure, page-agnostic fiscal-year filter functions implementing Model B (two-input precedence: persistent dropdown FY + Compare-FYs override). No pages are wired in this PR; this is a no-behavior-change extraction step.

Functions added: `isAllSentinel`, `normalizeFYTag`, `deriveDropdownValue`, `resolveForAPI`, `deriveFYTags` (with optional `filterKey` for BLI compatibility), `handleFYTagRemoval`.

Also adds `fiscalYearFilter.helpers.test.js` with 45 unit tests.

This is **PR 1 of 3** for sub-issue #6256. The helpers will be wired to the Agreements list files in PR 2. Projects, BLI, CANs, and Reporting will adopt them in follow-up PRs per #5332.

## Issue

#6256 (sub-issue of #5332 — Standardize filter behavior across all list pages)

## How to test

1. `cd frontend && bun run test --watch=false src/helpers/fiscalYearFilter.helpers.test.js` — 45 tests pass
2. `bun run test:coverage --watch=false` — 90% coverage gate holds
3. No runtime surface to exercise — helpers only, no page wired yet

## A11y impact

- [x] No accessibility-impacting changes in this PR

## Storybook

- [x] No UI component changes in this PR

## Screenshots

N/A

## Definition of Done Checklist
- [ ] OESA: Code refactored for clarity
- [ ] OESA: Dependency rules followed
- [x] Automated unit tests updated and passed
- [ ] Automated integration tests updated and passed
- [ ] Automated quality tests updated and passed
- [ ] Automated load tests updated and passed
- [ ] Automated a11y tests updated and passed
- [ ] Automated security tests updated and passed
- [x] 90%+ Code coverage achieved
- [ ] Form validations updated

## Links

- Story: https://github.com/HHS/OPRE-OPS/issues/5332
- Sub-issue: https://github.com/HHS/OPRE-OPS/issues/6256

Issue #5332, opened 2026-03-20
Standardize filter behavior across all list pages


josbell:
## Overview

Standardize how filters, fiscal year (FY) selection, and filter tags behave consistently across all list pages.

**Pages in scope:** Agreements list, CANs list, Projects list, Budget Line Items (BLI) list, OPRE Budget Reporting list, Project details — Spending tab

### Page-Level FY Dropdown (Shortcut)
- Reflects the active FY at all times
- Displays "Multi" when 2+ FYs are selected in the Compare Fiscal Years filter
- Reverts to "All FYs" when all FY filters are cleared
- If the user applies other filters but makes no selection in Compare Fiscal Years, the query uses the dropdown's current FY

### Compare Fiscal Years Filter (Filter Panel)
- Retain "All FYs" option in the component
- Selecting "All FYs" in the filter panel shows an "All FYs" tag and sets the dropdown to "All FYs"
- The Compare Fiscal Years selection is the **source of truth** for FY — it drives the page-level dropdown, not the other way around
- Compare Fiscal Years remains only on pages where it currently exists (Agreements, Projects, BLI)

### Filter Tags ("Filters Applied")
- Tags appear only when the user selects filters in the filter panel — not from dropdown-only changes
- FY tags only appear on pages that have the Compare Fiscal Years component
- Remove any duplicate FY entries from the tag list
- If all filters are cleared, tags disappear and dropdown reverts to "All FYs"

### Filter Modal Buttons
- **Reset** — clears all form fields including Compare Fiscal Years; does not fire a query; dropdown remains unchanged
- **Apply** — always commits current modal state; an empty modal clears all active filters and filter tags, results revert to dropdown FY only

kimschulke:
as discussed with @josbell today in stand-up, he set up 2 options we had discussed and UXers will review and experiment with them today or early next week 🎉

kimschulke:
@josbell did your PR get reviewed/approved yet? 😄 Let me know when I can start testing, thank you!

jonnalley:
moving back to sprint backlog as I don't think this is actively being worked

kimschulke:
@josbell Let me know when you want to collab on this again. For now, I will remove the UX/UI Review label. Thanks!

jonnalley:
@josbell @kimschulke @andressacurvelo @npage123 I would humbly submit this may need even further revision in the advent of the changes coming with #5627 and beyond.

jonnalley:
[...]

PR #6265, opened 2026-09-21
feat: standardize FY filter behavior on Agreements list (Model B)

josbell:
## What changed

Standardizes fiscal-year (FY) selection, filter tags, and filter modal behavior on the Agreements list page per story #5332 (sub-issue #6256). This is PR 2 of 3 — PR 1 (shared helpers) is already merged.

**Core architecture (Model B — two-input precedence):**
- `selectedFiscalYear` — the page-level shortcut dropdown (never produces a tag)
- `filters.fiscalYear` — the Compare Fiscal Years modal selection (the only source of FY tags; drives the dropdown display when non-empty)

**Changes:**
- Removes `getFiscalYearFilter()` and `hasOtherFilters` from `AgreementsList.jsx`; replaces with `resolveForAPI()` + `deriveDropdownValue()` from the shared helper module
- Dropdown now correctly shows **Multi** (2+ FYs selected in panel) and **All FYs** (sentinel selected in panel)
- Changing the dropdown shortcut now clears only `filters.fiscalYear`, not all filters
- **Reset** clears modal form fields only — no query fires, dropdown unchanged
- **Apply** with empty modal reverts results to the dropdown FY only
- FY tags deduplicated; no more "FY FY 2024" double-prefix; "All FYs" sentinel shows correctly
- Removing all FY tags reverts the dropdown to "All FYs"
- Removed "Fiscal Year XXXX" placeholder from Compare Fiscal Years combobox
- Modal buffers reseed from parent state on every open (fixes stale state after Reset-without-Apply)
- `"Multi"` dropdownValue collapses to `"All"` for child components (table headers, export columns, summary cards)

**Files changed:**
- `AgreementsList.jsx` — main page wiring
- `AgreementsFilterButton.hooks.js` — reset/apply/reseed behavior
- `AgreementsFilterButton.jsx` — controlled modal state, placeholder removal
- `AgreementsFilterTags.hooks.js` — FY tag generation and removal via shared helpers
- New: `AgreementsFilterTags.hooks.test.js`
- Updated: `AgreementsList.test.jsx`

## Issue

#6256 (sub-issue of #5332 — Standardize filter behavior across all list pages)

## How to test

1. Open the Agreements list
2. **Compare Fiscal Years → Multi**: open Filters, select 2 FYs → dropdown shows "Multi", two deduped tags appear
3. **Compare Fiscal Years → All FYs**: select "All FYs" in panel → one "All FYs" tag, dropdown shows "All"
4. **Tag removal reverts to All**: apply a FY from the panel, remove its tag → dropdown reverts to "All"
5. **Dropdown keeps other filters**: set a portfolio filter, then change the FY dropdown → portfolio filter remains active
6. **Reset fires no query**: apply filters, open modal, click Reset → no new results load; click Apply → filters cleared
7. **Reset-without-Apply**: apply filters, open modal, click Reset, close modal (X) → reopen modal → fields show the still-active filters (not empty)
8. **No placeholder**: Compare Fiscal Years field shows no default placeholder text

## A11y impact

## Storybook

## Screenshots

_N/A — behavioral/logic changes; no visual layout changes_

## Links

- Parent story: https://github.com/HHS/OPRE-OPS/issues/5332
- Sub-issue: https://github.com/HHS/OPRE-OPS/issues/6256
- PR 1 (shared helpers, merged): https://github.com/HHS/OPRE-OPS/pull/6262


josbell:
Addressed all four Copilot review comments in commit 3307e686d:

**#1 — null crash (`filters.fiscalYear === null`):** Fixed. The `useEffect` now normalizes via `filters.fiscalYear ?? []` before any `.length` access, handling the `null` emitted by `FiscalYearComboBox`'s clear control.

**#2 — FY_OBLIGATED sort not reset for panel-driven transitions:** Fixed. Added a `useEffect` on `displayFY` that resets the sort to `AGREEMENT` whenever `displayFY` transitions to `"All"` from any cause — panel sentinel, Multi (collapsed to All), or tag removal. The guard in `handleChangeFiscalYear` is kept alongside it for the direct dropdown case.

**#3 — ref guard missed `handleChangeFiscalYear` + Apply paths:** Fixed. Replaced the single-ref approach with a two-ref approach: `dropdownChangedFYRef` (set in `handleChangeFiscalYear` to suppress the reset when the dropdown itself cleared the FYs) combined with `prevFYLengthRef` to only reset `selectedFiscalYear` on genuine non-zero → zero transitions. This correctly ignores Apply-after-Reset (was already length 0, no transition) and handles the null normalization.

**#4 — untested modal state changes:** Added a test covering the dropdown year → switch-to-All → `fiscalYear: []` in query path. The Reset-no-query and modal-reseed behaviors are blocked from direct hook testing by a pre-existing circular dependency between `utils.js` and `budgetLines.helpers.js` that causes an OOM during vitest module graph resolution for the `AgreementsFilterButton` directory. These paths are covered by E2E tests in `agreementList.cy.js`.

rajohnson90:
Code Review surfaced 4 findings: 
frontend/src/pages/agreements/list/AgreementsList.test.jsx
       ● 1198 [test-coverage] The test named for the tag-removal-reverts-to-All behavior never exercises tag removal; it only
                              fires a direct dropdown change to "All", and AgreementsFilterButton/Tags are fully mocked in this
                              file, so the new ref-guarded revert logic in AgreementsList.jsx (lines 120-141) has zero real test
                              coverage.
     frontend/src/pages/agreements/list/AgreementsFilterButton/AgreementsFilterButton.hooks.js
       ●   18 [test-coverage] No unit test exists for useAgreementsFilterButton's new showModal-reseed effect or the changed
                              resetFilter (which no longer calls setFilters), and AgreementsFilterButton is mocked wholesale in
                              AgreementsList.test.jsx, leaving this stateful reseed/reset logic completely untested.
     frontend/src/pages/agreements/list/AgreementsList.jsx
       ●  158 [duplication]   handleChangeFiscalYear still manually resets FY_OBLIGATED sort inline (lines 162-164), duplicating
                              the new displayFY-transition-to-"All" effect (lines ~146-152) that was added to generalize this
                              exact reset; the special case should be removed now that the effect covers it.
     frontend/src/helpers/fiscalYearFilter.helpers.js
       ●   62 [correctness]   deriveDropdownValue returns the raw compareFYs id as the dropdown value even when that FY isn't in
                              the hardcoded constants.fiscalYears list rendered by <FiscalYear> (AgreementsList.jsx doesn't pass
                              a fiscalYears prop), so the <select> can end up with no matching <option>.
                              
I'll resolve them all and then approve.

rajohnson90 (approved):

Santi-3rd (approved):

frontend/src/pages/agreements/list/AgreementsFilterButton/AgreementsFilterButton.hooks.js
```diff
@@ -1,65 +1,75 @@
...
-        if (filters.awardType) {
-            setAwardType(filters.awardType);
-        }
+        setAwardType(filters.awardType ?? []);
     }, [filters.awardType]);
 
     const applyFilter = () => {
+        // Signal to the page-level FY reset effect that this emptying came from Apply,
+        // not from tag removal — so it should NOT revert selectedFiscalYear to "All".
+        if (applyFiredFYRef) applyFiredFYRef.current = true;
```
weimiao67, line 72: **Bug:** `applyFiredFYRef` can get stuck `true` and later suppress the revert-to-"All" behavior.

This flag is set unconditionally on every Apply click, but it's only cleared inside the `useEffect` in `AgreementsList.jsx` keyed on `[filters.fiscalYear]` (lines 139-154) — which only re-runs when that array's *reference* changes.

Repro:
1. Apply a Compare FY (e.g. 2024) — `filters.fiscalYear` gets a new array reference, the effect runs, and correctly clears the ref.
2. Reopen the modal, change only the Portfolio filter (leave the FY combobox untouched, so the local `fiscalYear` buffer keeps the same array reference), and click Apply. `setFilters` writes back the *same* `fiscalYear` reference, so the effect's dependency doesn't change and it never re-runs — the ref stays stuck at `true`.
3. Remove the FY tag. This creates a new (empty) array, the effect fires, sees the stale `true` ref, treats it as "came from Apply," and returns early — skipping `setSelectedFiscalYear("All")". The dropdown stays on the stale FY instead of reverting to All.

Consider setting the ref only when the FY buffer actually changed (e.g. compare against `filters.fiscalYear` before writing, or diff the arrays) rather than unconditionally on every Apply.

